PR #3, opened 2021-10-08
A quick fix to backslash

libingye816:
Fix the following:

1.  ".table_name"
2. 1 at collation column
3. \d doesn't work for some tables

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/statements/MatcherStatement.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/statements/MatcherStatement.java
@@ -55,10 +55,7 @@ public class MatcherStatement extends IntermediateStatement {
     for (Command currentCommand :
         Command.getCommands(sql, this.connection, this.commandMetadataJSON)) {
       if (currentCommand.is()) {
-        System.out.println("DEBUG: ");
-        String s = currentCommand.translate();
-        System.out.println(s);
-        return s;
+        return currentCommand.translate();
       }
     }
     return sql;
--- a/src/main/java/com/google/cloud/spanner/pgadapter/statements/MatcherStatement.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/statements/MatcherStatement.java
@@ -55,7 +55,10 @@ public class MatcherStatement extends IntermediateStatement {
     for (Command currentCommand :
         Command.getCommands(sql, this.connection, this.commandMetadataJSON)) {
       if (currentCommand.is()) {
-        return currentCommand.translate();
+        System.out.println("DEBUG: ");
+        String s = currentCommand.translate();
+        System.out.println(s);
+        return s;
       }
     }
     return sql;
--- a/src/main/resources/metadata/default_command_metadata.json
+++ b/src/main/resources/metadata/default_command_metadata.json
@@ -8,7 +8,7 @@
     },
     {
       "input_pattern": "^SELECT c\\.oid,[ \n]+n\\.nspname,[ \n]+c\\.relname\nFROM pg_catalog\\.pg_class c[ \n]+LEFT JOIN pg_catalog\\.pg_namespace n ON n\\.oid = c\\.relnamespace\nWHERE c\\.relname OPERATOR\\(pg_catalog\\.~\\) '\\^\\((?<tablename>.*)\\)\\$'[ \n]+AND pg_catalog\\.pg_table_is_visible\\(c\\.oid\\)\nORDER BY 2, 3;$",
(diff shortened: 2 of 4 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/statements/MatcherStatement.java
```diff
@@ -55,7 +55,10 @@ private String translateSQL(String sql) {
     for (Command currentCommand :
         Command.getCommands(sql, this.connection, this.commandMetadataJSON)) {
       if (currentCommand.is()) {
-        return currentCommand.translate();
+        System.out.println("DEBUG: ");
```
Vizerai: I don't think these debug statements should be here.

Vizerai (commented):
We no longer need the sha256 hash?

src/main/java/com/google/cloud/spanner/pgadapter/statements/MatcherStatement.java
libingye816: Looks like in PG, hashing is not used. `SELECT a.attname,
  pg_catalog.format_type(a.atttypid, a.atttypmod),
  (SELECT pg_catalog.pg_get_expr(d.adbin, d.adrelid, true)
   FROM pg_catalog.pg_attrdef d
   WHERE d.adrelid = a.attrelid AND d.adnum = a.attnum AND a.atthasdef),
  a.attnotnull,
  (SELECT c.collname FROM pg_catalog.pg_collation c, pg_catalog.pg_type t
   WHERE c.oid = a.attcollation AND t.oid = a.atttypid AND a.attcollation <> t.typcollation) AS attcollation,
  a.attidentity,
  a.attgenerated
FROM pg_catalog.pg_attribute a
WHERE a.attrelid = '16385' AND a.attnum > 0 AND NOT a.attisdropped
ORDER BY a.attnum;`
check a.attrelid = '16385' here.

If there is any reason a hashing fucntion is needed we can add it back once the backend issue is fixed. This is just a quick patch.

libingye816: > I don't think these debug statements should be here.

done

Vizerai (approved):
LGTM

src/main/resources/metadata/default_command_metadata.json
```diff
@@ -8,7 +8,7 @@
     },
     {
       "input_pattern": "^SELECT c\\.oid,[ \n]+n\\.nspname,[ \n]+c\\.relname\nFROM pg_catalog\\.pg_class c[ \n]+LEFT JOIN pg_catalog\\.pg_namespace n ON n\\.oid = c\\.relnamespace\nWHERE c\\.relname OPERATOR\\(pg_catalog\\.~\\) '\\^\\((?<tablename>.*)\\)\\$'[ \n]+AND pg_catalog\\.pg_table_is_visible\\(c\\.oid\\)\nORDER BY 2, 3;$",
-      "output_pattern": "SELECT sha256(t.table_name::BYTEA) as oid, '' as nspname, t.table_name as relname FROM information_schema.tables AS t WHERE t.table_schema='public' AND LOWER(t.table_name) = LOWER('%s');",
+      "output_pattern": "SELECT t.table_name as oid, 'public' as nspname, t.table_name as relname FROM information_schema.tables AS t WHERE t.table_schema='public' AND LOWER(t.table_name) = LOWER('%s');",
```
snehashah16, line 11: ` t.table_name as oid `

 will this become backward incompatible when we have real oids ?
our plans for OID: "Postgres has a stable unsigned four-bytes oid for every object directly listed in its schema "

libingye816: Hi Seneha, can you give an example to elaborate more? Like Why this may break the backward compatibility. Sorry, I am not familiar with this concept.

snehashah16: Sorry, for the lack of context

Cloud Spanner and Spangres are releasing a stable version of OID [1]. If we currently use table_names, and then eventually use the bytes generated internally. Imagine a customer builds an application/tool based on these table_name as oid, and then we silently release real OIDs  which will then break that customer app/tool - this discrepancy is called backward-incompatibility.. 

While we are in preview mode (before GA release), we are ok to be backward-incompatible. But for GA and beyond, we would need to be super careful to not create such incompatibility.

[1] https://docs.google.com/document/d/1tYXvRqelO61TaDNA3W19xHJyULJ84XmC0DhdChqyIOg/edit?usp=sharing

PR #4, opened 2021-10-08
Merge to main

Vizerai:


olavloite:
Closing for now as it is currently only causing additional spamming and Gtihub Actions runs that are not needed.

PR #5, opened 2021-10-11
Adding support of PSQL -c batch option.

Vizerai:


.ci/e2e-expected/ddl-batching.txt
```diff
@@ -0,0 +1,8 @@
+       List of relations
```
voulg, line 1: It seems that ci/e2e-psql-v* tests failed. I'm not sure if it's due to this change or something else changed. Can you debug what was the issue?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -77,6 +86,36 @@ private static ResultType extractResultType(Statement statement) throws SQLExcep
     }
   }
 
+  // Split statements by ';' delimiter, but ignore anything that is nested with '' or "".
```
voulg, line 87: How about ';' inside comments? Should we first remove all comments, is this already done at a higher layer? Can you explain?

Can you add a test for `;` inside comments?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -77,6 +86,36 @@ private static ResultType extractResultType(Statement statement) throws SQLExcep
...
+      statements.add(sql.substring(index, sql.length()).trim());
+    }
+    return statements;
+  }
+
+  protected List<String> parseStatements(String sql) {
+    Preconditions.checkNotNull(sql);
+    List<String> statements = splitStatements(sql);
+    if (statements.size() > 1) {
+      batchedQuery = true;
```
voulg: Nit: Instead of defining a new property, I'd probably use statements.size() > 1 to all places used. It's much clearer what it means.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -222,6 +272,10 @@ public String getCommand() {
     return this.command;
   }
 
+  public boolean isBatchedQuery() {
```
voulg, line 293: Is this only used for testing? If yes consider adding a comment explaining.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
```diff
@@ -186,6 +186,11 @@ protected int getHeaderLength() {
    * setting for {@link ConnectionHandler}.
    */
   public void nextHandler() throws Exception {
-    this.connection.setMessageState(ControlMessage.create(this.connection));
+    try {
+      this.connection.setMessageState(ControlMessage.create(this.connection));
+    } catch (Exception e) {
+      System.out.println("**** " + e.toString());
+      throw e;
+    }
```
voulg: Is this intentional? I would expect these exceptions to be handled at a higher level....

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -300,4 +300,31 @@ public void testPortalStatementDescribePropagatesFailure() throws Exception {
 
     intermediateStatement.describe();
   }
+
+  @Test
+  public void testBatchStatements() throws Exception {
+    String sql =
+        "INSERT INTO users (id) VALUES (1); INSERT INTO users (id) VALUES (2);INSERT INTO users (id) VALUES (3);";
```
voulg, line 321: Also add a test for something like:

1. Double ';', e.g: SELECT 1;; SELECT 2;
Would this work? with the current code we might send an empty query, would the backend work with this?
2. ';' within comments
3. Perhaps at least one e2e test with real spanner

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
Vizerai: I added a check to remove comments.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
Vizerai: Not intentional. Removed now.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
Vizerai: Yes, this is only for testing. I added a note for that.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
Vizerai: I added some additional test for the other edge cases.

I have an e2e DML test and an e2e DDL test.

.ci/e2e-expected/ddl-batching.txt
Vizerai: They are fixed now.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
Vizerai: Done.

.ci/e2e-batching/ddl-batch.txt
```diff
@@ -0,0 +1 @@
+CREATE TABLE table1 (id integer PRIMARY KEY, data integer); CREATE TABLE table2 (id integer PRIMARY KEY, data integer); CREATE TABLE table3 (id integer PRIMARY KEY, data integer);
```
voulg, line 1: Can we also add some negative tests? e.g: SELECT statements in batch, DML + DDL used together and see what errors we return?

.ci/e2e-expected/dml-batching.txt
```diff
@@ -0,0 +1 @@
+INSERT 0 0
```
voulg: Should we return the total number of rows inserted? I think we can get this number from JDBC

src/main/java/com/google/cloud/spanner/pgadapter/utils/StatementParser.java
```diff
@@ -141,9 +141,6 @@ private static String removeCommentsAndTrimPostgres(String sql) {
       throw new IllegalArgumentException(
           "SQL statement contains an unterminated block comment: " + sql);
     }
-    if (res.length() > 0 && res.charAt(res.length() - 1) == SEMICOLON) {
```
voulg, line 144: Why do we need this? If I understand correctly this removes the last semicolon that should be safe to do? (and it might be likely to be the correct thing to do)

src/test/java/com/google/cloud/spanner/pgadapter/StatementParserTest.java
```diff
@@ -54,7 +54,7 @@ public void testRemoveCommentsAndTrim() {
             + "  INSERT VALUES (v.column1, v.column2, v.column3)\n"
             + "WHEN MATCHED\n"
             + "  UPDATE SET FirstName = v.column2,\n"
-            + "             LastName = v.column3";
+            + "             LastName = v.column3;";
```
voulg, line 57: Similar comment to above, are we sure this won't cause a problem?

src/main/java/com/google/cloud/spanner/pgadapter/utils/StatementParser.java
Vizerai: This was removed because the code now removes comments at the query processing phase which means the pre-created queries in our regexes would all need to be changed to remove the ending ';'.  It also doesn't need to be removed so I think it is better to just change the code here.

src/test/java/com/google/cloud/spanner/pgadapter/StatementParserTest.java
Vizerai: Sending the ';' with the query doesn't impact it.  I'm not sure why the comment stripping code removed it in the first place.

.ci/e2e-batching/ddl-batch.txt
Vizerai: I added additional tests.

.ci/evaluate-with-psql.sh
```diff
@@ -74,3 +74,24 @@ sed -i -E 's/[0-9]{16,}//' .ci/e2e-result/backslash-dn-param.txt
...
+RETURN_CODE=$((${RETURN_CODE}||$?))
+
+echo "------Test \"-c option ddl-dml mix batching\"------"
+/usr/lib/postgresql/"${PSQL_VERSION}"/bin/psql -h localhost -p 4242 -d "${GOOGLE_CLOUD_DATABASE_WITH_VERSION}" -c "$(echo $(cat .ci/e2e-batching/ddl-dml-batch.txt))" &> .ci/e2e-result/ddl-dml-batching.txt
+diff -i -w -s .ci/e2e-result/ddl-dml-batching.txt .ci/e2e-expected/ddl-dml-batching.txt
+RETURN_CODE=$((${RETURN_CODE}||$?))
+
+echo "------Test \"-c option SELECT batching\"------"
+/usr/lib/postgresql/"${PSQL_VERSION}"/bin/psql -h localhost -p 4242 -d "${GOOGLE_CLOUD_DATABASE_WITH_VERSION}" -c "$(echo $(cat .ci/e2e-batching/select-batch.txt))" &> .ci/e2e-result/select-batching.txt
+diff -i -w -s .ci/e2e-result/select-batching.txt .ci/e2e-expected/select-batching.txt
```
voulg: It is quite interesting this is an empty file. Should it be empty, or it would be better to explicitly throw an error that this is not supported?

Vizerai: No, I was not done editing that file.  I have to push my changes to run the e2e tests.  This produces an error as expected.

.ci/e2e-expected/dml-batching.txt
Vizerai: Added.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -165,13 +207,21 @@ public Exception getException() {
    *
    * @throws SQLException If an issue occurred in extracting result metadata.
    */
-  protected void executeHelper() throws SQLException {
+  protected void executeHelper(int[] updateCounts) throws SQLException {
```
voulg: This method has a quite generic name and it's unclear why it should get an updateCounts argument. The issue is probably pre-existing, but adding this argument here makes it worse.

One simple way that (imo) this would be more readable is if we update the this.updateCount at the callsite for batch statements and here we check if this.updateCount != null and use it directly, else we use this.statement.getUpdateCount()

Thoughts?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
```diff
@@ -31,11 +32,11 @@
 
   public QueryMessage(ConnectionHandler connection) throws Exception {
     super(connection);
+    String query = StatementParser.removeCommentsAndTrim(this.readAll());
```
voulg, line 35: Wouldn't this result in all queries we sent to the backend lack comments? I'd hope that we keep comments when it's possible to do so. Thoughts?

voulg (commented):
A few nits for discussion.

Can you also:
1. Add Sneha as a reviewer
2. Make sure we have alignment on the scope and caveats of the batching. I believe that the choices here make sense and we return appropriate errors, but it would be even better if we have alignment with Adam / Sneha.

.ci/e2e-batching/select-batch.txt
```diff
@@ -0,0 +1 @@
+SELECT * FROM users; INSERT INTO users (id, age, name) VALUES (99, 99, 'person'); SELECT * FROM users;
```
snehashah16: should this be renamed to select-dml batch ?

snehashah16: can u also add a test case where SELECT is not the first statement in the batch

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -192,9 +242,17 @@ protected void handleExecutionException(SQLException e) {
...
     this.executed = true;
+    int[] updateCounts = null;
     try {
-      this.statement.execute(this.sql);
-      this.executeHelper();
+      if (statements.size() > 1) {
+        for (String stmt : statements) {
+          this.statement.addBatch(stmt);
+        }
+        updateCounts = this.statement.executeBatch();
```
snehashah16, line 254: can this return an exception ? example, what if there is a commit message in here

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -300,4 +314,71 @@ public void testPortalStatementDescribePropagatesFailure() throws Exception {
...
+
+    Assert.assertTrue(intermediateStatement.isBatchedQuery());
+    List<String> result = intermediateStatement.getStatements();
+    Assert.assertEquals(result.size(), 2);
+    Assert.assertEquals(result.get(0), "INSERT INTO users (name) VALUES (';;test;;');");
+    Assert.assertEquals(result.get(1), "INSERT INTO users (name1, name2) VALUES ('''''', ';'';');");
+  }
+
+  @Test
+  public void testBatchStatementsWithComments() throws Exception {
```
snehashah16, line 374: * can we add connection statements (BEGIN; COMMIT) in this test suite too 
* can we also add tests to cover statements that start with "WITH" clause
* I assume we support CREATE | ALTER | DROP DDL statements

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
Vizerai: Both 'addBatch' and 'executeBatch' can return exceptions.  This is inside of a try/catch block.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
Vizerai: We could also just rename executeHelper to something like updateResultCount which better signifies what the intended purpose was.

I split them into 2 different methods. One for batched update and one for non-batched update.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
Vizerai: The comments will be stripped out anyways. I think this is ok. We need to do this for batch statements, otherwise ';' in comments will be parsed.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
Vizerai: - WITH expressions are not supported by the backend yet.
- BEGIN and COMMIT are not supported within Cloud Spanner batch statements.
- CREATE | ALTER | DROP are supported.

.ci/e2e-batching/select-batch.txt
Vizerai: Renamed.

Added another SELECT test.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
snehashah16: I think its important to test the "not supported" features to make sure we are doing the right behavior, i.e. throwing an error message that is meaningful to the customer

also, WITH is mainly to make sure the PgAdapter Parsing is handling it (not crashing/erroring in weird ways).

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
snehashah16: Its important to think of error responses and scenarios. A meaningful error message will help the customer, example: if it only says InvalidArg but its not clear to the customer since PGAdapter supports COMMIT as a single statement but fails in -c option now.

.ci/e2e-expected/select-ddl-batching.txt
```diff
@@ -0,0 +1 @@
+ERROR:  INVALID_ARGUMENT: The statement is not suitable for batching. Only DML and DDL statements are allowed for batching.
```
snehashah16: I think it will be helpful to add a little more clarity here.

Currently we only support a batch of either DDL or DML statements. List of non-supported statements include (but not. limited to) are SELECT, Connection statement and Instance/Database Admin statements. Other statements are still supported within single execution [feel free to change terminology here.. maybe call it non-batch or serial transactions].

.ci/e2e-expected/select-dml-batching.txt
```diff
@@ -0,0 +1 @@
+ERROR:  INVALID_ARGUMENT: The statement is not suitable for batching. Only DML and DDL statements are allowed for batching.
```
snehashah16: print the argument i.e. the exact statement that is causing the error

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -102,7 +102,7 @@ public void execute() {
     this.executed = true;
     try {
       ((PreparedStatement) this.statement).execute();
-      this.executeHelper();
+      this.updateResultCount(null);
```
snehashah16: not a fan of null.. can we either create a new method definition that doesnt take input or maybe use empty type ?

.ci/e2e-expected/select-ddl-batching.txt
snehashah16: it seems we also support a limited format for sql statements.. is tis documented anywhere ?
Can that be included in the error message from either of these:
* while parsing
* while executing ?


 STATEMENT_DELIMITER = ';';
 SINGLE_QUOTE = '\'';

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -300,4 +314,87 @@ public void testPortalStatementDescribePropagatesFailure() throws Exception {
...
+    Assert.assertEquals(result.size(), 3);
+    Assert.assertEquals(result.get(0), "INSERT INTO users (id) VALUES (1);");
+    Assert.assertEquals(result.get(1), "INSERT INTO users (id) VALUES (2);");
+    Assert.assertEquals(result.get(2), "INSERT INTO users (id) VALUES (3);");
+  }
+
+  @Test
+  public void testAdditionalBatchStatements() throws Exception {
+    String sql =
+        "BEGIN TRANSACTION; INSERT INTO users (id) VALUES (1); INSERT INTO users (id) VALUES (2); INSERT INTO users (id) VALUES (3); COMMIT;";
```
snehashah16, line 335: I assume tis will fail within JDBC because it wont accept BEGIN within PreparedStatement ?

snehashah16 (commented):
Can u also add tests for failed cases:
* CREATE database statement
* connection statements
* WITH statement

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
Vizerai: The error is not vague.  For BEGIN and COMMIT it will produce this error:
"ERROR:  INVALID_ARGUMENT: The statement is not suitable for batching. Only DML and DDL statements are allowed for batching."  I also appended the sql string to the error.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
Vizerai: Yes, this is only testing the parsing of the statements.  This wouldn't work in a real batch statement.

.ci/e2e-expected/select-ddl-batching.txt
Vizerai: I think it would be easier to include this in the README/documentation rather than the error message.  Otherwise the error message will be quite verbose.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
Vizerai: I added more tests for these other cases.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
Vizerai: I refactored this into a batch and non-batch version.  The null argument was removed.

.ci/e2e-expected/select-dml-batching.txt
Vizerai: I added the sql string so the user can see what caused the problem.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
snehashah16: discussed adding a section in README to provide all this context and info

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
snehashah16: add info in README

snehashah16 (approved):
lets make sure the changes for README explaining this feature are a fast follow

voulg (approved):
Let's get format approval from Adam and Sneha in the design doc section as well before submit.

PR #14, opened 2021-12-06
Adding JavaCC parser for COPY command.

Vizerai:


src/main/javacc/build.xml
```diff
@@ -0,0 +1,29 @@
...
+
+  <target name="copy" description="Builds 'copy parser'">
+    <delete dir="copy"/>
+    <mkdir dir="copy"/> 
+    <copy file="copy.jjt" todir="copy"/>
+    <copy file="ASTID.java" todir="copy"/>
+    <copy file="ASTColumnElement.java" todir="copy"/>
+    <copy file="ASTCopyOptionElement.java" todir="copy"/>
+    <copy file="ASTFilename.java" todir="copy"/>
+    <copy file="CopyDumpVisitor.java" todir="copy"/>
```
snehashah16: lets remove this if u remove the actual  file

src/main/javacc/CopyDumpVisitor.java
```diff
@@ -0,0 +1,105 @@
+package com.google.cloud.spanner.pgadapter.parsers;
+
+public class CopyDumpVisitor implements CopyVisitor
```
snehashah16: discussed removing this file in 1:1

pom.xml
```diff
@@ -89,6 +89,16 @@
       <artifactId>commons-cli</artifactId>
       <version>1.4</version>
     </dependency>
+    <dependency>
```
snehashah16, line 92: let's add unit tests when integrated with the existing code

snehashah16 (approved):

src/main/javacc/copy.jjt
```diff
@@ -0,0 +1,247 @@
...
+  <O_BINARY: "BINARY">
+| <O_CSV: "CSV">
+| <O_FORMAT: "FORMAT">
+| <O_FREEZE: "FREEZE">
+| <O_DELIMITER: "DELIMITER">
+| <O_NULL: "NULL">
+| <O_HEADER: "HEADER">
+| <O_QUOTE: "QUOTE">
+| <O_ESCAPE: "ESCAPE">
+| <O_FORCE_QUOTE: "FORCE_QUOTE">
```
tinaspark, line 62: some of these options, as well as COPY FROM <filename> won't be supported in the first version of COPY. Should we throw a parse exception here if those options are used or handle the exception later on?

Vizerai: I added exceptions for the unused options. Later we can add support for them if necessary.

src/main/javacc/CopyDumpVisitor.java
Vizerai: Done.

voulg (approved):

src/main/javacc/build.xml
Vizerai: Done.

src/main/javacc/ASTColumnElement.java
```diff
@@ -0,0 +1,25 @@
+package com.google.cloud.spanner.pgadapter.parsers;
+
+public class ASTColumnElement extends SimpleNode {
```
libingye816, line 3: Why do we manually add those AST classes? I found that there are a lot more AST classes referenced in the visitor but they are not here so why we have to treat those AST classes specially?

src/main/javacc/copy.jjt
```diff
@@ -0,0 +1,252 @@
...
+{
+  < #DIGIT: ["0"-"9"] >
+|
+  < #LETTER: ["_","a"-"z","A"-"Z"] >
+|
+  < #LOWER_CASE_LETTER: ["_","a"-"z"] >
+|
+  < #ID: <LETTER> (<LETTER>|<DIGIT>)*>
+|
+  < IDENTIFIER: (<K_QUOTE> <ID> <K_QUOTE> | <ID>)>
```
libingye816, line 88: Why is the identifier wrapped with a single quote? Also, why quoted identifier still has the restrictions that only digits and letters are allowed?

src/main/javacc/copy.jjt
```diff
@@ -0,0 +1,252 @@
...
+  | {} /* EMPTY */
+}
+
+/** Column List. */
+void columnList() #ColumnList : {}
+{
+    LOOKAHEAD(2)
+    "(" columnList() ")"
+  | LOOKAHEAD(2)
+    columnElem() "," columnList()
```
libingye816, line 129: I am trying to understand why we need this given that we already have L127. Does this imply that columns are not necessarily wrapped by parenthesis?

src/main/javacc/copy.jjt
```diff
@@ -0,0 +1,252 @@
...
+/** Copy From. */
+void copyFrom() #CopyFrom : 
+{
+  Object o;
+}
+{
+  <K_FROM>	{o = Boolean.TRUE;}
+  | <K_TO>	{
+      o = Boolean.FALSE;
+      throw new ParseException("COPY TO is not supported.");
```
libingye816, line 154: A little confused here. As Tina mentioned that COPY FROM is not supported, it looks like neither of them is supported.

libingye816 (commented):
Not very familiar with COPY work so I have a couple of questions. Thanks for the patience.

A general question: What are the command we are going to support in this PR? I noticed that not all COPY syntax is supported. Do we plan to support them in the future?

src/main/javacc/copy.jjt
```diff
@@ -0,0 +1,252 @@
...
+{
+  <K_WITH>
+  | {} /* EMPTY */
+}
+
+void copyOptions() #CopyOptions : {}
+{
+  LOOKAHEAD(2)
+  "(" copyOptionList() ")"
+  | copyOptionList()
```
libingye816, line 178: I am trying to understand why we need L178 given that we already have L177. Does this imply that copyOptions are not necessarily wrapped by parenthesis?

Vizerai: Yes, the ( ) are optional.

src/main/javacc/copy.jjt
Vizerai: The exception is only thrown on <K_TO>.

src/main/javacc/copy.jjt
Vizerai: Identifiers can be wrapped in ' ' marks.  Identifiers should also allow underscores '_'.

src/main/javacc/ASTColumnElement.java
Vizerai: I manually added ones that I wanted to give specific functionality to.

PR #15, opened 2021-12-13
Adding support for COPY command.

Vizerai:


snehashah16:
Hey Jim, This is a 68 file code change.. Is there any way we ca break this up and request reviews ?

Vizerai:
> Hey Jim, This is a 68 file code change.. Is there any way we ca break this up and request reviews ?

Yes, I'm currently breaking it up into some smaller PRs.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -314,10 +315,23 @@ public ConnectionMetadata getConnectionMetadata() {
     return connectionMetadata;
   }
 
+  public IntermediateStatement getActiveStatement() {
```
olavloite: I think that this method needs to be synchronized, as it is accessing the `activeStatementsMap`, which could be accessed by multiple threads simultaneously.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -314,10 +315,23 @@ public ConnectionMetadata getConnectionMetadata() {
...
+    return activeStatementsMap.get(this.connectionId);
+  }
+
+  public ConnectionStatus getStatus() {
+    return status;
+  }
+
+  public void setStatus(ConnectionStatus status) {
+    this.status = status;
```
olavloite, line 327: I don't know all the implementation details here, so please correct me if I'm wrong, but I think that writing to the `this.status` flag might also need synchronization. The field is marked as `volatile`, which indicates that there will be multi-threaded access to it. Is it safe to assume that only one thread will be calling this method?

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,267 @@
...
+import java.util.Map;
+import org.apache.commons.csv.CSVFormat;
+
+public class CopyStatement extends IntermediateStatement {
+
+  private static final String COLUMN_NAME = "column_name";
+  private static final String SPANNER_TYPE = "spanner_type";
+  private static final String CSV = "CSV";
+
+  protected String tableName;
```
olavloite: nit: Why are these declared as `protected` instead of `private`?

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,267 @@
...
+    this.connection = connection;
+    this.statement = connection.createStatement();
+  }
+
+  /** @return Mapping of table column names to column type. */
+  public Map<String, TypeCode> getTableColumns() {
+    return this.tableColumns;
+  }
+
+  protected void setTableColumns(Map tableColumns) {
```
olavloite: nit: Is there a specific reason why the argument is not declared as `Map<String, TypeCode>`?

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,267 @@
...
+  protected void setCopyColumnNames(List<String> copyColumnNames) {
+    this.copyColumnNames = copyColumnNames;
+  }
+
+  /** @return Format type specified in COPY statement, if provided. */
+  public String getFormatType() {
+    return this.formatType;
+  }
+
+  protected void setFormatType(String formatType) {
```
olavloite: (Here and for the other format-related properties)
Do we need a setter for this property (and why is it `protected`)? Could we instead remove the setter for this property and the other format-related properties, and make the properties final?

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,267 @@
...
+    return this.tableColumns;
+  }
+
+  protected void setTableColumns(Map tableColumns) {
+    this.tableColumns = tableColumns;
+  }
+
+  /** @return CSVFormat for parsing copy data based on COPY statement options specified. */
+  public CSVFormat getParserFormat() {
+    CSVFormat format = CSVFormat.POSTGRESQL_TEXT;
```
olavloite: Could we make the format properties such as `this.formatType` final, and only calculate the `CSVFormat` once in the constructor, instead of generating every time this getter is called? (See also below)

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  private void verifyCopyColumns() throws SQLException {
+    if (this.copyColumnNames.size() > this.tableColumns.size()) {
+      throw new SQLException("Number of copy columns provided exceed table column count");
+    }
+    LinkedHashMap<String, TypeCode> tempTableColumns = new LinkedHashMap<>();
+    // Verify that every copy column given is a valid table column name
+    for (String copyColumn : this.copyColumnNames) {
+      if (!this.tableColumns.containsKey(copyColumn)) {
```
olavloite: This seems to do a case-sensitive lookup of the column name, which means that the operation will be case-sensitive. Is that intentional?

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,267 @@
...
+        throw new SQLException(
+            "Column \"" + copyColumn + "\" of relation \"" + this.tableName + "\" does not exist");
+      }
+      // Update table column ordering with the copy column ordering
+      tempTableColumns.put(copyColumn, tableColumns.get(copyColumn));
+    }
+    this.tableColumns = tempTableColumns;
+  }
+
+  private static TypeCode parseSpannerDataType(String columnType) {
```
olavloite: This does not seem to support `NUMERIC`. Is that intentional? (It also does not support `JSON`, but as far as I know, that is also not supported on Spangres).

Would it maybe be better to change the implementation of this so that it:
1. Removes any optional length specification at the end of the type name (i.e. automatically strip `(100)` from `STRING(100)`.
2. Look up the `TypeCode` directly by looking for an enum constant with the same name as the type name without length?

If there are specific types that we do not want to support, then we could add specific code for that after the above mentioned lookup.

That would make this easier to maintain in case of new types.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,267 @@
...
+    ResultSet result =
+        this.connection
+            .createStatement()
+            .executeQuery(
+                "SELECT "
+                    + COLUMN_NAME
+                    + ", "
+                    + SPANNER_TYPE
+                    + " FROM information_schema.columns WHERE table_name = '"
+                    + this.tableName
```
olavloite: This lookup is case sensitive. Is that intentional?

nit: It should preferably use a query parameter instead of string concatenation.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,194 @@
+package com.google.cloud.spanner.pgadapter.utils;
```
olavloite, line 15: nit: missing copyright header

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+                + record.size());
+      }
+      WriteBuilder builder = Mutation.newInsertBuilder(this.tableName);
+      // Iterate through all table column to copy into
+      for (String columnName : this.tableColumns.keySet()) {
+        TypeCode columnType = this.tableColumns.get(columnName);
+        String recordValue = "";
+        try {
+          recordValue = record.get(columnName).trim();
+          switch (columnType) {
```
olavloite, line 98: Are we intentionally not supporting `NUMERIC` here?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+   * @return True if adding payload to current batch will fit under mutation limit and batch size
+   *     limit, false otherwise.
+   */
+  private boolean payloadFitsInCurrentBatch(int rowMutationCount, int payloadLength) {
+    return (this.mutationCount + rowMutationCount <= MUTATION_LIMIT
+        && this.batchSize + payloadLength <= BATCH_LIMIT);
+  }
+
+  /** @return list of CSVRecord rows parsed with CSVParser from CopyData payload byte array */
+  public List<CSVRecord> parsePayloadData(byte[] payload) throws IOException {
```
olavloite: nit: I think this can be private

src/main/javacc/copy.jjt
```diff
@@ -0,0 +1,251 @@
...
+| "\t"
+| "\n"
+| "\r"
+| <"//" (~["\n","\r"])* ("\n"|"\r"|"\r\n")>
+| <"/*" (~["*"])* "*" (~["/"] (~["*"])* "*")* "/">
+}
+
+TOKEN : /* KEYWORDS */
+{
+  <K_STDIN: "STDIN">
```
olavloite: This does not include the `PROGRAM` keyword that is supported from [PostgreSQL 9.3 and later](https://www.postgresql.org/docs/9.3/sql-copy.html). That seems like a valid choice, but I think it would be good to document that both for users and for ourselves.

olavloite (commented):
This PR does not contain any tests. Unit tests for this feature are probably difficult to implement (and probably not really add very much value), but would it be possible to add a system test for it?

src/main/javacc/copy.jjt
Vizerai: We are currently only supporting a subset of the COPY functionality.  I shared an internal design doc with details.

Vizerai:
I submitted the JavaCC Parser changes in another PR (https://github.com/GoogleCloudPlatform/pgadapter/pull/19) and rebased this one.  This made the PR significantly smaller.  68 files changed -> 11 after rebase.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: Done.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: This code was originally written by an intern during the summer of 2021.  I think NUMERIC was not supported at that time.  We could add it now.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: Added.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: Case sensitive table names are allowed.  I think it should be case sensitive here since this table name is read in form the user supplied query. 

I changed it to use query parameters.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: Changed to private.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
Vizerai: I made it synchronized.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
Vizerai: It is volatile, but it is not static.  Each connection (thread) will have it's own status.  The connectionHandler thread will be executing all of the code which then reads from this status variable.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: I removed all of the setters and made them private instead of protected.  We can't make them final because many of them are set when execute() is called and the user supplied statement is parsed.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: getParserFormat() is only called once when the mutationBuilder is built.  The getter getFormatType() just returns the already built formatType.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: The column names are taken from the query to information schema.  This will return case sensitive names.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
olavloite: Sorry, I should have been a little more specific: My worry regards the `this.copyColumnNames`. As far as I understand, those come from the `COPY` statement.

1. Assume that the user is trying to execute the following statement:
```sql
COPY Singers (SingerId, LastName) FROM 'some_file.csv'
```
2. The actual column names in the information schema are `singerid` and `lastname`

I think that would fail in this case.

Having this as a limitation (at least in this version) is probably OK, but we should include a hint in the error message if we do.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
olavloite: See also above: A table name in the PostgreSQL `COPY` command would normally not be case sensitive (unless it is quoted). I think it is OK to have this as a limitation in this version, but we should include a hint in the error message so the user knows what might be going on.

See https://www.postgresql.org/docs/9.4/sql-syntax-lexical.html for more information on identifier case-sensitivity in PostgreSQL. The TLDR version of the rules are:
1. Unquoted identifiers are case-insensitive. `SELECT * FROM FOO` is equivalent to `SELECT * FROM foo`.
2. Quoted identifiers are case-sensitive. `SELECT * FROM "FOO"` is selecting from a different table than `SELECT * FROM "foo"`.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  private void queryInformationSchema() throws SQLException {
+    Map<String, TypeCode> tableColumns = new LinkedHashMap<>();
+    PreparedStatement statement =
+        this.connection.prepareStatement(
+            "SELECT ?, ? FROM information_schema.columns WHERE table_name = '?'");
+    statement.setString(1, COLUMN_NAME);
+    statement.setString(2, SPANNER_TYPE);
+    statement.setString(3, this.getTableName());
```
olavloite: I don't think this works.

1. Column names cannot be specified as query parameters. This query will now return the literal strings 'COLUMN_NAME' and 'SPANNER_TYPE' instead of the actual column names and types.
2. The `'?'` is interpreted as a string literal and not as a query parameter.


```suggestion
            "SELECT column_name, spanner_type FROM information_schema.columns WHERE table_name = ?");
    statement.setString(1, this.getTableName());
```

Vizerai: I updated it.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.Map;
+import org.apache.commons.csv.CSVFormat;
+
+public class CopyStatement extends IntermediateStatement {
+
+  private static final String COLUMN_NAME = "column_name";
+  private static final String SPANNER_TYPE = "spanner_type";
+  private static final String CSV = "CSV";
+
+  private String tableName;
```
tinaspark: table, format, from, and delimiter are all in the CopyTreeParser.CopyOptions. And the version of format in CopyOptions is an enum (TEXT/BINARY/CSV), which is closer to what we want than a string.

Can we just use the CopyOptions instead of duplicating fields here?

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
tinaspark: Does the jjt parser handle identifiers in the same way as the PostgreSQL parser? If so, the table and column names should already be lowercased if there are no quotes and remain uppercase if there were quotes

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+
+  /** @return number of rows copied into Spanner */
+  public int getRowCount() {
+    return this.rowCount;
+  }
+
+  /** Build mutation to add to mutations list with data contained within a CopyData payload */
+  public void buildMutation(ConnectionHandler connectionHandler, byte[] payload) throws Exception {
+    List<CSVRecord> records = parsePayloadData(payload);
+    if (!records.isEmpty()
```
tinaspark, line 77: this version of MutationBuilder allows writes beyond the Spanner mutation limit. It will close out the current batch if we hit the mutation limit but it won't fail the entire transaction. If we exceed the mutation limit we'll want to rollback the entire transaction, including batches that have already been sent to Spanner. But this can be done in a separate CL

tinaspark (changes requested):
Can you break this PR down even further? There are still too many moving pieces. I set a breakdown for Kayla that should work for you too. For each PR, can you also include in the description which parts are copied as-is from Kayla's branch and what modifications were made? It will help me focus my review on the recent changes since anything copied from Kayla's branch was already reviewed by at least 2 people.

1. Parse the original COPY statement and have tests that all the options, table name, and column names are parsed correctly
2. Get the table data from INFORMATION_SCHEMA and parse the column types (https://github.com/googleapis/pgadapter-private/pull/27)
3. Add CopyInResponse (https://github.com/googleapis/pgadapter-private/pull/29)
4. Add the mutation builder that currently exists on the branch (https://github.com/googleapis/pgadapter-private/pull/29)
5. modify the mutation builder to return an error if the spanner mutation  limit is exceeded
6. Add CopyDone, CopyData, and CopyFail support (https://github.com/googleapis/pgadapter-private/pull/33)

Vizerai:
This has been broken up into several smaller PRs:

https://github.com/GoogleCloudPlatform/pgadapter/pull/19
https://github.com/GoogleCloudPlatform/pgadapter/pull/22
https://github.com/GoogleCloudPlatform/pgadapter/pull/24
https://github.com/GoogleCloudPlatform/pgadapter/pull/32

Vizerai:
This is no longer needed as all the changes were incorporated into other PRs.

PR #19, opened 2022-01-24
Adding generated JavaCC parser code.

Vizerai:
This change renames CopyTreeVisitor -> CopyTreeParser. It also adds several CopyOptions to the CopyTreeParser which can be used when parsing a query.

Everything under src/main/java/com/google/cloud/spanner/pgadapter/parsers/copy is generated code from the JavaCC parser.


olavloite (commented):
If I understand this correctly, this PR only contains code changes generated by JavaCC. That probably means that the code itself is not that interesting, but rather:

1. What is the input that is used to generate this code? Is it correct that it is generated based on the contents of https://github.com/GoogleCloudPlatform/pgadapter/blob/postgresql-dialect/src/main/javacc/copy.jjt ?
2. What version of PostgreSQL is used as a reference? (I noticed that the `FROM PROGRAM` option seems to be missing, which was introduced in [9.3](https://www.postgresql.org/docs/9.3/sql-copy.html))

Vizerai:
> If I understand this correctly, this PR only contains code changes generated by JavaCC. That probably means that the code itself is not that interesting, but rather:
> 
> 1. What is the input that is used to generate this code? Is it correct that it is generated based on the contents of https://github.com/GoogleCloudPlatform/pgadapter/blob/postgresql-dialect/src/main/javacc/copy.jjt ?
> 2. What version of PostgreSQL is used as a reference? (I noticed that the `FROM PROGRAM` option seems to be missing, which was introduced in [9.3](https://www.postgresql.org/docs/9.3/sql-copy.html))

Yes, the copy.jjt file is the main input used to generate the parser code.

The core reference is roughly PostgreSQL 14, but we are supporting a limited subset of COPY functionality.  I shared a design doc internally which describes it.  Currently we are only supporting COPY FROM STDIN.  More support could be added later.

tinaspark (approved):

PR #20, opened 2022-01-26
test: add integration tests using PG native JDBC driver

olavloite:
Adds integration tests that use the native PostgreSQL driver to connect
to PGAdapter, and tweaks the protocol response from PGAdapter to be in
line with what real PostgreSQL does. Notable changes to protocol:
1. standard_conforming_strings should return 'on' instead of 'true'
2. Describe Portal should respond with NoData if the portal does not
   contain any columns (i.e. it is an update count)

This change also fixes a wrong assumption in IntermediateStatement that
a zero update count means NO_RESULT. Zero update count means that no
rows were updated/deleted by the UPDATE or DELETE statement.


olavloite:
cc @voulg @tinaspark @snehashah16 @Vizerai 

I don't have access to the main repository, so I cannot add any of you as reviewers (and I have to create the PR from a fork). PTAL.

The tests are failing, as GitHub Actions cannot download the required dependencies, and does not have access to the credentials needed for integration tests.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -47,43 +51,23 @@
...
     this.statement = connection.createStatement();
-  }
-
-  protected IntermediateStatement() {
-    this.executed = false;
-    this.exception = null;
-    this.resultType = null;
-    this.hasMoreData = false;
-    this.statementResult = null;
-    this.updateCount = null;
```
olavloite, line 64: Removed, as this is a no-op (it just sets a number of properties to their default values)

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -47,43 +51,23 @@
...
-  }
-
-  /**
-   * Extracts what type of result exists within the statement. In JDBC a statement update count is
-   * positive if it is an update statement, 0 if there is no result, or negative if there are
```
olavloite, line 69: A zero update count does not mean that there is no result. It means that zero rows were updated/deleted.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/BootstrapMessage.java
```diff
@@ -102,7 +102,7 @@ public static void sendStartupMessage(
     new ParameterStatusResponse(output, "client_encoding".getBytes(), "utf8".getBytes()).send();
     new ParameterStatusResponse(output, "DateStyle".getBytes(), "ISO".getBytes()).send();
     new ParameterStatusResponse(output, "IntervalStyle".getBytes(), "iso_8601".getBytes()).send();
-    new ParameterStatusResponse(output, "standard_conforming_strings".getBytes(), "true".getBytes())
+    new ParameterStatusResponse(output, "standard_conforming_strings".getBytes(), "on".getBytes())
```
olavloite, line 105: See https://www.postgresql.org/docs/14/runtime-config-compatible.html

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/DescribeMessage.java
```diff
@@ -87,13 +87,21 @@ public void handleDescribePortal() throws Exception {
...
-              this.outputStream,
-              this.statement,
-              ((DescribePortalMetadata) this.statement.describe()).getMetadata(),
-              this.connection.getServer().getOptions(),
-              QueryMode.EXTENDED)
-          .send();
+      switch (this.statement.getResultType()) {
+        case UPDATE_COUNT:
+        case NO_RESULT:
+          new NoDataResponse(this.outputStream).send();
```
olavloite, line 93: PostgreSQL returns `NoData` as as response to a `Describe Portal` for an update count or other statement that does not return a result set.

src/test/java/com/google/cloud/spanner/pgadapter/PgAdapterTestEnv.java
```diff
@@ -105,30 +107,36 @@
...
+
   // Spanner URL.
   private URL spannerURL;
 
   // Log stream for the test process.
   private static final Logger logger = Logger.getLogger(PgAdapterTestEnv.class.getName());
 
-  // Utility class for setting up test connection.
-  private RemoteSpannerHelper spannerHelper;
-
```
olavloite, line 116: Removed, as it was never used

src/test/java/com/google/cloud/spanner/pgadapter/PgAdapterTestEnv.java
```diff
@@ -105,30 +107,36 @@
...
 
   public void setUp() throws Exception {
     spannerURL = new URL(getHostUrl());
     options = createSpannerOptions();
   }
 
-  public SpannerOptions spannerOptions() {
-    return options;
+  public Spanner getSpanner() {
+    if (spanner == null) {
```
olavloite, line 127: Create a single shared `Spanner` instance instead of returning `SpannerOptions`. The `Spanner` instance is closed when the test environment is cleaned up.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -145,9 +145,9 @@ public void testBasicNoResultStatement() throws Exception {
     Mockito.verify(statement, Mockito.times(1))
         .execute("UPDATE users SET name = someName WHERE id = -1");
     Assert.assertFalse(intermediateStatement.containsResultSet());
-    Assert.assertEquals((int) intermediateStatement.getUpdateCount(), -2);
```
olavloite, line 148: An actual update statement will never return -2 in real life.

olavloite:
Closing, replaced by #23

PR #22, opened 2022-01-27
Adding Copy parser unit tests.

Vizerai:


tinaspark (changes requested):
Can you add tests for identifier variations so we can measure how similar JJT is to the original PostgreSQL

For example:
- COPY USERS
- COPY UsErS
- COPY "Users"
- COPY "USERS"

and similar for column names

Vizerai:
> Can you add tests for identifier variations so we can measure how similar JJT is to the original PostgreSQL
> 
> For example:
> 
> * COPY USERS
> * COPY UsErS
> * COPY "Users"
> * COPY "USERS"
> 
> and similar for column names

I Added more tests.  It currently does not change the case of the input and just accepts it as is.  I could change it to always lower case it unless it is in quotes.

tinaspark:
> > Can you add tests for identifier variations so we can measure how similar JJT is to the original PostgreSQL
> > For example:
> > 
> > * COPY USERS
> > * COPY UsErS
> > * COPY "Users"
> > * COPY "USERS"
> > 
> > and similar for column names
> 
> I Added more tests. It currently does not change the case of the input and just accepts it as is. I could change it to always lower case it unless it is in quotes.

We need to make sure that we adhere to all of the PostgreSQL rules (https://www.postgresql.org/docs/13/sql-syntax-lexical.html). That includes lowercasing table names and column names that are not in quotes. This is so that everything lines up when we build the INFORMATION_SCHEMA query and verify the column names. According to the documentation we need to support escaped unicode characters for the table names and column names too -- but I'm tempted to say that's unsupported. There are also rules about the characters that are allowed, but I think we can skip verifying those rules because the INFORMATION_SCHEMA query will return empty results if the table name is invalid and the column name verification will fail if the column names are invalid

We also need to handle variations of COPY/copy/CoPy for all keywords.

For the constants, I think we mostly only support constants that are single characters and you already enforced that in the JJT parser so we don't need to do any additional verification/processing.

One other concern I have is cases where a namespace is provided. For example, COPY public."Users" from STDIN. It doesn't look like the JJT parser will handle this correctly.

src/main/javacc/CopyTreeParser.java
```diff
@@ -158,6 +154,16 @@ public Object visit(ASTColumnList node, Object data) {
   }
 
   public Object visit(ASTQualifiedName node, Object data) {
+    if(node.jjtGetNumChildren() > 1) {
```
tinaspark: can you make this a for loop so it's flexible for any number of children?

src/main/javacc/ASTID.java
```diff
@@ -7,12 +7,20 @@ public ASTID(int id) {
...
+  public void setName(String name) {
+    this.name = name;
   }
 
   public String getName() {
-    return name;
+    return this.name;
+  }
+
+  public String getFormattedName() {
```
tinaspark: can you do the lowercasing in the copy.jjt file instead? It looks like the parser there distinguishes between an identifier and a quoted identifier. So in line 259/265 where the identifier() function is implemented, you could do the lowercasing there

src/main/javacc/copy.jjt
```diff
@@ -41,39 +40,40 @@ SKIP :
 
 TOKEN : /* KEYWORDS */
 {
-  <K_STDIN: "STDIN">
```
tinaspark, line 44: did you look into using the [IGNORE_CASE] qualifier here?

tinaspark (changes requested):

tinaspark:
> > > Can you add tests for identifier variations so we can measure how similar JJT is to the original PostgreSQL
> > > For example:
> > > 
> > > * COPY USERS
> > > * COPY UsErS
> > > * COPY "Users"
> > > * COPY "USERS"
> > > 
> > > and similar for column names
> > 
> > 
> > I Added more tests. It currently does not change the case of the input and just accepts it as is. I could change it to always lower case it unless it is in quotes.
> 
> We need to make sure that we adhere to all of the PostgreSQL rules (https://www.postgresql.org/docs/13/sql-syntax-lexical.html). That includes lowercasing table names and column names that are not in quotes. This is so that everything lines up when we build the INFORMATION_SCHEMA query and verify the column names. According to the documentation we need to support escaped unicode characters for the table names and column names too -- but I'm tempted to say that's unsupported. There are also rules about the characters that are allowed, but I think we can skip verifying those rules because the INFORMATION_SCHEMA query will return empty results if the table name is invalid and the column name verification will fail if the column names are invalid
> 
> We also need to handle variations of COPY/copy/CoPy for all keywords.
> 
> For the constants, I think we mostly only support constants that are single characters and you already enforced that in the JJT parser so we don't need to do any additional verification/processing.
> 
> One other concern I have is cases where a namespace is provided. For example, COPY public."Users" from STDIN. It doesn't look like the JJT parser will handle this correctly.

I talked to Adam and he's ok with us skipping unicode support as long as we return a good error message. So for the non-quoted identifiers, can you check if any begin with "U&" and if so, return an error that unicode identifiers are not supported

src/main/javacc/copy.jjt
Vizerai: Done.

src/main/javacc/ASTID.java
Vizerai: Done.

src/main/javacc/CopyTreeParser.java
Vizerai: Done.  This now accounts for multiple namespaces.

Vizerai:
> > > > Can you add tests for identifier variations so we can measure how similar JJT is to the original PostgreSQL
> > > > For example:
> > > > 
> > > > * COPY USERS
> > > > * COPY UsErS
> > > > * COPY "Users"
> > > > * COPY "USERS"
> > > > 
> > > > and similar for column names
> > > 
> > > 
> > > I Added more tests. It currently does not change the case of the input and just accepts it as is. I could change it to always lower case it unless it is in quotes.
> > 
> > 
> > We need to make sure that we adhere to all of the PostgreSQL rules (https://www.postgresql.org/docs/13/sql-syntax-lexical.html). That includes lowercasing table names and column names that are not in quotes. This is so that everything lines up when we build the INFORMATION_SCHEMA query and verify the column names. According to the documentation we need to support escaped unicode characters for the table names and column names too -- but I'm tempted to say that's unsupported. There are also rules about the characters that are allowed, but I think we can skip verifying those rules because the INFORMATION_SCHEMA query will return empty results if the table name is invalid and the column name verification will fail if the column names are invalid
> > We also need to handle variations of COPY/copy/CoPy for all keywords.
> > For the constants, I think we mostly only support constants that are single characters and you already enforced that in the JJT parser so we don't need to do any additional verification/processing.
> > One other concern I have is cases where a namespace is provided. For example, COPY public."Users" from STDIN. It doesn't look like the JJT parser will handle this correctly.
> 
> I talked to Adam and he's ok with us skipping unicode support as long as we return a good error message. So for the non-quoted identifiers, can you check if any begin with "U&" and if so, return an error that unicode identifiers are not supported

It now returns an exception if unicode characters are detected within the identifiers.

src/main/javacc/copy.jjt
```diff
@@ -255,7 +258,9 @@ void identifier() #ID :
 {
   t=<IDENTIFIER>
   {
-    jjtThis.setName(t.image.toLowerCase());
+    // Only support US_ASCII for now. UTF-8 support will be added later.
```
tinaspark, line 270: two requests:
1. can you explicitly check for escaped unicode -- anything that starts with U& or u&
2. can you explicitly set the error message to something like "Unicode identifiers are not supported" instead of relying on the StandardCharsets library

src/test/java/com/google/cloud/spanner/pgadapter/ParserTest.java
```diff
@@ -669,16 +670,26 @@ public void testCopyParserWithNamespace() throws Exception {
       assertThat(options.getFormat(), is(equalTo(Format.TEXT)));
     }
 
+    {
```
tinaspark, line 694: can you modify the test above to be public.\"USERS\" FROM STDIN

tinaspark (changes requested):

src/main/javacc/copy.jjt
Vizerai: Input form STDIN is not escaped.  Any strings with U& or u& will not be interpreted as unicode but just those characters instead.  The user would have to manually put the unicode characters into the input in which case the lexer would not match it as an identifier. We only allow ASCII at the moment.

snehashah16:
@tinaspark & @Vizerai 

Most comments here are related to input validations. While the client code can check for some, we need to create a fine balance between maintainability/duplicating v/s user experience.

If we create many client side checks, the code will have lots of duplicate efforts on server & client side. Though we dont want to create a bad user experience for the customer to wait (for the network round trip) to get trivial errors from their input from the server.

That said, our client libraries, drivers/ORMs and tools perform no client side check on the user input (i.e. table names, data, SQL syntax etc). Hopefully we will create the Emulator support to improve the user experience for PG so that development and tests can be performed locally. 
For the current scenario, where customers are using the PgAdapter for psql (which is interactive), we shouldn't be concerned about the latency impact to get errors from the server (instead of running too many client side checks).

src/test/java/com/google/cloud/spanner/pgadapter/ParserTest.java
Vizerai: I added some more tests which explicitly check that case was preserved on the quoted identifiers with namespaces.

src/main/javacc/copy.jjt
Vizerai: Done.

Added tests for this as well.

tinaspark:
> @tinaspark & @Vizerai
> 
> Most comments here are related to input validations. While the client code can check for some, we need to create a fine balance between maintainability/duplicating v/s user experience.
> 
> If we create many client side checks, the code will have lots of duplicate efforts on server & client side. Though we dont want to create a bad user experience for the customer to wait (for the network round trip) to get trivial errors from their input from the server.
> 
> That said, our client libraries, drivers/ORMs and tools perform no client side check on the user input (i.e. table names, data, SQL syntax etc). Hopefully we will create the Emulator support to improve the user experience for PG so that development and tests can be performed locally. For the current scenario, where customers are using the PgAdapter for psql (which is interactive), we shouldn't be concerned about the latency impact to get errors from the server (instead of running too many client side checks).

I agree with this observation for statements that will be sent to the server -- there is no need to do extensive checks in the client. However, this statement is not going to the server at all. It is only being parsed in PgAdapter. As a result, we need to be a lot more thorough with validation in the client. In this case, there is a tradeoff between cutting scope and extensive validation. That's why we cut unicode support. If we end up sending the COPY command to the server in the future then I anticipate that we'll drop the JJT parser and all of these checks and rely on the server for validation

tinaspark (approved):

PR #23, opened 2022-01-27
feat: support JDBC PreparedStatement

olavloite:
Adds integration tests that use the native PostgreSQL driver to connect
to PGAdapter, and tweaks the protocol response from PGAdapter to be in
line with what real PostgreSQL does. Notable changes to protocol:

1. standard_conforming_strings should return 'on' instead of 'true'
2. Describe Portal should respond with NoData if the portal does not
   contain any columns (i.e. it is an update count)

Adds support for decoding parameters that have been given in binary
format.

This change also fixes a wrong assumption in IntermediateStatement that
a zero update count means NO_RESULT. Zero update count means that no
rows were updated/deleted by the UPDATE or DELETE statement.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/parsers/BinaryParser.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/parsers/BinaryParser.java
@@ -32,6 +32,10 @@ public class BinaryParser extends Parser<byte[]> {
     this.item = (byte[]) item;
   }
 
+  public BinaryParser(byte[] item, FormatCode formatCode) {
+    this.item = item;
+  }
+
   @Override
   protected String stringParse() {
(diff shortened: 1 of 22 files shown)


olavloite:
Replaces #20

olavloite:
@Vizerai I don't know if you had already started looking into this, but I've broken it up into a number of smaller parts to make it a little easier:
- #27 Support binary parameter values
- #28 Add JDBC integration tests and make sure we can connect using the native PG JDBC driver
- #29 Determine statement type based on the actual SQL string

Once those are merged, I'll rebase this one so we can get in all the JDBC integration tests.

olavloite:
Closing as this has been replaced by a number of smaller PRs containing the same plus additional features.

PR #24, opened 2022-01-28
Adding in CopyStatement and MutationBuilder

Vizerai:


src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,216 @@
...
+  private char quoteChar;
+  private char escapeChar;
+  private boolean hasHeader;
+  private int formatCode;
+
+  private Map<String, TypeCode> tableColumns;
+  private MutationBuilder mutationBuilder;
+
+  public CopyStatement(String sql, Connection connection) throws SQLException {
+    super();
```
olavloite: nit: this is not necessary. A no-arg super constructor is called automatically if you leave it out

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,216 @@
+// Copyright 2021 Google LLC
```
olavloite: nit: 2022

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,216 @@
...
+    return this.tableColumns;
+  }
+
+  /** @return CSVFormat for parsing copy data based on COPY statement options specified. */
+  public CSVFormat getParserFormat() {
+    CSVFormat format = CSVFormat.POSTGRESQL_TEXT;
+    if (this.formatType != null && this.formatType.equalsIgnoreCase("CSV")) {
+      format = CSVFormat.POSTGRESQL_CSV;
+    }
+    if (this.nullString != null && !this.nullString.isEmpty()) {
```
olavloite: nit: You can use `com.google.common.base.Strings.isNullOrEmpty`

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,216 @@
...
+  private void queryInformationSchema() throws SQLException {
+    Map<String, TypeCode> tableColumns = new LinkedHashMap<>();
+    PreparedStatement statement =
+        this.connection.prepareStatement(
+            "SELECT "
+                + COLUMN_NAME
+                + ", "
+                + SPANNER_TYPE
+                + " FROM information_schema.columns WHERE table_name = ?");
+    statement.setString(1, this.tableName);
```
olavloite: Note that this requires that the table name is correctly handled by the parser and folded to lower case unless it was quoted:
1. Spangres will store the table name in lower-case, unless it was quoted when it was created.
2. If the user specifies the table name in the copy statement in upper case or mixed case without quoting, it will not be found by this query and cause an exception, unless the parser correctly applies PostgreSQL case folding of identifiers.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  @Override
+  public void execute() {
+    this.executed = true;
+    try {
+      parseCopyStatement();
+      mutationBuilder =
+          new MutationBuilder(this.tableName, this.tableColumns, getParserFormat(), hasHeader());
+      this.updateResultCount(); // Gets update count, set hasMoreData false and statement result
```
olavloite: nit: `this` is not needed

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+    this.hasHeader = hasHeader;
+    this.isHeaderParsed = false;
+    this.tableName = tableName;
+    this.tableColumns = tableColumns;
+    this.format = format;
+    this.mutations = new ArrayList<>();
+  }
+
+  /** @return list of CopyData mutations not yet written into Spanner */
+  public List<Mutation> getMutations() {
```
olavloite: Why do we want a public getter for this? It seems to me like this is internal state of this class.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+    return this.mutations;
+  }
+
+  /** @return number of rows copied into Spanner */
+  public int getRowCount() {
+    return this.rowCount;
+  }
+
+  /** Build mutation to add to mutations list with data contained within a CopyData payload */
+  public void buildMutation(ConnectionHandler connectionHandler, byte[] payload) throws Exception {
```
olavloite: If find the name of this public method slightly misleading. The method not only builds mutations, it also writes them to Spanner. Maybe rename to `writeMutations`?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+    return parser.getRecords();
+  }
+
+  /**
+   * Write list of mutations to Spanner and reset for next batch
+   *
+   * @return count of the number of rows updated.
+   */
+  public int writeToSpanner(ConnectionHandler connectionHandler) throws SQLException {
```
olavloite: Should this be public?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+    File unsuccessfulCopy = new File("./output.txt");
+    if (unsuccessfulCopy.createNewFile()) {
+      this.fileWriter = new FileWriter("output.txt");
+      writeToErrorFile(payload);
+    } else {
+      System.err.println("File " + unsuccessfulCopy.getName() + " already exists");
+    }
+  }
+
+  public void writeToErrorFile(byte[] payload) throws IOException {
```
olavloite: public?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+        connection.unwrap(CloudSpannerJdbcConnection.class);
+    spannerConnection.write(this.mutations); // Write mutation to spanner
+    // Reset mutations, mutation counter, and batch size count for a new batch
+    this.mutations = new ArrayList<>();
+    this.mutationCount = 0;
+    this.batchSize = 0;
+    return this.rowCount;
+  }
+
+  public void createErrorFile(byte[] payload) throws IOException {
```
olavloite: public?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+
+  public void writeToErrorFile(byte[] payload) throws IOException {
+    if (this.fileWriter != null) {
+      this.fileWriter.write(new String(payload, StandardCharsets.UTF_8).trim() + "\n");
+    }
+  }
+
+  public void closeErrorFile() throws IOException {
```
olavloite: public?

Vizerai: This will be accessed by the Copy messages (CopyDone, CopyData, ... not included in this PR).

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: This will be accessed by the CopyDone message (not included in this PR).

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: This will be accessed by the Copy messages (CopyDone, CopyData, ... not included in this PR).

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: This will be accessed by the CopyDone message (not included in this PR).

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: Renamed to MutationWriter.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
Vizerai: Removed for now.  I don't think this is needed.  Much of this code was original code that the intern wrote during the summer.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: I think we will enforce this in the Parser.  If not, I'll update this here.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -0,0 +1,203 @@
...
+  public int getRowCount() {
+    return this.rowCount;
+  }
+
+  /** Build mutation to add to mutations list with data contained within a CopyData payload */
+  public void buildMutation(ConnectionHandler connectionHandler, byte[] payload) throws Exception {
+    List<CSVRecord> records = parsePayloadData(payload);
+    if (!records.isEmpty()
+        && !payloadFitsInCurrentBatch(records.size() * records.get(0).size(), payload.length)) {
+      writeToSpanner(connectionHandler); // Close out the current batch and reset for the next batch
```
olavloite: This part does not feel right here. The method is called `buildMutation`. We also have a `writeToSpanner` method that is called by another class. That means that a call to `buildMutation` will potentially write some of the mutations to Spanner, but the other class also needs to call `writeToSpanner` to write mutations.

My understanding is that this implementation will not be supporting copy operations that exceed the standard Spanner transaction limits (20,000 mutations). Do we need this call to `writeToSpanner` here if we accept that the copy will fail if it exceeds the limit? If it is not needed, I would suggest that we remove this to avoid any confusion.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+    for (CSVRecord record : records) {
+      // Check that the number of columns in a record matches the number of columns in the table
+      if (record.size() != this.tableColumns.keySet().size()) {
+        writeToSpanner(connectionHandler);
```
olavloite: Same as above: Do we need this if we are currently not supporting copies that exceed the mutation limit?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -0,0 +1,203 @@
...
+              break;
+            case BYTES:
+              builder.set(columnName).to(Byte.parseByte(recordValue));
+              break;
+            case TIMESTAMP:
+              builder.set(columnName).to(com.google.cloud.Timestamp.parseTimestamp(recordValue));
+              break;
+          }
+        } catch (NumberFormatException | DateTimeParseException e) {
+          writeToSpanner(connectionHandler);
```
olavloite: Same as above: Do we need this if we are currently not supporting copies that exceed the mutation limit?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -0,0 +1,203 @@
...
+          createErrorFile(payload);
+          throw new SQLException(
+              "Invalid input syntax for type "
+                  + columnType.toString()
+                  + ":"
+                  + "\""
+                  + recordValue
+                  + "\"");
+        } catch (IllegalArgumentException e) {
+          writeToSpanner(connectionHandler);
```
olavloite: Same as above: Do we need this if we are currently not supporting copies that exceed the mutation limit?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -0,0 +1,203 @@
...
+          createErrorFile(payload);
+          throw new SQLException("Invalid input syntax for column \"" + columnName + "\"");
+        } catch (Exception e) {
+          writeToSpanner(connectionHandler);
```
olavloite: Same as above: Do we need this if we are currently not supporting copies that exceed the mutation limit?

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
olavloite: I actually meant the method, not the class, as I was slightly confused by all the `writeToSpanner` calls in the method. See my other comments in the last review regarding those.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationBuilder.java
olavloite: Ok, thanks for clarifying.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Removed.  I think we should rollback in the case that it cannot fit inside the mutation limit.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Removed.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Changed to rollback.  Errors should invalidate a COPY FROM (https://www.postgresql.org/docs/14/sql-copy.html).

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Changed to rollback.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Changed to rollback.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
olavloite: Sounds good, thanks.

olavloite (approved):
LGTM.

The e-to-e failures are caused by a new table that has been added to pg_catalog and is fixed in #26

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,203 @@
...
+
+  private void queryInformationSchema() throws SQLException {
+    Map<String, TypeCode> tableColumns = new LinkedHashMap<>();
+    PreparedStatement statement =
+        this.connection.prepareStatement(
+            "SELECT "
+                + COLUMN_NAME
+                + ", "
+                + SPANNER_TYPE
+                + " FROM information_schema.columns WHERE table_name = ?");
```
tinaspark: does the table name here preserve quotes? If not, can you add quotes in the INFORMATION_SCHEMA query? Then it should work for "foo" and "FOO"

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -0,0 +1,203 @@
...
+      updateResultCount(); // Gets update count, set hasMoreData false and statement result
+    } catch (Exception e) {
+      SQLException se = new SQLException(e.toString());
+      handleExecutionException(se);
+    }
+  }
+
+  private void parseCopyStatement() throws Exception {
+    try {
+      parse(sql, this.options);
```
tinaspark, line 210: can you turn the getParserFormat function into a setParserFormat function that is called here and sets the resulting value on this.format? It doesn't look like that member variable is being used at all right now. And then getParserFormat will just return the member variable

tinaspark (changes requested):
can you pull in the mutation  builder tests that Kayla added in https://github.com/googleapis/pgadapter-private/pull/32 and https://github.com/googleapis/pgadapter-private/pull/35 and the error file tests in https://github.com/googleapis/pgadapter-private/pull/37

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: Done.

tinaspark (approved):

PR #26, opened 2022-01-31
fix: return correct transaction status

olavloite:
PgAdapter should correctly return either 'I' (idle) or 'T' (transaction)
based on whether the session is still in a transaction. Some drivers
(psycopg) inspect and use the returned value to determine the state of the
connection.

This PR also updates the expected results of one of the psql e-to-e tests as `pg_catalog.pg_type` is now available and returned by the `\dt` command.


Vizerai:
This only applies to sync, query, and flush messages?

olavloite:
> This only applies to sync, query, and flush messages?

Yes, these are specifically the messages that could respond with a 'normal' `Ready` message and still be in a transaction. The startup message stream will also respond with a `Ready` message, but the session will never be in a transaction at that point, so we don't need to do anything with that.

There is however one other general case that needs to be considered, but that needs a little bit more discussion, and therefore kept out of this PR. Currently, all errors are followed by a `Ready` message with status `Idle`. That is not consistent with PostgreSQL, which will exhibit the following behavior after an **error in a transaction** (note: the current behavior of PgAdapter is correct for auto commit):
1. If the error occurred during a `Commit` / `Rollback` the server will respond with the error followed by status `Idle`. The transaction was ended, but without success.
2. If the error occurred for a normal statement during the transaction, `PostgreSQL` will respond with `Ready` and status `Error`. PostgreSQL will abort the current transaction if an error occurs during the transaction, and reject all subsequent statements on the transaction. The only way to recover from an error in a transaction is to either rollback the entire transaction or rollback to a previously set savepoint.

Point 1 above is trivial and does not need any special handling.

Point 2 shows an important difference between PostgreSQL and Cloud Spanner, as Cloud Spanner is able to continue a transaction after an error occurred. It would therefore not be a problem if PgAdapter would return `Ready` with status `Transaction` after an error occurred in a transaction. It would however not be semantically compatible with PostgreSQL, and would mean that application code would behave differently based on whether it is executed against Cloud Spanner or PostgreSQL.

Vizerai (approved):
LGTM

PR #27, opened 2022-02-02
feat: add support incoming binary values

olavloite:
Adds support for incoming binary values, such as statement parameters that are binary encoded by the client. This is required for native driver support (e.g. PG JDBC), as many drivers encode at least some parameter values using the binary format instead of text format.
This change also removes the requirement that all parameters **must** be typed, as some clients will send some parameter values untyped.


.github/workflows/ci.yaml
```diff
@@ -6,9 +6,9 @@ on:
 name: ci
 env:
   GOOGLE_CLOUD_PROJECT: "span-cloud-testing"
-  GOOGLE_CLOUD_INSTANCE: "spanner-testing"
+  GOOGLE_CLOUD_INSTANCE: "pgadapter-testing"
```
Vizerai: Why are these changing?  We should still be able to test with the staging endpoint.

olavloite: The build of this PR suddenly started failing on Friday for no apparent reason, and I was trying out a number of things to see if I could get it back to green. That didn't succeed, and it is probably caused by an issue on the staging env. I'll change it back as soon as it is clear exactly what is causing it (or whether it is unrelated).

olavloite: @Vizerai See #31 for a number of fixes to the CI build scripts. I think we need to change both the default test instance and the default endpoint, as:
1. The current test instance (`spanner-testing`) located in us-central1, and that region is no longer allow-listed for creating PG databases. Any attempt to create a PG database there will therefore return an INVALID_ARGUMENT error.
2. The staging endpoint can contain features that are not supported on production, and we should ensure that PgAdapter only uses features that are available on production. (See also email)

olavloite: Replaced by the changes in #31

Vizerai (approved):
LGTM

PR #28, opened 2022-02-02
feat: enable native JDBC connections

olavloite:
Enables connections to PgAdapter from the native PostgreSQL JDBC driver. This requires a couple of minor tweaks to the startup protocol:
1. Some older versions of the JDBC driver require the encoding values to be in upper case (i.e. `UTF8` instead of `utf8`).
2. The date style should also include the year-month-day order.
3. The value for `standard_conforming_strings` should be `on` or `off` (not `true` or `false`).


src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
```diff
@@ -0,0 +1,146 @@
...
+                .build()));
+  }
+
+  @Test
+  public void testSelectHelloWorld() throws SQLException {
+    try (Connection connection =
+        DriverManager.getConnection(
+            String.format("jdbc:postgresql://localhost:%d/", testEnv.getPort()))) {
+      try (ResultSet resultSet =
+          connection.createStatement().executeQuery("SELECT 'Hello World!'")) {
```
Vizerai, line 139: Will this work with all other queries as well using pg-jdbc?  I assume the wireprotocol is the same for pg-jdbc?

Vizerai (approved):
Overall LGTM.  Maybe we should add in more pg-jdbc testing, but it doesn't have to be done in this PR.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
olavloite: Yes, the wireprotocol is the same for pg-jdbc, so yes it will in theory work with other queries. However:
1. The pg-jdbc driver uses binary format for many query parameter types, so we need to support that for parameterized queries to work. That is included in #27 
2. The pg-jdbc driver sends `timestamp` and `date` parameter values as untyped parameters. Support for that is also included in #27 (only for timestamp at this moment, as we currently do not support date).
3. The pg-jdbc driver prefers the extended query mode over simple query mode (which is normally used in psql). That will probably also need some tweaking.

olavloite:
> Overall LGTM. Maybe we should add in more pg-jdbc testing, but it doesn't have to be done in this PR.

Yeah, I'll bring in a lot more test cases in separate PRs. #23 already contains a lot of test cases, but I decided to split it into several smaller PRs so it would be easier to review. Many of the test cases in #23 will only succeed once #27 has been merged, as the tests use parameterized queries and the pg-jdbc driver sends parameter values in binary format as opposed to the default text format used by psql.

PR #29, opened 2022-02-02
fix: command should determine result type

olavloite:
The result type was determined by looking at the update count returned
by the JDBC statement that was executed on Cloud Spanner. This is
however wrong, as:
1. An update count of 0 does not mean 'no result'. It means that zero
   rows were updated/deleted.
2. The result type is determined by the type of statement. A DML
   statement will always return an update count. A DDL statement will
   always return no result. A query will always return a result set.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
@@ -16,7 +16,6 @@ package com.google.cloud.spanner.pgadapter.statements;
 
 import com.google.cloud.spanner.Dialect;
 import com.google.cloud.spanner.connection.AbstractStatementParser;
-import com.google.cloud.spanner.jdbc.JdbcConstants;
 import com.google.cloud.spanner.pgadapter.metadata.DescribeMetadata;
 import com.google.common.base.Preconditions;
 import java.sql.Connection;
@@ -69,11 +68,11 @@ public class IntermediateStatement {
   }
 
   /**
-   * Determines the result type based on the given sql string. The sql string has already been
-   * stripped of
+   * Determines the result type based on the given sql string. The sql string must already been
+   * stripped of any comments that might precede the actual sql string.
    *
-   * @param sql
-   * @return
+   * @param sql The sql string to determine the type of result for
+   * @return The {@link ResultType} that the given sql string will produce
    */
   protected static ResultType determineResultType(String sql) {
     if (PARSER.isUpdateStatement(sql)) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -47,23 +51,39 @@
...
+    this.resultType = determineResultType(sql);
   }
 
-  protected IntermediateStatement() {
-    this.executed = false;
-    this.exception = null;
-    this.resultType = null;
-    this.hasMoreData = false;
-    this.statementResult = null;
-    this.updateCount = null;
```
olavloite, line 64: Removed as it is void. A constructor call will always automatically initialize properties to their default values (e.g. false for booleans, null for reference types, ...)

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -47,23 +51,39 @@
...
 
+  /**
+   * Determines the result type based on the given sql string. The sql string has already been
+   * stripped of
+   *
+   * @param sql
+   * @return
+   */
+  protected static ResultType determineResultType(String sql) {
+    if (PARSER.isUpdateStatement(sql)) {
```
olavloite, line 78: The type of statement (Query, DML, DDL) determines the type of result that will be returned.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/DescribeMessage.java
```diff
@@ -87,13 +87,21 @@ public void handleDescribePortal() throws Exception {
...
-              this.outputStream,
-              this.statement,
-              ((DescribePortalMetadata) this.statement.describe()).getMetadata(),
-              this.connection.getServer().getOptions(),
-              QueryMode.EXTENDED)
-          .send();
+      switch (this.statement.getResultType()) {
+        case UPDATE_COUNT:
+        case NO_RESULT:
+          new NoDataResponse(this.outputStream).send();
```
olavloite, line 93: This should return a `NoDataResponse` when a row description is requested for statements that return an update count (DML) or no result (DDL). This would previously return an empty `RowDescriptionResponse` for DML/DDL statements, which again would cause some clients (native PG JDBC driver) to return an error as it does not expect an empty `RowDescriptionResponse` to be returned for DML/DDL.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -14,6 +14,12 @@
 
 package com.google.cloud.spanner.pgadapter;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
```
olavloite, line 21: Added static imports for commonly used mock and assertion methods. This is the main reason behind the large number of changes in this file, and most changes are only replacements of `Mockito.when(...)` with `when(...)` etc.

Vizerai (approved):
LGTM

PR #30, opened 2022-02-04
feat: [WIP] support batching of all types of statements

olavloite:
Adds support for batching all types of statements together when using simple query mode. This is required for drivers that automatically batch multiple statements together, such as psycopg that tend to include transaction statements with the first actual statement, e.g.

```sql
BEGIN; SET TRANSACTION_ISOLATION_LEVEL TO SERIALIZABLE; SELECT FOO FROM BAR;
```


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -167,8 +167,10 @@ private void handleError(DataOutputStream output, Exception e) throws Exception
 
   /** Closes portals and statements if the result of an execute was the end of a transaction. */
   public void cleanUp(IntermediateStatement statement) throws Exception {
-    if (!statement.isHasMoreData() && statement.isBound()) {
-      statement.close();
+    for (int index = 0; index < statement.getStatementCount(); index++) {
+      if (!statement.isHasMoreData(index) && statement.isBound()) {
+        statement.close();
```
libingye816, line 172: Why this is inside the for loop? I don't understand the logic here.

src/test/java/com/google/cloud/spanner/pgadapter/AbstractMockServerTest.java
```diff
@@ -0,0 +1,245 @@
...
+                  .build())
+          .setMetadata(SELECT1_METADATA)
+          .build();
+  private static final com.google.spanner.v1.ResultSet SELECT2_RESULTSET =
+      com.google.spanner.v1.ResultSet.newBuilder()
+          .addRows(
+              ListValue.newBuilder()
+                  .addValues(Value.newBuilder().setStringValue("2").build())
+                  .build())
+          .setMetadata(SELECT1_METADATA)
```
libingye816, line 104: SELECT2_METADATA?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcTest.java
```diff
@@ -0,0 +1,158 @@
...
+        assertEquals(2, updateCounts.length);
+        assertEquals(1, updateCounts[0]);
+        assertEquals(2, updateCounts[1]);
+      }
+    }
+
+    // The PostgreSQL JDBC driver will send the DML statements as separated statements to PG when
+    // executing a batch using simple mode. This means that Spanner will receive two separate DML
+    // requests.
+    assertTrue(mockSpanner.getRequestsOfType(ExecuteBatchDmlRequest.class).isEmpty());
```
libingye816, line 119: simple mode?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcTest.java
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+
+    // The DML statements are split by the JDBC driver and sent as separate statements to PgAdapter.
+    // PgAdapter therefore also simply executes these as separate DML statements.
+    assertTrue(mockSpanner.getRequestsOfType(ExecuteBatchDmlRequest.class).isEmpty());
+    List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+    // We get three requests: One keep-alive and the two DML statements.
+    assertEquals(3, requests.size());
+    assertEquals(SELECT1.getSql(), requests.get(0).getSql());
```
libingye816, line 97: Where does select come from?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcTest.java
```diff
@@ -0,0 +1,158 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class JdbcTest extends AbstractMockServerTest {
```
libingye816, line 34: I think batching is only supported in simple mode? Does postgresql jdbc internally separate and execute statements individually in sequence?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
```diff
@@ -0,0 +1,361 @@
...
+        mockSpanner.getRequestsOfType(ExecuteBatchDmlRequest.class);
+    assertEquals(1, requests.size());
+    ExecuteBatchDmlRequest request = requests.get(0);
+    assertEquals(2, request.getStatementsCount());
+    assertEquals(INSERT_STATEMENT.getSql(), request.getStatements(0).getSql());
+    assertEquals(UPDATE_STATEMENT.getSql(), request.getStatements(1).getSql());
+  }
+
+  @Test
+  public void testJdbcBatch() throws SQLException {
```
libingye816, line 111: Looks like jdbc batch is not a real batching even in simple mode?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
```diff
@@ -0,0 +1,361 @@
...
+      }
+    }
+
+    // The server will also receive a 'SELECT 1' statement from PgAdapter, as PgAdapter calls
+    // connection.isAlive() before using it.
+    List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+    assertEquals(3, requests.size());
+    assertEquals(SELECT1.getSql(), requests.get(0).getSql());
+    assertEquals(SELECT2.getSql(), requests.get(1).getSql());
+    assertEquals(UPDATE_STATEMENT.getSql(), requests.get(2).getSql());
```
libingye816, line 205: Why ExecuteSqlRequest, shouldn't UPDATE use BatchDmlRequest?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -264,6 +299,78 @@ public void execute() {
...
+        statement.addBatch(statements.get(index));
+        index++;
+      } else {
+        // End the batch here, as the statement type on this index can not be batched together with
+        // the other statements in the batch.
+        break;
+      }
+    }
+    int[] updateCounts = this.statement.executeBatch();
+    updateBatchResultCount(fromIndex, updateCounts);
```
libingye816, line 361: This doesn't seem right in error handling as the statement results preceding the error is not updated.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -201,31 +228,41 @@ public Exception getException() {
     return e;
   }
 
+  private void setResultSet(int index, ResultSet resultSet) throws SQLException {
+    if (this.statementResults == null) {
+      this.statementResults = new ResultSet[getStatementCount()];
+    }
+    this.statementResults[index] = resultSet;
+    this.hasMoreData[index] = resultSet.next();
```
libingye816, line 236: Is this array really necessary?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -264,6 +299,78 @@ public void execute() {
...
+      if (canBeBatchedTogether(batchType, statementType)) {
+        statement.addBatch(statements.get(index));
+        index++;
+      } else {
+        // End the batch here, as the statement type on this index can not be batched together with
+        // the other statements in the batch.
+        break;
+      }
+    }
+    int[] updateCounts = this.statement.executeBatch();
```
libingye816, line 360: Looks like Connection.runBatch() only returns 0/1 not the actually number of updated rows.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
```diff
@@ -75,28 +75,30 @@ public IntermediateStatement getStatement() {
...
-      if (this.statement.containsResultSet()) {
-        new RowDescriptionResponse(
-                this.outputStream,
-                this.statement,
-                this.statement.getStatementResult().getMetaData(),
-                this.connection.getServer().getOptions(),
-                QueryMode.SIMPLE)
-            .send();
+    for (int index = 0; index < statement.getStatementCount(); index++) {
+      if (this.statement.hasException()) {
```
libingye816, line 82: How do we return statement results despite of errors? How do we provide enough context on which statements cause errors, which statements have results and which statements are not executed?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
```diff
@@ -0,0 +1,361 @@
...
+        statement.execute("COMMIT");
+      }
+    }
+
+    // The server will also receive a 'SELECT 1' statement from PgAdapter, as PgAdapter calls
+    // connection.isAlive() before using it. This should happen before the transaction is started.
+    List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+    assertEquals(3, requests.size());
+    assertEquals(SELECT1.getSql(), requests.get(0).getSql());
+    assertTrue(requests.get(0).getTransaction().hasSingleUse());
```
libingye816, line 271: What is this checking?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
```diff
@@ -0,0 +1,361 @@
...
+
+    // The first statement in the transaction will include the BeginTransaction option.
+    assertEquals(UPDATE_STATEMENT.getSql(), requests.get(1).getSql());
+    assertTrue(requests.get(1).getTransaction().hasBegin());
```
libingye816, line 275: same here

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
```diff
@@ -0,0 +1,361 @@
...
+    assertEquals(INSERT_STATEMENT.getSql(), requests.get(2).getSql());
+    assertTrue(requests.get(2).getTransaction().hasId());
```
libingye816, line 277: same here

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Seems like a mistake. I think that in some version of the code `statement.close()` took an index as an argument, or otherwise this was added as I was planning on adding index as an argument to `statement.close()`. But your current observation is right: This does not make sense in this way (but do note the `statement.hasMoreData(index)` call; it might very well be that in a final version something like this will be needed to close all resources that were opened by the batch)

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: Maybe... If we receive a batch containing multiple queries, then we need to keep track of each result set and whether it has more data or not. So assume that the batch contained the following:

```SQL
SELECT Value FROM KeyValue WHERE Key=1;
SELECT Value FROM KeyValue WHERE Key=2;
```

Then we will be returning two result sets. The `hasMoreData` field keeps track of whether a result set (a `Portal` in PostgreSQL terminology) has more rows or not. This field is however only really needed in Extended query mode. Batches are only supported in Simple query mode. So the answer to the question will depend on what makes it the easiest and most clear in the implementation.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: No, that is only true for DDL batches. DML batches return the actual update counts:

From the `Connection.runBatch()` documentation:
```
Returns:
the update counts in case of a DML batch. Returns an array containing 1 for each successful statement and 0 for each failed statement or statement that was not executed in case of a DDL batch.
```

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: Yes, that is a very good point. It is probably an improvement that is needed. If `Connection.runBatch()` fails, it will throw a `SpannerBatchUpdateException` that contains the update counts of the statements that succeeded, and the error for the statement that failed. That should be used to update the result counts of the statements that did succeed.

But more in general: Error handling is not really implemented in detail in this draft PR, so that is probably something you would want to look into.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
olavloite: We should send a result for each statement sequentially. So assume that you have the following batch:

```sql
insert into my_table (key, value) values (1, 'one');
insert into my_table (key, value) values (2, 'two');
select * from my_table;
insert into my_table (key, value) values (3, 'three');
insert into my_table (key, value) values (3, 'three'); -- This fails on a primary key constraint violation
insert into my_table (key, value) values (4, 'four'); -- This statement is never executed, as the previous failed
```

The results that we should send are:

```
INSERT 0 1
INSERT 0 1
SELECT 2 -- Plus the actual data, not shown in this example response
INSERT 0 1
ERROR: Primary key constraint violation. A row with the given identifier already exists.
```

So the user (or the client application) can see which statement failed based on how many statements first succeeded.

Note that the above is not a design choice on our side. This is specified by the PostgreSQL wire-protocol. This is also an error in the current implementation of batching in PGAdapter. Currently, if you execute the following batch:

```sql
insert into my_table (key, value) values (1, 'one');
insert into my_table (key, value) values (2, 'two');
```

PGAdapter will return:

```
INSERT 0 2
```

That is __incorrect__. It should return:

```
INSERT 0 1
INSERT 0 1
```


Additional note: The PostgreSQL wire protocol specifies that the result of a statement in simple query mode should be `command <count>`. That is why the select statement returns `SELECT 2`, as the command was `SELECT` and it selected 2 rows. One special case in the wire protocol is `INSERT`. Originally, the protocol was designed in such a way that `INSERT` should return the `Oid` of the table where the row was inserted, followed by the row count. That was dropped at a later moment, but is still in the protocol, and the server must return a zero instead. So all `INSERT` commands must respond with a zero between the `INSERT` and the row count.

src/test/java/com/google/cloud/spanner/pgadapter/AbstractMockServerTest.java
olavloite: We should change the name of the `SELECT1_METADATA` constant to something like `SELECT_INT64_METADATA`. The metadata for `SELECT 1` and `SELECT 2` are equal (and also for any other result set that contains a single `INT64` column), so we reuse it here.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
olavloite: Correct, the people who implemented the PostgreSQL JDBC driver chose to not use the PostgreSQL batching feature for this. The reason is _probably_ the intricate logic of implicit/explicit transactions in PostgreSQL batching, which does not match what the JDBC specification says about JDBC batches.

So:
1. If you use JDBC batching, even in simple mode, you will get sequential single statements. So we can't use that to test PostgreSQL batching.
2. If you set the JDBC driver in `preferQueryMode=simple`, you _can_ test PostgreSQL batching by using a SQL string with multiple statements separated with semicolons. So calling `connection.execute("insert into KeyValue values (1, 'One'); insert into KeyValue values (1, 'One'); ");` will use the PostgreSQL batching that we are interested in here.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
olavloite: No, only if we have more than one DML statements in a row. A single DML statement should just use the normal `executeUpdate` method in the Connection API, which again calls `ExecuteSql`. In this case we have a batch that looks like this:

```sql
SELECT 2;
UPDATE FOO SET BAR=1 WHERE BAZ=2;
```

That should result in the following requests:

```
ExecuteStreamingSql("SELECT 2");
ExecuteSql("UPDATE FOO SET BAR=1 WHERE BAZ=2");
```

However, if we had executed two DML statements after the SELECT, then the two DML statements should be combined into a BatchDML call:

```sql
SELECT 2;
UPDATE FOO SET BAR=1 WHERE BAZ=2;
INSERT INTO FOO VALUES (2, 2);
```

Should lead to the following sequence of requests on the Spanner server:

```
ExecuteStreamingSql("SELECT 2");
ExecuteBatchDml("UPDATE FOO SET BAR=1 WHERE BAZ=2", "INSERT INTO FOO VALUES (2, 2)");
```

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
olavloite: This checks what (type of) transaction the statement uses. `hasSingleUse()` returns true if the request uses a single-use transaction. This means that the statement 'has its own transaction', so kind of an auto-commit like feature (in this case it is a read-only single use transaction).

This specific `SELECT 1` statement comes from the way PGAdapter used to open connections to Cloud Spanner. PGAdapter would first open a connection and then execute a `SELECT 1` statement to verify that the connection is valid. This has been removed since this code was written, as it is inefficient to do so for every connection that we open. Instead, the current implementation of PGAdapter relies on the automatic dialect detection in the client library to check that a connection is valid. This happens automatically in the background, and only the first time a connection to a specific database is opened.

The reason that this select statement should use a single-use read-only transaction is that it is not part of the transaction that is initiated by the batch (it is not at all related to the batch).

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
olavloite: (See also my comment above regarding the `SELECT 1` statement, and remember that the `SELECT 1` statement is something that was executed internally by PGAdapter, and is not part of the batch that we are actually executing in this test)

This is a very good question, and I'll try to explain it using the specific example, as it is a little complex and requires some knowledge of how the Java client library and Connection API has been implemented.

This test executes the following series of statements.

```sql
statement.execute("BEGIN; UPDATE FOO SET BAR=1 WHERE BAZ=2;");
statement.execute("INSERT INTO FOO VALUES (1)");
statement.execute("COMMIT;");
```

The Java client library and the Connection API use what is called 'inlined `BeginTransaction`'. This means that the request to begin a read/write transaction will be inlined (included) with the first statement in the transaction, instead of being executed as a separate `BeginTransaction` RPC. So this means:
1. The `BEGIN;` statement is internally in the Connection API only translated to setting a flag that the application has started a transaction. It does __not__ execute any RPCs.
2. The `UPDATE FOO SET BAR=1 WHERE BAZ=2;` is the first statement in the read/write transaction. The request will therefore also carry an additional option to begin a new read/write transaction, and to use that transaction for the statement. The statement will return two things:
2.1. The ID of the transaction that was started. This is set as a property on the connection internally in the Connection API and is not returned to the client application.
2.2. The update count of the statement. This is returned to the client application.
3. The `INSERT INTO FOO VALUES (1)` will use the transaction ID that was returned by step 2.
4. The `COMMIT;` will commit the transaction that was started by step 2.

So:
1. `assertTrue(requests.get(1).getTransaction().hasBegin());` verifies that the `UPDATE` statement really included a `BeginTransaction` option in the request.
2. `assertTrue(requests.get(2).getTransaction().hasId());` verifies that the `INSERT` statement uses an existing transaction ID (the transaction ID that was returned by the `UPDATE` statement).

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeTest.java
olavloite: (See above)

src/test/java/com/google/cloud/spanner/pgadapter/JdbcTest.java
olavloite: > I think batching is only supported in simple mode?

Correct.

> Does postgresql jdbc internally separate and execute statements individually in sequence?

In most cases: Yes.

1. As you already noticed: PostgreSQL batching is only supported in simple mode. The JDBC driver by default uses extended mode. So everything the PostgreSQL driver does in the default extended mode will use individual statements that are executed sequentially.
2. As you also already noticed in a different comment: If you set the JDBC driver in simple mode and then execute a JDBC batch (by calling something like `statement.addBatch(String sql);`), the JDBC driver will __also execute the JDBC batch as sequential individual statements__.
3. The only way to execute PostgreSQL batches using the JDBC driver is if you:
3.1. Set the JDBC driver in simple mode.
3.2. AND execute a statement containing multiple statements, e.g. `statement.execute("insert into foo values (1); insert into foo values (2)");`

src/test/java/com/google/cloud/spanner/pgadapter/JdbcTest.java
olavloite: PGAdapter used to execute a keep-alive (or maybe a better name: check-if-valid) query everytime a new connection was opened. This query was executed internally in PGAdapter, which is why you don't see it in the sequence of statements in the test.

Note however: In the current version of PGAdapter this query has been removed. Instead, PGAdapter internally verifies that the connection is valid using the automatic dialect detection that was built in to the client library very recently, and you don't have to worry about it any more in these tests.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcTest.java
olavloite: Good catch. That is a typo. The JDBC driver uses extended mode. But regardless of whether the JDBC driver is in simple or extended mode: JDBC batches will always execute sequential statements, which by PGAdapter also will be executed as separate `ExecuteSql` requests.

olavloite:
Closing, no longer relevant

PR #31, opened 2022-02-06
fix: CI integration tests ignored environment variable values

olavloite:
The CI integration tests failed for a number of reasons (sometimes randomly):
1. The spanner-testing instance is located in us-central1 which is no longer allow-listed for creating PG databases. The default instance has therefore been adjusted to one in us-east1.
2. The staging endpoint currently returns an INVALID_ARGUMENT error for all CreateDatabase requests. The default endpoint has therefore been adjusted to the production endpoint.
3. Changing the values for the above two in the environment variables for the CI script did not have any effect, as the Maven command executing the tests did not take these into account. The integration tests expect the values to be given as system properties, and not environment variables.
4. The generated database ids could be invalid, as they could be longer than the max allowed 30 characters, and they could end with a hyphen or an underscore.
5. The integration tests would assign a random port to use for the PgAdapter proxy server. This was however just a random number within the range of valid port numbers, but without a check that the port was actually available. This could cause random failures if the randomly chosen port was already used by another service on the host machine. Instead, the proxy server can now be started with port number 0, which means that the OS should choose a random __available__ port number.


src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -149,15 +153,15 @@ void runServer() throws IOException {
       logger.log(
           Level.INFO,
           "Socket exception on port {0}: {1}. This is normal when the server is stopped.",
-          new Object[] {this.options.getProxyPort(), e});
+          new Object[] {getLocalPort(), e});
```
olavloite, line 156: Log the actual port instead of the configured port, as this will be a randomly assigned port if the user has specified port number 0 (zero).

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -91,8 +91,12 @@ public void startServer() {
           String.valueOf(this.options.getProxyPort()));
       this.status = ServerStatus.STARTING;
       start();
-      logger.log(
-          Level.INFO, "Server started on port {0}", String.valueOf(this.options.getProxyPort()));
+      // TODO: Consider changing the server to use ApiService, so we don't have to manage the state
+      //  of the server manually.
+      while (this.status == ServerStatus.STARTING) {
```
olavloite, line 96: We should wait until the server has actually started before logging the port that is being used, and before returning so the caller is not trying to connect to the server before it has actually started.

.github/workflows/ci.yaml
```diff
@@ -6,9 +6,9 @@ on:
 name: ci
 env:
   GOOGLE_CLOUD_PROJECT: "span-cloud-testing"
-  GOOGLE_CLOUD_INSTANCE: "spanner-testing"
+  GOOGLE_CLOUD_INSTANCE: "pgadapter-testing"
```
olavloite, line 9: The `spanner-testing` instance is in us-central1. That region is no longer allow-listed as a region where PG databases may be created, and will cause an `INVALID_ARGUMENT` error to be returned when we try to create a database. `pgadapter-testing` is located in us-east1 which is allowed to create PG databases.

.github/workflows/ci.yaml
```diff
@@ -6,9 +6,9 @@ on:
...
   GOOGLE_CLOUD_DATABASE: "testdb_e2e_psql"
-  GOOGLE_CLOUD_ENDPOINT: "staging-wrenchworks.sandbox.googleapis.com"
+  GOOGLE_CLOUD_ENDPOINT: "spanner.googleapis.com"
```
olavloite, line 11: We should not use the staging endpoint anymore, for two reasons:
- Currently, the staging endpoint returns an `INVALID_ARGUMENT` for multiple admin calls (this is temporary, and so not a permanent reason not to use it)
- The staging environment can contain features that will not be available on the production endpoint, and we should ensure that PgAdapter only uses features that are actually available on production.

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -206,6 +210,13 @@ public int getNumberOfConnections() {
     return this.handlers.size();
   }
 
+  public int getLocalPort() {
+    if (this.serverSocket != null) {
+      return this.serverSocket.getLocalPort();
```
olavloite, line 215: Return the actual port that has been assigned to the socket. As we now allow port number 0 (zero) to be configured, the port that is used by the socket can have been dynamically assigned by the OS.

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
```diff
@@ -49,7 +49,7 @@
       "gcpga -p <project> -i <instance> -d <database> -c <credentials_file>";
   private static final String OPTION_HELP = "h";
   private static final String DEFAULT_PORT = "5432";
-  private static final int MIN_PORT = 1, MAX_PORT = 65535;
+  private static final int MIN_PORT = 0, MAX_PORT = 65535;
```
olavloite, line 52: Allow port 0 to be used to let the OS assign a random available port number.

src/test/java/com/google/cloud/spanner/pgadapter/ITParameterizedQueryTest.java
```diff
@@ -66,20 +67,24 @@ public void setUp() throws Exception {
...
-      "-i",
-      testEnv.getInstanceId(),
-      "-d",
-      db.getId().getDatabase(),
-      "-c",
-      testEnv.getCredentials(),
-      "-s",
-      String.valueOf(testEnv.getPort()),
-      "-e",
-      "staging-wrenchworks.sandbox.googleapis.com"
```
olavloite, line 81: Remove hardcoded endpoint

src/test/java/com/google/cloud/spanner/pgadapter/ITParameterizedQueryTest.java
```diff
@@ -66,20 +67,24 @@ public void setUp() throws Exception {
...
+                testEnv.getProjectId(),
+                "-i",
+                testEnv.getInstanceId(),
+                "-d",
+                db.getId().getDatabase(),
+                "-s",
+                String.valueOf(testEnv.getPort()),
+                "-e",
+                testEnv.getUrl().getHost());
+    if (credentials != null) {
```
olavloite, line 84: Do not require credentials to be specified in the arguments. This enables the usage of the default credentials of the environment.

src/test/java/com/google/cloud/spanner/pgadapter/PgAdapterTestEnv.java
```diff
@@ -151,10 +151,8 @@ public String getCredentials() {
   }
 
   public int getPort() {
-    if (port == 0) {
-      Random rand = new Random(System.currentTimeMillis());
-      int defaultPort = rand.nextInt(10000) + 10000;
```
olavloite, line 156: Do not use random number as port number, as there is no guarantee that this port number is not used by some other service on the host machine, even if it is a port number higher than 10,000. Instead we should use port number 0, as that instructs the OS to assign a randomly chosen __available__ port number.

src/test/java/com/google/cloud/spanner/pgadapter/PgAdapterTestEnv.java
```diff
@@ -184,8 +182,19 @@ public String getDatabaseId() {
...
       databaseId = System.getProperty(TEST_DATABASE_PROPERTY, DEFAULT_DATABASE_ID);
     }
-    return String.format(
-        "%s_%d_%d", databaseId, System.currentTimeMillis(), new Random().nextInt(Short.MAX_VALUE));
+    String id =
+        String.format(
+            "%s_%d_%d",
+            databaseId, System.currentTimeMillis(), new Random().nextInt(Short.MAX_VALUE));
+    // Make sure the database id is not longer than the max allowed 30 characters.
+    if (id.length() > 30) {
```
olavloite, line 190: The randomly chosen database ID could be invalid depending on the length of the configured prefix. If the prefix was long, the addition of a random number could push the total length of the id beyond the allowed max of 30 characters.

.ci/run-with-credentials.sh
```diff
@@ -39,7 +39,7 @@ clirr)
   mvn clirr:check
   ;;
 integration)
-  mvn verify -Dclirr.skip=true -DskipITs=false
+  mvn verify -Dclirr.skip=true -DskipITs=false -DPG_ADAPTER_HOST="https://${GOOGLE_CLOUD_ENDPOINT}" -DPG_ADAPTER_INSTANCE="${GOOGLE_CLOUD_INSTANCE}" -DPG_ADAPTER_DATABASE="${GOOGLE_CLOUD_DATABASE}"
```
olavloite, line 42: The CI script sets a number of environment variables that should be used as the endpoint, instance, database etc. Those environment variables were not used by the integration tests, as the integration tests expect those to be given as system properties and not as environment variables. This ensures that the values that have been configured are actually fed in to the integration tests as system properties.

.github/workflows/ci.yaml
Vizerai: Shouldn't it be in nam7-preview?

I believe the list of pg compatible allocs is
nam7-preview, europe-north1, us-west2, asia-east2

Vizerai (approved):
LGTM - but we should check the alloc for the test instance.  I think we want nam7-preview (see comment).

.github/workflows/ci.yaml
olavloite: According to the UI these are the allowed regions:

image

olavloite:
> LGTM - but we should check the alloc for the test instance. I think we want nam7-preview (see comment).

I'll merge this for now to unblock the other PRs. We can always revisit the exact location for the test instance in a separate PR.

PR #32, opened 2022-02-08
Adding Copy messages and CopyInResponse.

Vizerai:


src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -188,6 +188,11 @@ private void queryInformationSchema() throws SQLException {
     }
   }
 
+  @Override
+  public void handleExecutionException(SQLException e) {
+    super.handleExecutionException(e);
+  }
```
olavloite, line 194: nit: this method can be removed in its entirety, as it only calls the super implementation

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/CopyInResponse.java
```diff
@@ -0,0 +1,81 @@
+// Copyright 2021 Google LLC
```
olavloite: nit:
```suggestion
// Copyright 2022 Google LLC
```

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDoneMessage.java
```diff
@@ -26,15 +30,26 @@
...
-        "Spanner does not currently support the copy functionality through the proxy.");
+    // If backend error occurred during copy-in mode, drop any subsequent CopyDone messages.
+    MutationWriter mw = this.statement.getMutationWriter();
+    if (!statement.hasException()) {
+      int rowCount = mw.writeToSpanner(this.connection); // Write any remaining mutations to Spanner
+      statement.addUpdateCount(rowCount); // Increase the row count of number of rows copied.
+      this.sendSpannerResult(this.statement, QueryMode.SIMPLE, 0L);
+      new ReadyResponse(this.outputStream, ReadyResponse.Status.IDLE).send();
+    } else {
+      mw.closeErrorFile();
```
olavloite, line 49: Shouldn't we send a `ReadyResponse` here as well? If I understand the implementation correctly, we abort the copy operation if an error occurs, and do not expect any more Copy messages, which means that we should be in the `Idle` state.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyFailMessage.java
```diff
@@ -26,18 +31,28 @@
...
   @Override
   protected void sendPayload() throws Exception {
-    throw new IllegalStateException(
-        "Spanner does not currently support the copy functionality through the proxy.");
+    // If backend error occurred during copy-in mode, drop any subsequent CopyFail messages.
+    if (!statement.hasException()) {
+      new ErrorResponse(this.outputStream, new Exception(this.errorMessage), State.IOError).send();
+      new ReadyResponse(this.outputStream, ReadyResponse.Status.IDLE).send();
+    } else {
+      MutationWriter mw = this.statement.getMutationWriter();
```
olavloite, line 50: Same as above: Shouldn't we send a `ReadyResponse` in this case as well? Put another way: What is the state of the connection if we reach this point, and what is the next message that we expect from the client?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -314,10 +315,23 @@ public ConnectionMetadata getConnectionMetadata() {
     return connectionMetadata;
   }
 
+  public synchronized IntermediateStatement getActiveStatement() {
+    return activeStatementsMap.get(this.connectionId);
+  }
+
+  public synchronized ConnectionStatus getStatus() {
```
tinaspark: these aren't being used. are they for a future PR?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -314,10 +315,23 @@ public ConnectionMetadata getConnectionMetadata() {
...
+    return status;
+  }
+
+  public synchronized void setStatus(ConnectionStatus status) {
+    this.status = status;
+  }
+
   /** Status of a {@link ConnectionHandler} */
-  private enum ConnectionStatus {
+  public enum ConnectionStatus {
```
tinaspark: why does this need to be public?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDataMessage.java
```diff
@@ -29,19 +32,32 @@
...
   protected void sendPayload() throws Exception {
-    throw new IllegalStateException(
-        "Spanner does not currently support the copy functionality through the proxy.");
+    // If backend error occurred during copy-in mode, drop any subsequent CopyData messages.
+    MutationWriter mw = this.statement.getMutationWriter();
+    if (!statement.hasException()) {
+      try {
+        mw.buildMutation(this.connection, this.payload);
+      } catch (SQLException e) {
+        statement.handleExecutionException(e);
```
tinaspark, line 56: If we don't write to the error file here, couldn't we potentially lose rows? Although, I guess it's risky to write because we could also end up with  duplicate rows depending on whether the error occurred before or after the rows were sent to Spanner

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -161,6 +161,10 @@ public Integer getUpdateCount() {
     return this.updateCount;
   }
 
+  public void addUpdateCount(int count) {
+    this.updateCount += count;
```
tinaspark, line 165: is the updateCount ever used for anything? It seems like we're doing a lot of work to manage it but I  don't see how it's used

tinaspark (changes requested):

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
Vizerai: This was used in a later change I added.  I'll remove it from this PR.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
Vizerai: Needed for a later change.  I'll revert for now.

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/CopyInResponse.java
Vizerai: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: The super version is protected and not public.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
Vizerai: This value is returned to the user at the end showing how many rows were updated. It is used by all Query operations, not just Copy.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDoneMessage.java
Vizerai: Yes, we should send a ReadyResponse in this case.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyFailMessage.java
Vizerai: Yes, I think you are correct.  We should send a ReadyResponse in this case as well.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDataMessage.java
Vizerai: We could write to the error file here on an exception if we wait until CopyDone to actually write to spanner.  Everything should be rolled back on a failure/error, so I think it is best to wait for CopyDone to do the write/commit.

olavloite (approved):

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDataMessage.java
tinaspark: agreed. The current implementation that has sendPayload inside of CopyDataMessage is based on the stretch  goal that we plan to revert. I think this looks good for now. In the next CL let's make sure to get all of the non-stretch goal logic correct.

1. we will no longer track the 64KB batch size. Instead, we'll track the 100MB mutation API commit size limit and continue tracking the 20K mutation count limit. If either is exceeded, CopyDataMessage should write everything to the error file and return an error
2. In CopyDone, if none of the limits have been exceeded, we'll write to Spanner. Note that this can still fail because the mutation count inside of PGAdapter is an approximation and is not accurate if the table has indexes or is interleaved with other tables. If the write to Spanner fails, CopyDone should write everything to the error file and return an error

tinaspark (approved):

PR #35, opened 2022-02-11
fix: support null parameters

olavloite:
- Adds more integration tests using the native PG JDBC driver.
- Adds support for null values in query parameters.

The PG JDBC driver sends DATE/TIMESTAMP parameters with type code
Oid.UNSPECIFIED. Untyped NULL values are currently not supported by the
Spangres backend, and also not by the Java client library / JDBC driver.
A patch for the Java client library has been submitted here: https://github.com/googleapis/java-spanner/pull/1680


Vizerai (approved):
LGTM

PR #36, opened 2022-02-14
fix: add support for arrays in ResultSets

olavloite:
Adds support for arrays in query result sets. Arrays are not yet implemented on the backend, so this feature is for now only tested against a mock Spanner server.


pom.xml
```diff
@@ -105,6 +105,20 @@
       <version>2.3.0</version>
     </dependency>
     <!-- Test dependencies -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-spanner</artifactId>
+      <version>6.17.4</version>
```
Vizerai, line 111: Don't we want the PG version of the library?
6.17.4-pg-SNAPSHOT

I don't believe the PG changes have been rolled into mainline yet.

olavloite: This is the test-lib that contains the mock Spanner server. That is (AFAIK) not included in the PG snapshot release. So this will only be used for test and starting the mock Spanner server.

That being said, the PG feature has now been [merged](https://github.com/googleapis/java-spanner/pull/1673) into the Java client library and released, so I'll update this PR to use that version instead.

olavloite: Hmm... I tried using the newest version of the Spanner testlib locally, but that does not work because of the API changes between 6.17.4 and 6.19.0. So we are better off using the 6.17.4 testlib here, and then update everything once the JDBC driver also has been updated to use 6.19.0 (or we have gotten rid of the JDBC driver altogether).

Vizerai: We pull in the jdbc library 2.5.7-pg-SNAPSHOT which then pulls in the pg version of the client library 6.17.4-pg-SNAPSHOT.

Won't this lead to some problems by importing both versions of the library?
6.17.4-pg-SNAPSHOT
6.17.4

olavloite: That's a good point. They are actually importing two different things:
1. The JDBC library indeed imports `spanner-6.17.4-pg-SNAPSHOT`. This is the main library (basically everything in `src/main/java` of the library).
2. This import has the tag `<type>testlib</type>` which means that it only imports the test code of the library (basically everything in `src/test/java` of the library)

The above allows us to reuse test infrastructure from the client library. As it is imported with `<scope>test</scope>` it will also not be included in the build of PgAdapter, so it is not shipped to customers.

We actually do the same in the JDBC driver itself:
1. This [line imports the Spanner client library](https://github.com/googleapis/java-spanner-jdbc/blob/0f5383262f08a81c768ad2a66b3f754e525aff12/pom.xml#L106).
2. This [line imports the Spanner client library testlib](https://github.com/googleapis/java-spanner-jdbc/blob/0f5383262f08a81c768ad2a66b3f754e525aff12/pom.xml#L157) using scope test.

Vizerai (approved):
LGTM

PR #37, opened 2022-02-16
Enforcing commit size and mutation count limits.

Vizerai:


src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -148,7 +154,7 @@ public void buildMutation(ConnectionHandler connectionHandler, byte[] payload) t
    */
   private boolean payloadFitsInCurrentBatch(int rowMutationCount, int payloadLength) {
     return (this.mutationCount + rowMutationCount <= MUTATION_LIMIT
-        && this.batchSize + payloadLength <= BATCH_LIMIT);
+        && this.batchSize + payloadLength <= COMMIT_LIMIT);
```
olavloite, line 157: This assumes that the PG wire protocol payload byte size will translate 1-to-1 to Spanner API commit size. I don't think that is correct. At the same time it is probably on the safe side, as a Spanner commit call will be larger than the PG wire protocol payload size. So if we hit this limit here, then we know for sure that it will fail on teh backend. But we should at least be aware that we could also get errors from the backend for commits that are larger than 100MB that were not caught by this check.

The same also applies to the mutation count. The mutation count that is calculated here only includes the columns in the COPY statement. If there is a secondary index on the columns where the data is being inserted, the actual mutation count will be larger. So also for this we could get an error from the backend.

olavloite (approved):

tinaspark (approved):

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Yes, it won't be 1:1, but this gives us an estimate of how close it is.

Yes, we still could get an error from the backend even though we didn't detect the limit being exceeded on the PGAdapter.

PR #38, opened 2022-02-16
refactor: replace JDBC with Connection API

olavloite:
Replaces the Cloud Spanner JDBC driver with the internal Connection API.
This reduces the number of times that we need to parse the SQL string
and translate between different types of parameters. It also enables us
to more easily make changes in the client library (Connection API) that
we can use for PgAdapter, as the Connection API is an internal API
without a strict contract on the interface.

@aseering I can't add you directly as a reviewer, as you are not listed as a member of the repository.


.ci/e2e-expected/ddl-dml-batching.txt
```diff
@@ -1 +1 @@
-ERROR:  INVALID_ARGUMENT: Mixing DML and DDL statements in a batch is not allowed. "CREATE TABLE testtable (id integer PRIMARY KEY, data integer); INSERT INTO testtable (id, data) VALUES (1, 2); INSERT INTO testtable (id, data) VALUES (3, 4);"
+ERROR: FAILED_PRECONDITION: Executing updates is not allowed for DDL batches. "CREATE TABLE testtable (id integer PRIMARY KEY, data integer); INSERT INTO testtable (id, data) VALUES (1, 2); INSERT INTO testtable (id, data) VALUES (3, 4);"
```
olavloite, line 1: The switch to the Connection API means that some error messages slightly change. Instead of parsing the command type ourselves in PGAdapter, we leave that to the Connection API. This means that we in some cases like this will be returning error messages from the Connection API instead of errors generated by PGAdapter itself after parsing the statement(s).

pom.xml
```diff
@@ -38,36 +38,6 @@
     developed for PostgreSQL to be used with Spanner.
   </description>
 
-  <!-- artifact repository -->
```
olavloite: This is no longer needed as we can use a public build of the Cloud Spanner client library.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -67,22 +69,40 @@
   private static AtomicInteger incrementingConnectionId = new AtomicInteger(0);
   private ConnectionMetadata connectionMetadata;
   private WireMessage message;
-  private Connection jdbcConnection;
+  private Connection spannerConnection;
```
olavloite, line 72: We'll be using a `com.google.cloud.spanner.connection.Connection` instance instead of a `java.sql.Connection`.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -67,22 +69,40 @@
...
-  ConnectionHandler(ProxyServer server, Socket socket) throws SQLException {
+  ConnectionHandler(ProxyServer server, Socket socket) {
     super("ConnectionHandler-" + CONNECTION_HANDLER_ID_GENERATOR.incrementAndGet());
     this.server = server;
     this.socket = socket;
     this.secret = new SecureRandom().nextInt();
-    this.jdbcConnection =
-        DriverManager.getConnection(server.getOptions().getConnectionURL(), server.getProperties());
+    String uri = server.getOptions().getConnectionURL();
+    if (uri.startsWith("jdbc:")) {
```
olavloite, line 80: This allows users to still specify a normal JDBC connection string in the options when starting PGAdapter, meaning that this change will not break any existing configurations in use by customers.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -67,22 +69,40 @@
...
+    ConnectionOptions connectionOptions = ConnectionOptions.newBuilder().setUri(uri).build();
+    this.spannerConnection = connectionOptions.getConnection();
     setDaemon(true);
     logger.log(
         Level.INFO,
         "Connection handler with ID {0} created for client {1}",
         new Object[] {getName(), socket.getInetAddress().getHostAddress()});
   }
 
+  private String appendPropertiesToUrl(String url, Properties info) {
```
olavloite, line 93: JDBC allows additional properties to be given in a `Properties` instance. That is not supported by the Connection API, but all properties that are supported by JDBC are also supported in the connection URL for the Connection API.

src/main/java/com/google/cloud/spanner/pgadapter/commands/ListCommand.java
```diff
@@ -50,10 +49,7 @@ public Pattern getPattern() {
 
   @Override
   public String translate() {
-    try {
-      return String.format(OUTPUT_QUERY, this.connection.getCatalog());
-    } catch (SQLException e) {
-      return String.format(OUTPUT_QUERY, "UNKNOWN");
-    }
+    // TODO: Pass in the database name, or add it to the Connection API surface and get it here.
+    return String.format(OUTPUT_QUERY, "");
```
olavloite, line 53: Note: `this.connection.getCatalog()` would always return an empty string when using the JDBC driver. This is therefore not a change in behavior.

See https://github.com/googleapis/java-spanner-jdbc/blob/f9daa19453b33252bf61160ff9cde1c37284ca2b/src/main/java/com/google/cloud/spanner/jdbc/JdbcConnection.java#L388

src/main/java/com/google/cloud/spanner/pgadapter/metadata/DescribePortalMetadata.java
```diff
@@ -14,12 +14,12 @@
...
 
-import java.sql.ResultSetMetaData;
+import com.google.cloud.spanner.ResultSet;
 
 /** Simple POJO to hold describe metadata, specific to portal describes. */
-public class DescribePortalMetadata extends DescribeMetadata<ResultSetMetaData> {
+public class DescribePortalMetadata extends DescribeMetadata<ResultSet> {
 
-  public DescribePortalMetadata(ResultSetMetaData metadata) {
+  public DescribePortalMetadata(ResultSet metadata) {
```
olavloite, line 22: The Spanner client library does not have a separate class for `ResultSetMetaData`. Instead the methods are defined directly on `ResultSet`.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
```diff
@@ -28,29 +32,66 @@
...
+      this.item = toList(item.getValue(position), this.arrayElementType.getCode());
+      this.isStringEquivalent = stringEquivalence(this.arrayElementType.getCode());
+    } else {
+      arrayElementType = null;
+      isStringEquivalent = false;
+    }
+  }
+
+  private List<?> toList(Value value, Code arrayElementType) {
+    switch (arrayElementType) {
```
olavloite, line 61: The switch to the Connection API means that we get more concise information about the type of the array. JDBC only has a generic `java.sql.Types#ARRAY` constant. Spangres returns more information, which we now can use here.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
```diff
@@ -84,38 +125,49 @@ private String stringify(String value) {
...
   protected byte[] binaryParse() {
+    if (this.item == null) {
+      return null;
+    }
     ByteArrayOutputStream arrayStream = new ByteArrayOutputStream();
     try {
       arrayStream.write(IntegerParser.binaryParse(1)); // dimension
       arrayStream.write(IntegerParser.binaryParse(1)); // Set null flag
-      arrayStream.write(IntegerParser.binaryParse(this.arrayType)); // Set type
+      arrayStream.write(IntegerParser.binaryParse(Parser.toOid(this.arrayElementType))); // Set type
```
olavloite, line 163: Translate Spanner array element type to a PostgreSQL OID

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePortalStatement.java
```diff
@@ -37,21 +37,21 @@
...
 
+  void setBoundStatement(Statement statement) {
+    this.statement = statement;
+  }
+
   public short getParameterFormatCode(int index) {
     if (this.parameterFormatCodes.size() == 0) {
       return 0;
-    } else if (this.parameterFormatCodes.size() == 1) {
+    } else if (index >= this.parameterFormatCodes.size()) {
```
olavloite, line 54: The PG wire format spec says that if a parameter has a higher index than the list of format codes, the default should be used (the first format code given).

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -78,34 +56,25 @@ private int parseType(byte[][] parameters, int index) throws IllegalArgumentExce
     if (this.parameterDataTypes.size() > index) {
       return this.parameterDataTypes.get(index);
     } else {
-      throw new IllegalArgumentException("PgAdapter does not support untyped parameters.");
+      return Oid.UNSPECIFIED;
```
olavloite, line 59: We must support untyped parameters. At least the JDBC driver will send some parameters untyped (notably all date/time parameters).

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -78,34 +56,25 @@ private int parseType(byte[][] parameters, int index) throws IllegalArgumentExce
...
 
   public List<Integer> getParameterDataTypes() {
     return this.parameterDataTypes;
   }
 
-  public void setParameterDataTypes(List<Integer> parameterDataTypes)
-      throws IllegalArgumentException {
+  public void setParameterDataTypes(List<Integer> parameterDataTypes) {
     this.parameterDataTypes = parameterDataTypes;
-    if (this.parameterDataTypes.size() != this.parameterCount) {
```
olavloite, line 92: Check removed as we need to support untyped parameters, which means that there could be more parameters than types.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -118,36 +87,22 @@ public void execute() {
...
-          "Wrong number of parameters for prepared statement, expected: "
-              + parameterCount
-              + " but got: "
-              + parameters.length);
-    }
+    Statement.Builder builder = Statement.newBuilder(sql);
     for (int index = 0; index < parameters.length; index++) {
       short formatCode = portal.getParameterFormatCode(index);
       int type = this.parseType(parameters, index);
-      for (Integer position : parameterIndexToPositions.get(index)) {
```
olavloite, line 145: We don't need to translate back-and-forth between positional parameters (?) and indexed parameters ($1, $2, ...) anymore, which removes the requirement for the translation map `parameterIndexToPositions`.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -35,27 +38,25 @@
...
   protected ResultSet statementResult;
   protected boolean hasMoreData;
-  protected Exception exception;
+  protected SpannerException exception;
   protected String sql;
   protected String command;
   protected boolean executed;
   protected Connection connection;
-  protected Integer updateCount;
+  protected Long updateCount;
```
olavloite, line 49: Spanner returns `Long` for update counts.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -170,8 +171,8 @@ public void setHasMoreData(boolean hasMoreData) {
     this.hasMoreData = hasMoreData;
   }
 
-  public Statement getStatement() {
```
olavloite, line 173: We don't create and hold on to a reference to a `Statement` anymore, as Cloud Spanner `Statement` instances are immutable and can only be used to execute one SQL string with one set of parameters. JDBC statements can be re-used to execute different SQL strings.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -237,21 +236,33 @@ protected void handleExecutionException(SQLException e) {
...
+          connection.startBatchDml();
+        } else if (resultType == ResultType.NO_RESULT) {
+          connection.startBatchDdl();
+        } else {
+          throw SpannerExceptionFactory.newSpannerException(
+              ErrorCode.INVALID_ARGUMENT, "Statement type is not supported for batching");
+        }
         for (String stmt : statements) {
-          this.statement.addBatch(stmt);
+          connection.execute(Statement.of(stmt));
```
olavloite, line 253: This automatically adds the statement to the current batch on the connection.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java
```diff
@@ -14,120 +14,26 @@
...
-
-  /**
-   * PostgreSQL parameters occur as $\\d+, whereas JDBC expects a question mark. For now, doing a
-   * simple iteration to correct. Taking into account both quoted strings and escape sequences.
-   *
-   * @param sql The PostgreSQL String.
-   * @return A {@link SQLMetadata} object containing both the corrected SQL String as well as the
-   *     number of parameters iterated.
-   */
-  public static SQLMetadata toJDBCParams(String sql) {
```
olavloite, line 38: This is no longer needed, as we can feed the SQL string with the PostgreSQL-style parameters directly to the Connection API.

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/DataRowResponse.java
```diff
@@ -48,20 +46,18 @@ public DataRowResponse(
...
     this.mode = mode;
     this.resultSet = statement.getStatementResult();
-    this.metadata = resultSet.getMetaData();
-    this.columns = new ArrayList<>(this.metadata.getColumnCount());
+    this.columns = new ArrayList<>(this.resultSet.getColumnCount());
     this.length = HEADER_LENGTH + COLUMN_NUMBER_LENGTH;
     // TODO profile this with immense rows/column count to see if it's an aread of optimization.
-    for (int column_index = 1; /* column indices start at 1 */
-        column_index <= this.metadata.getColumnCount();
+    for (int column_index = 0; /* column indices start at 0 */
```
olavloite, line 52: JDBC starts counting at 1.
Connection API starts counting at 0.

Vizerai:
Copy changes have all been merged at this point.  You can go ahead and resolve the conflicts.

.ci/e2e-expected/ddl-dml-batching.txt
snehashah16: good thing this is done before GA now. Error Codes changes are considered breaking changes because customers can rely on these for appropriate handling within the application

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
snehashah16: do we need to support this ?

Yes I understand its a breaking change, but we can throw a pretty error to explain they need to drop that "jdbc" prefix. I rather not support something that doesn't make sense here and leave confusing tech debt

snehashah16: To make sure customers dont break, this change should be released with a major version change if we were GA. Though our PG releases AFAIR are still beta versioning (i.e. 0.*.*), so we should have the changelog specify this breaking change

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
snehashah16: same as above, if there is no documented "Properties" that we support, lets avoid tech debt to support something that PGAdapter is not concerned with (i.e. JDBC interface)

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -169,12 +172,18 @@ void runServer() throws IOException {
...
     ConnectionHandler handler = new ConnectionHandler(this, socket);
-    if (!handler.getJdbcConnection().isValid(0)) {
-      throw new SQLException("Invalid JDBC connection. Make sure credentials are valid.");
+    // Note: Calling getDialect() will cause a SpannerException if the connection itself is
+    // invalid, for example as a result of the credentials being wrong.
+    if (handler.getSpannerConnection().getDialect() != Dialect.POSTGRESQL) {
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.INVALID_ARGUMENT,
+          String.format(
+              "The database uses dialect %s. Only databases using dialect %s are allowed.",
```
snehashah16: maybe even adding more specifics,

Currently PGAdapter only suports connection to the PosgreSQL dialect database, that can be created using https://cloud.google.com/spanner/docs/quickstart-console#postgresql

olavloite:
> Copy changes have all been merged at this point. You can go ahead and resolve the conflicts.

@Vizerai I've merged in the most recent changes and resolved all conflicts.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
```diff
@@ -28,29 +32,66 @@
...
+        return value.getDateArray();
+      case JSON:
+        return value.getJsonArray();
+      case BYTES:
+        return value.getBytesArray();
+      case INT64:
+        return value.getInt64Array();
+      case STRING:
+        return value.getStringArray();
+      case NUMERIC:
```
snehashah16: should we error on non-PG types ?

snehashah16 (commented):
Sorry, I have to admit I havent reviewed every line changed 

I am curious, why are we supporting JDBC-interface sementics ?

olavloite:
> I am curious, why are we supporting JDBC-interface sementics ?

@snehashah16 The reasoning behind that is the following:
1. PGAdapter allows users to specify a JDBC connection URL for the Spangres database that it should connect to. This is important, as it allows users to specify additional configuration for the underlying connection. This could for example include a custom number of gRPC channels to use, min/max sessions, custom credentials, ... Replicating all of these configuration parameters in PGAdapter instead of just relying on the configuration options in the JDBC driver would introduce a lot of code duplication, and require extra work every time we add new configuration parameters to JDBC.
2. We use the JDBC connection URL and not the Connection API, because the JDBC is a public API and included in the public documentation. The Connection API is an internal API and not documented publicly. So by using the JDBC URL we can re-use the public documentation for the JDBC driver instead of having to duplicate the documentation specifically for PGAdapter, while it would essentially be exactly the same as for the JDBC driver.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: In addition to not breaking any existing configurations, I think it is better to support the `jdbc:` prefix in the connection string. That makes the documentation and user experience simpler, as we can just simply state something like this in the documentation: 

> Use a JDBC connection URL to specify additional connection properties that PGAdapter should use for the connection to Cloud Spanner. The documentation for the JDBC connection URL can be found here: https://cloud.google.com/spanner/docs/use-oss-jdbc

That way we don't have to confuse users with internals about the Connection API.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Sorry, my comment on this wasn't completely clear.

What I mean is that:
1. Customers can already specify a Cloud Spanner JDBC Connection URL and Properties that PGAdapter should use for its underlying connection.
2. The JDBC Connection URL and Properties are all publicly documented, so we can just reference that in the PGAdapter documentation to inform customers about what is supported.
3. The Connection API does not support a `Properties` object. That is a JDBC specific thing (but one that is commonly used by developers who work much with JDBC). However; everything that is supported in a JDBC Properties instance is supported by the Connection API in its connection URL. So this means that customers can use the commonly known (and publicly documented) connection properties of JDBC without having to learn about the internals of the Connection API, and PGAdapter will internally automatically translate those Properties into a valid Connection API connection URL.

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
olavloite: Done

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
olavloite: Yes, that seems better. Moved `NUMERIC` to the end to let it throw an error instead.

pom.xml
```diff
@@ -45,8 +45,8 @@
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
-      <artifactId>google-cloud-spanner-jdbc</artifactId>
-      <version>2.6.0</version>
+      <artifactId>google-cloud-spanner</artifactId>
+      <version>6.20.0</version>
```
thiagotnunes, line 53: Aside from this PR: we should configure `renovate-bot`

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
thiagotnunes: Curious: Is that allowed in pure PostgreSQL databases?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -67,22 +69,40 @@
...
+    if (info == null || info.isEmpty()) {
+      return url;
+    }
+    StringBuilder res = new StringBuilder(url);
```
thiagotnunes: nit: don't use abbreviations `res` -> `result`

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -67,22 +69,40 @@
...
+    for (Entry<Object, Object> entry : info.entrySet()) {
+      if (entry.getValue() != null && !"".equals(entry.getValue())) {
```
thiagotnunes, line 99: nit: can we use `StringUtils.isBlank`?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -130,11 +150,11 @@ public void run() {
...
-        if (this.jdbcConnection != null) {
-          this.jdbcConnection.close();
+        if (this.spannerConnection != null) {
+          this.spannerConnection.close();
         }
         this.socket.close();
-      } catch (SQLException | IOException e) {
+      } catch (SpannerException | IOException e) {
         logger.log(
             Level.WARNING, "Exception while closing connection handler with ID {0}", getName());
```
thiagotnunes: Should we log the exception stack trace / name?

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -169,12 +172,19 @@ void runServer() throws IOException {
    * Creates and runs the {@link ConnectionHandler}, saving an instance of it locally.
    *
    * @param socket The socket the {@link ConnectionHandler} will read from.
-   * @throws SQLException if the {@link ConnectionHandler} is unable to Connect.
+   * @throws SpannerException if the {@link ConnectionHandler} is unable to Connect.
```
thiagotnunes: Or if can not retrieve the dialect?

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
```diff
@@ -28,29 +32,66 @@
...
+        return value.getStringArray();
+      case PG_NUMERIC:
+        return value.getNumericArray();
+      case TIMESTAMP:
+        return value.getTimestampArray();
+      case FLOAT64:
+        return value.getFloat64Array();
+      case NUMERIC: // Only PG_NUMERIC is supported
+      case ARRAY:
+      case STRUCT:
```
thiagotnunes, line 82: Oh, isn't array of structs supported? It is in Spanner, but not yet in Spangres?

src/main/java/com/google/cloud/spanner/pgadapter/parsers/BinaryParser.java
```diff
@@ -61,16 +63,20 @@ public int getSqlType() {
...
   }
 
   @Override
   protected byte[] binaryParse() {
-    return this.item;
+    return this.item == null ? null : this.item.toByteArray();
+  }
+
+  public void bind(Statement.Builder statementBuilder, String name) {
+    statementBuilder.bind(name).to(this.item == null ? null : this.item);
```
thiagotnunes: nit: we can remove the ternary here

src/main/java/com/google/cloud/spanner/pgadapter/parsers/IntegerParser.java
```diff
@@ -14,16 +14,16 @@
...
 import java.sql.Types;
 import org.postgresql.util.ByteConverter;
 
 /** Translate from wire protocol to int. */
 public class IntegerParser extends Parser<Integer> {
 
-  public IntegerParser(ResultSet item, int position) throws SQLException {
-    this.item = item.getInt(position);
+  public IntegerParser(ResultSet item, int position) {
+    this.item = Math.toIntExact(item.getLong(position));
```
thiagotnunes, line 26: Did this throw an exception before? Or is this a breaking change (`toIntExact`)?
Just curious, as it would be fine since we are pre-GA

src/main/java/com/google/cloud/spanner/pgadapter/parsers/NumericParser.java
```diff
@@ -14,18 +14,24 @@
...
 import org.postgresql.util.ByteConverter;
 
 /** Translate from wire protocol to {@link Number}. */
 public class NumericParser extends Parser<Number> {
-  public NumericParser(ResultSet item, int position) throws SQLException {
+  public NumericParser(ResultSet item, int position) {
     // This should be either a BigDecimal value or a Double.NaN.
-    this.item = (Number) item.getObject(position);
+    Value value = item.getValue(position);
+    if (Value.NAN.equals(value.getString())) {
```
thiagotnunes: `equalsIgnoreCase`? Or are we sure about the casing here?

src/main/java/com/google/cloud/spanner/pgadapter/parsers/Parser.java
```diff
@@ -119,71 +122,128 @@ public static Parser create(byte[] item, int oidType, FormatCode formatCode) {
...
-      case Types.DOUBLE:
+      case FLOAT64:
         return new DoubleParser(result, columnarPosition);
-      case Types.BIGINT:
+      case INT64:
         return new LongParser(result, columnarPosition);
-      case Types.INTEGER:
-        return new IntegerParser(result, columnarPosition);
-      case Types.NUMERIC:
+      case NUMERIC:
```
thiagotnunes: I thought we weren't supporting NUMERIC?

src/main/java/com/google/cloud/spanner/pgadapter/parsers/TimestampParser.java
```diff
@@ -70,16 +70,17 @@ public TimestampParser(byte[] item, FormatCode formatCode) {
...
-          this.item = Timestamp.from(Instant.from(temporalAccessor));
+          this.item =
+              Timestamp.ofTimeSecondsAndNanos(
+                  temporalAccessor.getLong(ChronoField.INSTANT_SECONDS),
+                  temporalAccessor.get(ChronoField.NANO_OF_SECOND));
           break;
         case BINARY:
           long pgMicros = ByteConverter.int8(item, 0);
           com.google.cloud.Timestamp ts = com.google.cloud.Timestamp.ofTimeMicroseconds(pgMicros);
           long javaSeconds = ts.getSeconds() + PG_EPOCH_SECONDS;
```
thiagotnunes, line 81: Wondering if we tested with BC timestamps?

Also in some cases we can have negative seconds (timestamp earlier than epoch?)

src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java
thiagotnunes: Nice

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
```diff
@@ -166,24 +166,29 @@ public void testSelectHelloWorld() throws SQLException {
 
   @Test
   public void testSelectWithParameters() throws SQLException {
```
thiagotnunes, line 168: Why are we doing a JDBC test in pgAdapter (that is what I got from the test name, but maybe I am wrong)?

src/test/java/com/google/cloud/spanner/pgadapter/ITQueryTest.java
```diff
@@ -189,7 +189,7 @@ public void basicSelectTest() throws Exception {
     // Check exact message results.
     byte[] rowDescriptionData = {
       0, 4, 105, 100, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 20, 0, 8, 0, 0, 0, 0, 0, 0, 110, 97, 109, 101,
-      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, -1, -1, 0, 0, 0, 0, 0, 0, 97, 103, 101, 0, 0, 0, 0, 0, 0, 0,
+      0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 19, -1, -1, 0, 0, 0, 0, 0, 0, 97, 103, 101, 0, 0, 0, 0, 0, 0, 0,
```
thiagotnunes, line 192: Curious on why did this change

pom.xml
olavloite: Good point. I'll set that up separately.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: This URL references the Cloud Spanner / Spangres database on the backend that PGAdapter should use. And yes; this is allowed for Spangres databases.

It's not allowed for real PostgreSQL databases, but that is not what this connection URL is for.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
aseering: Yeah, not yet in Spangres.  PostgreSQL doesn't support structs (though it has a few similar concepts) so we're going to have to get creative to make it work.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Not really in this case. `entry.getValue()` is an `Object` and not a `String`.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Done

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Done

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
olavloite: Good point, done.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/BinaryParser.java
olavloite: Oops... Done.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/IntegerParser.java
olavloite: Yes, this did throw an exception before (in this specific case a `SQLException`). The Cloud Spanner JDBC driver checks whether the value can be cast/converted to the requested type internally before it is returned.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/NumericParser.java
olavloite: Good catch. No, we cannot be sure as this value could come from a client (application). Done.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/Parser.java
olavloite: Yeah, that's true, as we require the backing database to have PostgreSQL dialect. Change here and elsewhere.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/TimestampParser.java
olavloite: To be honest: We could probably write a million tests for PG timestamps, and I am absolutely sure that someone will still be able to break it.... My favorite piece of comment in real-world open source software used all around the world can be found here: https://github.com/pgjdbc/pgjdbc/blob/c02aa7725a3c5c71314e0e8c441a72fc97ff83a0/pgjdbc/src/main/java/org/postgresql/jdbc/TimestampUtils.java#L1083

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
olavloite: You assumption is correct, but what we are doing here is an end-to-end test using the native PG JDBC driver to ensure that PGAdapter actually behaves the same as PostgreSQL. So the test uses the native PG JDBC driver and connects to PGAdapter, which again is connected to Cloud Spanner. Queries/DML statements etc. are then sent from the native PG driver through PGAdapter to Spanner and then back through the same channel to verify that PGAdapter behaves the same as PostgreSQL would do.

So basically the same as our existing psql end-to-end tests, but now using the PG JDBC driver.

src/test/java/com/google/cloud/spanner/pgadapter/ITQueryTest.java
olavloite: It was wrong. This represents the `ResultSetMetaData` of a PG ResultSet, and there was a type mapping missing, which meant that the column was included in the metadata as `Oid.UNSPECIFIED` (zero), while it actually was a `Oid.VARCHAR` (1043). And as most text clients can handle types being unspecified, it went unnoticed for quite a while.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/TimestampParser.java
thiagotnunes: That is a good one indeed hehe.

pom.xml
thiagotnunes: Actually we might be better off using the BOM here (when you configure the renovate-bot).

thiagotnunes (approved):
Thanks for addressing the comments!

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -169,12 +172,20 @@ void runServer() throws IOException {
...
-      throw new SQLException("Invalid JDBC connection. Make sure credentials are valid.");
+    // Note: Calling getDialect() will cause a SpannerException if the connection itself is
+    // invalid, for example as a result of the credentials being wrong.
+    if (handler.getSpannerConnection().getDialect() != Dialect.POSTGRESQL) {
+      throw SpannerExceptionFactory.newSpannerException(
+          ErrorCode.INVALID_ARGUMENT,
+          String.format(
+              "The database uses dialect %s. Currently PGAdapter only supports connections to %s dialect databases. "
+                  + "These can be created using https://cloud.google.com/spanner/docs/quickstart-console#postgresql",
+              handler.getSpannerConnection().getDialect(), Dialect.POSTGRESQL));
```
Vizerai: Why is this an argument when it appears fixed? Why not just put "Dialect.POSTGRESQL" in the string?

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
```diff
@@ -28,29 +32,66 @@
...
-  private static String PG_ARRAY_OPEN = "{", PG_ARRAY_CLOSE = "}";
-  private static String SPANNER_ARRAY_OPEN = "[", SPANNER_ARRAY_CLOSE = "]";
+  private final Type arrayElementType;
+  private final boolean isStringEquivalent;
 
-  int arrayType;
-  boolean isStringEquivalent;
-
-  public ArrayParser(ResultSet item, int position) throws SQLException {
+  public ArrayParser(ResultSet item, int position) {
```
Vizerai, line 44: Shouldn't this still be marked as throwing an exception?

Vizerai (approved):
Overall LGTM.  Just a few minor comments.

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
olavloite: Good point, it's not like the enum constant is ever going to change. Done.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
olavloite: As we are now only throwing `SpannerException` instead of `SQLException` we do not need to declare the exception, as `SpannerException` is a `RuntimeException` instead of a checked `Exception`. Although it is possible to declare a method as throwing a `RuntimeException`, it is not common to do so. For example all methods in the Java client library could throw a `SpannerException`, but these are also not declared.

PR #40, opened 2022-02-17
feat: Add support for multiple CopyData messages

Vizerai:


Vizerai:
I will add some additional tests to check row splitting between multiple CopyData messages.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -167,7 +173,7 @@ private void queryInformationSchema() throws SQLException {
                 + COLUMN_NAME
                 + ", "
                 + SPANNER_TYPE
-                + " FROM information_schema.columns WHERE table_name = \"?\"");
+                + " FROM information_schema.columns WHERE table_name = ?");
```
tinaspark: why remove the quotes?

tinaspark (changes requested):
It's hard to me to track all the different things going on, why they depend on each other, and if all the changes are adequately tested. I'm seeing a few things here

1. Instead of parsing the CopyData payloads as they come in, put them all in a buffer and parse when we receive a CopyDone -- this looks reasonable to me and fits the goal of this PR
2. Move parseCommand from IntermediateStatement to StatementParser -- this looks reasonable but should probably in another PR
3. Error checking for COPY_IN in ControlMessage.java and some status handling in QueryMessage.java -- this should probably be in another PR

Vizerai:
I broke off the other changes into separate PRs.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
olavloite: (Sorry for wading in on the comment). This was missed in one of the previous rounds of reviews, so it got merged. As we use the Cloud Spanner JDBC driver internally in PgAdapter, we need to specify query parameters using JDBC-style parameters. The correct format is `?` and not `"?"`.

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
```diff
@@ -1202,16 +1202,21 @@ public void testCopyDataMessage() throws Exception {
     MutationWriter mb = Mockito.mock(MutationWriter.class);
     Mockito.when(copyStatement.getMutationWriter()).thenReturn(mb);
 
-    WireMessage message = ControlMessage.create(connectionHandler);
-    Assert.assertEquals(message.getClass(), CopyDataMessage.class);
-    Assert.assertArrayEquals(((CopyDataMessage) message).getPayload(), payload);
+    {
```
olavloite, line 1252: nit: remove
```suggestion
```

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
```diff
@@ -1202,16 +1202,21 @@ public void testCopyDataMessage() throws Exception {
...
+      WireMessage message = ControlMessage.create(connectionHandler);
+      Assert.assertEquals(message.getClass(), CopyDataMessage.class);
+      Assert.assertArrayEquals(((CopyDataMessage) message).getPayload(), payload);
 
-    CopyDataMessage messageSpy = (CopyDataMessage) Mockito.spy(message);
-    messageSpy.send();
+      CopyDataMessage messageSpy = (CopyDataMessage) Mockito.spy(message);
+      messageSpy.send();
+    }
```
olavloite, line 1275: nit: remove
```suggestion
```

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
```diff
@@ -1202,16 +1202,21 @@ public void testCopyDataMessage() throws Exception {
...
 
-    Mockito.verify(mb, Mockito.times(1)).buildMutation(connectionHandler, payload);
+    Mockito.verify(mb, Mockito.times(1)).addMutations(connectionHandler, payload);
   }
 
+  @Test
+  public void testMultipleCopyDataMessages() throws Exception {}
```
olavloite: I assume that this is a placeholder for a test that is coming in a next PR?

olavloite (approved):

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
tinaspark: I see. In that case, we need to make sure that setString will put quotes around the table name to preserve casing.

We already made sure that the parser handles casing correctly -- lowercase everything if there are no quotes, preserve casing if there are quotes. If we don't have quotes here the information_schema query will return the wrong results

olavloite: `setString` won't add quotes, but I'm pretty sure this will work as intended if the given table name is already in the correct case (i.e. the parser has already folded it to lower case or kept it in the original case if it was quoted). In this case the table name is just a normal query parameter that is compared to a normal string value in the `table_name` column in the `columns` table.

tinaspark: oh yes, you're right. It's a string here, not an identifier. I'll resolve the conversation

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
Vizerai: Updated now.

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
Vizerai: Removed.

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
Vizerai: Removed.

olavloite:
I think the current build errors might be caused by [b/217744569](http://b/217744569)

olavloite:
> I think the current build errors might be caused by [b/217744569](http://b/217744569)

That assumption was wrong. It's caused by another issue.

olavloite (approved):

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -134,27 +131,39 @@ public void buildMutation(ConnectionHandler connectionHandler, byte[] payload) t
...
+      throw new SQLException(
+          "Mutation count: " + mutationCount + " has exceeded the limit: " + MUTATION_LIMIT);
+    }
+  }
+
+  /**
+   * @return True if current payload will fit within COMMIT_LIMIT. This is only an estimate and the
+   *     actual commit size may still be rejected by Spanner.
+   */
+  private boolean payloadFitsInCurrentBatch() {
```
tinaspark: can we name this commitSizeIsWithinLimit to  match the naming of mutationCountIsWithinLimit? Also because the batching has been removed

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -72,25 +72,22 @@ public int getRowCount() {
...
 
-  /** Build mutation to add to mutations list with data contained within a CopyData payload */
-  public void buildMutation(ConnectionHandler connectionHandler, byte[] payload) throws Exception {
-    List<CSVRecord> records = parsePayloadData(payload);
-    if (!records.isEmpty()
-        && !payloadFitsInCurrentBatch(records.size() * records.get(0).size(), payload.length)) {
-      rollback(connectionHandler, payload);
-      long mutationCount = this.mutationCount + records.size() * records.get(0).size();
-      long commitSize = this.batchSize + payload.length;
+  public void addMutations(ConnectionHandler connectionHandler, byte[] payload) throws Exception {
```
tinaspark: can you name this addCopyData since we aren't doing anything with mutations yet

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -186,39 +195,32 @@ public int writeToSpanner(ConnectionHandler connectionHandler) throws SQLExcepti
...
-    if (unsuccessfulCopy.createNewFile()) {
-      this.fileWriter = new FileWriter(ERROR_FILE);
-      writeToErrorFile(payload);
-    } else {
-      System.err.println("File " + unsuccessfulCopy.getName() + " already exists");
-    }
+    this.fileWriter = new FileWriter(unsuccessfulCopy, false);
   }
 
   public void writeToErrorFile(byte[] payload) throws IOException {
```
tinaspark: rename to: writeCopyDataToErrorFile

add a comment that this is used if there are problems while collecting CopyData messages or building the mutation list

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -186,39 +195,32 @@ public int writeToSpanner(ConnectionHandler connectionHandler) throws SQLExcepti
...
-    if (this.fileWriter != null) {
-      this.fileWriter.write(new String(payload, StandardCharsets.UTF_8).trim() + "\n");
+    if (this.fileWriter == null) {
+      createErrorFile();
     }
+    this.fileWriter.write(new String(payload, StandardCharsets.UTF_8).trim() + "\n");
   }
 
   public void writeMutationsToErrorFile() throws IOException {
```
tinaspark: add a comment that this is used if Spanner returns an error when we try to commit the mutations

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
```diff
@@ -1209,7 +1210,75 @@ public void testCopyDataMessage() throws Exception {
...
+
+  @Test
+  public void testMultipleCopyDataMessages() throws Exception {
+    Mockito.when(connection.createStatement()).thenReturn(statement);
+    Mockito.when(connection.prepareStatement(ArgumentMatchers.anyString()))
+        .thenReturn(preparedStatement);
+    Mockito.when(connectionHandler.getJdbcConnection()).thenReturn(connection);
+    Mockito.when(statement.getUpdateCount()).thenReturn(1);
+
+    byte[] messageMetadata = {'d'};
```
tinaspark, line 1224: this test is sending multiple copy data messages, but each message has a full row of data, which was already supported before this PR.

Can you modify the test to split a single row over multiple CopyData messages. For example, the first payload will be "1\t'o" and the second payload will be "ne'\n"

tinaspark (changes requested):

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Done.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Done.

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
Vizerai: The tests splits the data across rows now.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Done.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Done.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -203,21 +203,30 @@ public void rollback(ConnectionHandler connectionHandler) throws Exception {
...
+   */
+  public void writeCopyDataToErrorFile(byte[] payload) throws IOException {
     if (this.fileWriter == null) {
       createErrorFile();
     }
     this.fileWriter.write(new String(payload, StandardCharsets.UTF_8).trim() + "\n");
   }
 
+  /**
+   * The list of mutaitons will be written to the error file if a problem was encountered while
```
tinaspark: mutations

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
tinaspark: actually, I'm thinking about this again and why are we writing the mutations to the file at all? The original intention of the error file was for the user to look through the copy data that they sent to PGAdapter, fix it, and try again. I think we should always be writing the payload data, not the mutation data

tinaspark (changes requested):

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Reworded comment.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Changed to write the copy data instead.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -224,17 +224,11 @@ public void writeCopyDataToErrorFile(byte[] payload) throws IOException {
...
    */
-  public void writeMutationsToErrorFile() throws IOException {
-    if (this.fileWriter == null) {
-      createErrorFile();
-    }
-
-    for (Mutation mutation : this.mutations) {
-      this.fileWriter.write(mutation.toString());
-    }
+  public void writeCopyDataToErrorFile() throws IOException {
```
tinaspark: I think every time we write to the error file we write the full payload so you can just have a single  version  of writeCopyDataToErrorFile() without having a payload argument

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
tinaspark: looks good. thanks!

tinaspark (changes requested):

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
Vizerai: Done.

tinaspark (approved):

PR #42, opened 2022-02-17
Update Java version to 11 for all tests

Vizerai:


olavloite (commented):
This (implicitly) drops support for Java 8. Why do we want to do that?

Vizerai:
Java 8 did not support all of the features that Java 11 does.  I ran into a few issues in adding some changes.

It is not necessary, but it might be useful to upgrade unless we require supporting Java 8. We can keep Java 8 support for now.

olavloite:
> Java 8 did not support all of the features that Java 11 does. I ran into a few issues in adding some changes.
> 
> It is not necessary, but it might be useful to upgrade unless we require supporting Java 8. We can keep Java 8 support for now.

Yes, older Java versions can be a little troublesome at times. The reason that I think that it's better to keep Java 8 support is the following:
1. We allow PgAdapter to be included as a dependency, so Java applications can just execute it internally instead of having to spin up a separate process. Raising the minimum version to Java 11 in our build would also impose that minimum version for anyone who wants to use it as a dependency.
2. Our client libraries for Cloud Spanner [support Java 8](https://github.com/googleapis/java-spanner#supported-java-versions). Having a higher minimum version for PgAdapter than for the Java client library could cause confusion or even problems for customers currently using Java 8, and expecting to be able to use PgAdapter with Java 8 as well.

PR #43, opened 2022-02-18
feat: Add Copy command parsing in QueryMessage and basic psql e2e test

Vizerai:


src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/ControlMessage.java
```diff
@@ -49,35 +50,46 @@ public ControlMessage(ConnectionHandler connection) throws IOException {
...
-        return new FlushMessage(connection);
-      default:
-        throw new IllegalStateException("Unknown message");
+    if (connection.getStatus() == ConnectionStatus.COPY_IN) {
+      switch (nextMsg) {
+        case CopyDoneMessage.IDENTIFIER:
+          return new CopyDoneMessage(connection);
+        case CopyDataMessage.IDENTIFIER:
+          return new CopyDataMessage(connection);
+        default:
```
tinaspark, line 61: should we handle CopyFail here as well?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/ControlMessage.java
```diff
@@ -49,35 +50,46 @@ public ControlMessage(ConnectionHandler connection) throws IOException {
...
+          return new DescribeMessage(connection);
+        case ExecuteMessage.IDENTIFIER:
+          return new ExecuteMessage(connection);
+        case CloseMessage.IDENTIFIER:
+          return new CloseMessage(connection);
+        case SyncMessage.IDENTIFIER:
+          return new SyncMessage(connection);
+        case TerminateMessage.IDENTIFIER:
+          return new TerminateMessage(connection);
+        case CopyDoneMessage.IDENTIFIER:
```
tinaspark: are we allowed to receive CopyDone/CopyData/CopyFail if we're not in COPY_IN mode?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDoneMessage.java
```diff
@@ -51,13 +51,15 @@ protected void sendPayload() throws Exception {
         // Spanner returned an error when trying to commit the batch of mutations.
         mw.writeMutationsToErrorFile();
         mw.closeErrorFile();
+        this.connection.setStatus(ConnectionStatus.IDLE);
```
olavloite, line 56: Sorry, I didn't notice this before, but don't we need to return something like a `ReadyResponse` after this? Or will the exception that is thrown ensure that the `ConnectionHandler` returns a response?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDoneMessage.java
```diff
@@ -51,13 +51,15 @@ protected void sendPayload() throws Exception {
...
         this.connection.removeActiveStatement(this.statement);
         throw e;
       }
     } else {
       mw.closeErrorFile();
     }
     new ReadyResponse(this.outputStream, ReadyResponse.Status.IDLE).send();
+    this.connection.setStatus(ConnectionStatus.IDLE);
```
olavloite, line 64: I think that this call to `setStatus(..)` and also the above one are causing a build error, because `ConnectionStatus` is not imported in this file. Would you mind taking a look at that?

olavloite (commented):
The build fails because of a compile error, I think because `ConnectionStatus` is not imported by `CopyDoneMessage`.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDoneMessage.java
Vizerai: The ConnectionHandler will return a ready response of IDLE on handling an error.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CopyDoneMessage.java
Vizerai: Fixed now.  Originally this PR was supposed to require changes from https://github.com/GoogleCloudPlatform/pgadapter/pull/40

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/ControlMessage.java
Vizerai: No, it should not.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/ControlMessage.java
Vizerai: Done.

Vizerai:
The e2e failures seem to be due to a change on the backend.  There a new pg_catalog entries returned from the information schema.

olavloite:
The current e2e test failure should be fixed once https://github.com/GoogleCloudPlatform/pgadapter/pull/40 has been merged (it's caused by the "?" that is fixed in #40)

tinaspark (commented):
LGTM once the test failures are fixed by #40

olavloite (approved):

olavloite:
@Vizerai I was wondering why the e2e-tests were failing, so I had a look at this PR and ended up doing a couple of small changes. It might be good if you go through those to check that you are OK with them before merging.

Vizerai:
> @Vizerai I was wondering why the e2e-tests were failing, so I had a look at this PR and ended up doing a couple of small changes. It might be good if you go through those to check that you are OK with them before merging.

The changes look good.  Thanks for the update.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -169,11 +169,11 @@ private void queryInformationSchema() throws SQLException {
     Map<String, TypeCode> tableColumns = new LinkedHashMap<>();
     PreparedStatement statement =
         this.connection.prepareStatement(
-            "SELECT "
+            "/*GSQL*/SELECT "
```
tinaspark, line 175: why do we need to issue this as a GoogleSQL query?

Vizerai: The backend no longer returns these as GoogleSQL types for this query. It instead returns the PG types now. We need the GoogleSQL types to fill out the mutations.

This was just recently changed:
https://buganizer.corp.google.com/issues/198312860

tinaspark: I see. Can you add a comment?

// The mutation API requires GoogleSQL type names instead of PostgreSQL type names for the table column types
// Issue the information_schema table query as a GoogleSQL query so that it will return GoogleSQL type names.

tinaspark (approved):

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
Vizerai: Done.

PR #45, opened 2022-02-21
build: randomize database name for psql tests

olavloite:
Randomizes the database names used for psql tests to reduce the probability of two builds running in parallel interfering with each other. Each psql test used a fixed database name, which meant that psql tests for different versions of psql could run in parallel, but not multiple psql tests of the same psql version. This would occasionally cause random build errors if two or more PRs were opened shortly after each other.

Also removes the assumption that `pg_catalog` is present in the test database.

Fixes b/217744569


Vizerai (approved):
LGTM

Vizerai:
I noticed several of the e2e tests have been failing on other PRs.  Was there a change on the backend that broke them?

olavloite:
> I noticed several of the e2e tests have been failing on other PRs. Was there a change on the backend that broke them?

Yes. `pg_catalog` has been hidden for databases on the production endpoint.

PR #46, opened 2022-02-24
feat:  User agent header addition

gauravsnj:
Currently, client libraries send headers which is used for tracking adoption, version etc
We should extend that header (user-agent) to track PGAdapter binary and version.

fixes http://b/213946108


olavloite:
@gauravsnj To format the file according to the wishes of the lint check you must run the command `mvn com.coveo:fmt-maven-plugin:format` from the command line in the root of the project.

src/test/java/com/google/cloud/spanner/pgadapter/AbstractMockServerTest.java
```diff
@@ -213,6 +224,28 @@ public static void startMockSpannerAndPgAdapterServers() throws Exception {
...
+                      Metadata metadata,
+                      ServerCallHandler<ReqT, RespT> serverCallHandler) {
+
+                    if (SpannerGrpc.getExecuteStreamingSqlMethod()
+                        .getFullMethodName()
+                        .equals(serverCall.getMethodDescriptor().getFullMethodName())) {
+                      String userAgent =
+                          metadata.get(
+                              Metadata.Key.of(
+                                  "x-goog-api-client", Metadata.ASCII_STRING_MARSHALLER));
```
olavloite, line 241: The client library token header that includes the value is named `x-goog-api-client`.

src/test/java/com/google/cloud/spanner/pgadapter/AbstractMockServerTest.java
```diff
@@ -213,6 +224,28 @@ public static void startMockSpannerAndPgAdapterServers() throws Exception {
...
+                      assertNotNull(userAgent);
+                      assertTrue(userAgent.contains("pg-adapter"));
```
olavloite, line 243: The header value also includes information about the gax version used etc. So we should only verify that the `pg-adapter` value is included, not that it only contains `pg-adapter`.

olavloite (approved):

PR #47, opened 2022-02-27
test: add Golang tests

olavloite:
Adds test using the native Golang pgx driver.


src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -107,20 +110,27 @@ public IntermediatePortalStatement bind(
...
-        "new DescribeStatementMetadata(
-            Converter.getParams(
-                this.parameterDataTypes,
-                this.totalParameters)
-            );"
-        for non-SELECTs.
-    */
-    throw new IllegalStateException(
-        "Describe statements for non-bound statements are not allowed in spanner.");
+  public DescribeMetadata describe() {
```
olavloite, line 116: @pratickchokhani The Golang driver automatically uses prepared statements for all parameterized statements. This is where that ends up in PGAdapter. The 'implementation' in this PR is the simplest possible, and only intended to verify that parameterized statements can be used with the Golang driver, but it can also be a useful starting point for your current investigation.

What basically happens here is that:
1. The user application executes something like `conn.QueryRow(ctx, "SELECT * FROM FOO WHERE BAR=$1", "baz")` in the Golang application (see https://github.com/GoogleCloudPlatform/pgadapter/blob/4a64448855a626ceed7c9585cd6cc5b56feca1a3/src/test/golang/pgx.go#L87).
2. The Golang driver sends a `DescribeStatement` message to PGAdapter with only the SQL string `SELECT * FROM FOO WHERE BAR=$1` and no parameter values. It expects PGAdapter to respond with a `RowDescription` message that contains the metadata of the `ResultSet` that would be returned, and a list of the data types of the parameters in the SQL string.
3. PGAdapter currently responds by executing an `ExecuteSqlRequest` with `QueryMode.PLAN` on Cloud Spanner, and then returning the metadata that is returned by that statement as the `RowDescription`. In addition it just scans through the SQL string to count the number of parameters, and returns an array containing a zero for each parameter it found in the SQL string. Zero in this case means 'the data type of the parameter is unspecified'.

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/ParameterDescriptionResponse.java
```diff
@@ -26,16 +26,16 @@
   private static final int PARAMETER_NUMBER_LENGTH = 2;
   private static final int PARAMETER_LENGTH = 4;
 
-  private final List<Integer> parameters;
+  private final int[] parameters;
 
-  public ParameterDescriptionResponse(DataOutputStream output, List<Integer> parameters) {
+  public ParameterDescriptionResponse(DataOutputStream output, int[] parameters) {
```
pratickchokhani, line 31: Why is this change required? Doesn't List provide more extensibility also safer?

olavloite: Not necessarily, both `List<Integer>` and `int[]` have their pros and cons. `List<Integer>` is easier to work with if the collection is changing. It is however less efficient, as it requires boxing and unboxing of the primitive values, and it can more easily lead to errors caused by `NullPointerException`s. The default value of an item in a `List<Integer>` is `null`. The default value in an `int[]` is `0`. The latter is also what the PG wire protocol uses as the default value for parameter types with an unknown value. The latter is the reason that I changed it, as it makes the code less bloated and better correspond with what the wire protocol uses.

pratickchokhani (approved):

PR #48, opened 2022-02-28
build: use the publicly available JDBC driver

olavloite:
Removes the dependency on the private snapshot version of the Cloud
Spanner JDBC driver and instead uses the publicly available JDBC driver
that supports the PostgreSQL dialect.

Also removes the usage of `/*GSQL*/` headers, as those are no longer
supported in the JDBC driver.


Vizerai (approved):
LGTM

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -153,16 +153,34 @@ private void verifyCopyColumns() throws SQLException {
...
-    // Eliminate size modifiers in column type (e.g. STRING(100), BYTES(50), etc.)
+  private static TypeCode parsePostgreSQLDataType(String columnType) {
+    // Eliminate size modifiers in column type (e.g. character varying(100), etc.)
     int index = columnType.indexOf("(");
     columnType = (index > 0) ? columnType.substring(0, index) : columnType;
-    TypeCode type = TypeCode.valueOf(columnType);
-    if (type == null) {
-      throw new IllegalArgumentException(
-          "Unrecognized or unsupported column data type: " + columnType);
+    switch (columnType) {
```
Vizerai, line 160: This is not a comprehensive list of the types in TypeCode.

Are Struct and Array not supported?  I thought support for Arrays was added.

olavloite: Array support is being added as we speak, but is not yet available. We will however need to revisit this when it comes available, as the `data_type` column contains the string `ARRAY` for all array types (at least it does in PostgreSQL). That means that we will need to get additional data from another column or table to get the element type. The most logical table to use is `pg_catalog.pg_attribute`, but that table is currently not available.

Struct is currently not supported.

olavloite:
> LGTM

Thanks for the quick response

Vizerai:
I think this broke the artifact registry upload to span-cloud-testing.  Was this intentional?  If so, are you moving the jars to a new location?

PR #49, opened 2022-03-01
fix: Update Copy parser to handle ',' separated option list

Vizerai:
Update Copy parser to handle ',' separated option list. Also expand single char regex to wider set of characters [! - ~].


src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -242,6 +241,8 @@ private void parseCopyStatement() throws Exception {
       parse(sql, this.options);
     } catch (Exception e) {
       throw new SQLException("Invalid COPY statement syntax: " + e.toString());
+    } catch (TokenMgrError e) {
```
olavloite: nit: you can combine the two catch clauses into one like this:

```
    } catch (Exception | TokenMgrError e) {
      throw new SQLException("Invalid COPY statement syntax: " + e.toString());
    }
```

src/test/java/com/google/cloud/spanner/pgadapter/ParserTest.java
```diff
@@ -685,6 +685,43 @@ public void testCopyParserOptions() throws Exception {
...
+                + value2
+                + "');";
+        parse(sql, options);
+        assertThat(options.getTableName(), is(equalTo("users")));
+        assertThat(options.getFromTo(), is(equalTo(FromTo.FROM)));
+        assertThat(options.getFormat(), is(equalTo(Format.TEXT)));
+        assertThat(options.getDelimiter(), is(equalTo(value)));
+        assertThat(options.getQuote(), is(equalTo(value1)));
+        assertThat(options.getEscape(), is(equalTo(value2)));
+      }
```
olavloite, line 797: Can we add the following assertion (here and also for the other tests):

```suggestion
        assertEquals(StatementParser.parseCommand(StatementParser.removeCommentsAndTrim(sql)), "COPY");
      }
```

This assertion currently fails for `ESCAPE '\'`, as our parser (correctly?) sees '\' as an unclosed string literal.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java

src/test/java/com/google/cloud/spanner/pgadapter/ParserTest.java
Vizerai: Yes, removeCommentsAndTrim() does not allow a single '\\' in the statement.  Copy expects the delimiter, escape, and quote options to be single characters.

Le'ts drop ', ", and \ for now.  The logic in removeCommentsAndTrim() would break if we allowed those. We could add them back in at a later point after we rework removeCommentsAndTrim().

olavloite: yeah, that sounds reasonable.

olavloite (approved):

PR #52, opened 2022-03-04
feat: COPY supports large files and streaming input from PostgreSQL

olavloite:
Adds support for large files for COPY to import large amounts of data into Cloud Spanner using the PostgreSQL COPY statement. Importing large data sets is only possible after first executing `set autocommit_dml_mode='partitioned_non_atomic'`. The following COPY operation will then automatically be split into multiple transactions that are executed partly in parallel for efficient loading into Cloud Spanner.

The feature can be used both to import previously exported files as well as streaming data directly from a real PostgreSQL database by piping the output of a COPY .. TO STDOUT into a COPY .. FROM STDIN command.

The feature has been tested with large datasets to ensure that it does not run out of memory or in other ways breaks if it is fed a very large stream of records. The PR refactors the `MutationWriter` class extensively to achieve this, as it is necessary to connect the different streams of data directly to each other, instead of first loading everything into memory before processing it. This is achieved by connecting the data streams in the following ways:
1. The `MutationWriter.addCopyData` writes the received bytes into a `PipedOutputStream`. This is done by the normal worker thread of the connection (the `ConnectionHandler`).
2. The `PipedOutputStream` is connected to a `PipedInputStream`, which automatically receives all data from the `PipedOutputStream`. The pipe has a (default) buffer size of 65Kb.
3. The `CSVParser` is created with the `PipedInputStream` as its input source. This ensures that the `CSVParser` can read the records directly from the incoming data stream, instead of having to wait for all data to be received first.
4. Reading from the `CSVParser` is done using a separate worker thread that is created specifically for the entire `COPY` operation (this is the new `ExecutorService` in `CopyStatement`). This ensures that the main `ConnectionHandler` worker thread will write the data to the stream, while the `COPY` worker thread will read the data.
5. The `COPY` worker thread reads records from the CSVParser and creates a mutation for each CSVRecord. Note: the CSVParser iterator will block until there is enough data to construct a complete record, so we don't have to worry about getting an incomplete record from this iterator.
6. If we are using `PARTITIONED_NON_ATOMIC`:
6.1. When we reach one of the transaction limits (commit size or number of mutations), a `Commit` with all the mutations in that batch will be prepared.
6.2. The `Commit` operation is executed using a separate thread pool (the `ExecutorService` in `MutationWriter`). This enables multiple commits to be processed in parallel on the backend, as we know (or assume) that the records a non-overlapping.
6.3 Each `Commit` operation is also added to a `BlockingDeque` that allows at most 128 operations in parallel. This `BlockingDeque` ensures that we apply back pressure in the entire stream if the backend cannot keep up with the incoming data stream. A `Commit` operation can only be started if there is a free spot in this `BlockingDeque`, and if there is no space, the `COPY` worker thread will be blocked until there is space. This cascades backwards in the pipeline, as that also means that it pauses reading records from the CSVParser, which again means that writing to the `PipedOutputStream` that is used by `addCopyData` will block (when the 65Kb buffer is full). This all means that the client must wait before it can send any more data to the server.

The above behavior has been verified by streaming an (almost) endless stream of generated data from a real PostgreSQL server using the following command:

```bash
psql -h localhost -d my-test-db \
-c "copy (select (random()*1000000000)::bigint, random()<0.5, md5(random()::text || clock_timestamp()::text)::bytea, \
random()*123456789, (random()*999999)::int, (random()*999999)::numeric, now()-random()*interval '50 year', \
md5(random()::text || clock_timestamp()::text)::varchar \
from generate_series(1, 100000000) s(i)) to stdout" \
| psql -h localhost -p 5433 -d test \
-c "set autocommit_dml_mode='partitioned_non_atomic'" \
-c "copy all_types from stdin;"
```

The above example uses this data model:

```sql
create table all_types (
  col_bigint bigint not null primary key,
  col_bool bool,
  col_bytea bytea,
  col_float8 float8,
  col_int int,
  col_numeric numeric,
  col_timestamptz timestamptz,
  col_varchar varchar(100))
```

### Performance
Copying 1,000,000 rows using the above table definition (approx 180MB of text data) takes less than 1min15sec when executed on from local laptop against a Cloud Spanner instance:

Three test runs:

```
time psql -h localhost -d my-test-db -c "copy (select (random()*1000000000)::bigint, random()<0.5, md5(random()::text || clock_timestamp()::text)::bytea, random()*123456789, (random()*999999)::int, (random()*999999)::numeric, now()-random()*interval '50 year', md5(random()::text || clock_timestamp()::text)::varchar from generate_series(1, 1000000) s(i)) to stdout" | psql -h localhost -p 5433 -d test -c "set autocommit_dml_mode='partitioned_non_atomic'" -c "copy all_types from stdin;"
SET
COPY 1000000

real    1m13.944s
user    0m0.590s
sys     0m0.428s

-------------------------------------------

time psql -h localhost -d my-test-db -c "copy (select (random()*1000000000)::bigint, random()<0.5, md5(random()::text || clock_timestamp()::text)::bytea, random()*123456789, (random()*999999)::int, (random()*999999)::numeric, now()-random()*interval '50 year', md5(random()::text || clock_timestamp()::text)::varchar from generate_series(1, 1000000) s(i)) to stdout" | psql -h localhost -p 5433 -d test -c "set autocommit_dml_mode='partitioned_non_atomic'" -c "copy all_types from stdin;"
SET
COPY 1000000

real    1m14.647s
user    0m0.545s
sys     0m0.384s

-------------------------------------------

time psql -h localhost -d my-test-db -c "copy (select (random()*1000000000)::bigint, random()<0.5, md5(random()::text || clock_timestamp()::text)::bytea, random()*123456789, (random()*999999)::int, (random()*999999)::numeric, now()-random()*interval '50 year', md5(random()::text || clock_timestamp()::text)::varchar from generate_series(1, 1000000) s(i)) to stdout" | psql -h localhost -p 5433 -d test -c "set autocommit_dml_mode='partitioned_non_atomic'" -c "copy all_types from stdin;"
SET
COPY 1000000

real    1m7.251s
user    0m0.589s
sys     0m0.469s

```


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -161,9 +163,16 @@ private void handleError(DataOutputStream output, Exception e) throws Exception
         Level.WARNING,
         "Exception on connection handler with ID {0}: {1}",
         new Object[] {getName(), e});
-    this.status = ConnectionStatus.IDLE;
-    new ErrorResponse(output, e, ErrorResponse.State.InternalError).send();
-    new ReadyResponse(output, ReadyResponse.Status.IDLE).send();
+    if (this.status == ConnectionStatus.TERMINATED) {
```
olavloite, line 189: If the connection has been set to `Terminated`, the error handling should not reset it to `Idle`. Instead, the server should send the error message and a `TerminateResponse` and then close the connection.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -161,9 +163,16 @@ private void handleError(DataOutputStream output, Exception e) throws Exception
...
+      new ErrorResponse(output, e, ErrorResponse.State.InternalError, Severity.FATAL).send();
+      new TerminateResponse(output).send();
+    } else if (this.status == ConnectionStatus.COPY_IN) {
+      new ErrorResponse(output, e, ErrorResponse.State.InternalError).send();
```
olavloite, line 193: If we are in `CopyIn` mode, the connection should not be reset to `Idle` and it should not respond with `ReadyResponse`.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -43,7 +52,10 @@
 
   // Table columns read from information schema.
   private Map<String, TypeCode> tableColumns;
+  private int indexedColumnsCount;
   private MutationWriter mutationWriter;
+  private Future<Long> updateCount;
+  private final ExecutorService executor = Executors.newSingleThreadExecutor();
```
olavloite, line 61: This executor is responsible for reading from the stream of incoming `CopyData` messages. The `CopyData` messages are put into a pipeline by the main `ConnectionHandler` thread, and a different thread must read these messages to prevent potential deadlocks.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -53,6 +65,39 @@ public CopyStatement(String sql, Connection connection) throws SQLException {
     this.statement = connection.createStatement();
   }
 
+  @Override
+  public Exception getException() {
+    // Do not clear exceptions on a CopyStatement.
```
olavloite, line 72: The default implementation of `getException` automatically clears the exception after getting it. We need the error state to be persistent when we are in `CopyIn` mode, as there is no real error handling during a copy operation. This means that we in most cases just need to terminate the connection when an error occurs, and keeping the error message around after getting it the first time means that we can include it in the last message before we kill the connection.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -212,10 +258,34 @@ private void queryInformationSchema() throws SQLException {
     if (options.getColumnNames() != null) {
       verifyCopyColumns();
     }
+    this.indexedColumnsCount = queryIndexedColumnsCount(tableColumns.keySet());
+  }
+
+  private int queryIndexedColumnsCount(Set<String> columnNames) throws SQLException {
```
olavloite: We also need the number of columns that are included in secondary indexes to calculate the number of mutations that one record consists of.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -228,14 +298,35 @@ public void execute() {
...
           new MutationWriter(
-              options.getTableName(), getTableColumns(), getParserFormat(), hasHeader());
+              getTransactionMode(),
+              connection.unwrap(CloudSpannerJdbcConnection.class),
+              options.getTableName(),
+              getTableColumns(),
+              indexedColumnsCount,
+              getParserFormat(),
+              hasHeader());
+      updateCount = executor.submit(mutationWriter);
```
olavloite, line 314: Start the `MutationWriter` using the background executor. This will read from the CSVParser streaming iterator that is fed from the incoming `CopyData` messages.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -14,225 +14,580 @@
...
+  private final int maxBatchSize;
+  private final long commitSizeLimitForBatching;
+  private final int maxParallelism;
+  private final int pipeBufferSize;
+  private final CSVFormat format;
+  private final CSVParser parser;
+  private PrintWriter errorFileWriter;
+  private final PipedOutputStream payload = new PipedOutputStream();
+  private final AtomicBoolean rollback = new AtomicBoolean(false);
+  private final ListeningExecutorService executorService =
```
olavloite, line 136: This executor is responsible for the execution of the separate transactions when using `partitioned_non_atomic`. This significantly speeds up large imports compared to executing all of them sequentially.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -14,225 +14,580 @@
...
+  public void addCopyData(byte[] payload) throws SQLException {
+    synchronized (lock) {
+      if (this.exception != null) {
+        throw this.exception;
+      }
+    }
+    try {
+      this.payload.write(payload);
+    } catch (IOException e) {
+      throw new SQLException(e);
```
olavloite: This just keeps the number of different types of exceptions that we need to handle further down the chain smaller. (And it will be replaced with `SpannerException` once we have switched to the Connection API)

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
```diff
@@ -75,95 +58,6 @@ public void testQuery() throws SQLException {
     }
   }
 
-  @Test
```
olavloite, line 301: All copy tests using the mock server have been moved to a separate file.

aseering:
@tinaspark FYI

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -14,225 +14,622 @@
...
     this.isHeaderParsed = false;
     this.tableName = tableName;
     this.tableColumns = tableColumns;
+    int mutationLimit =
+        Math.max(
+            Integer.parseInt(
+                System.getProperty(
+                    "copy_in_mutation_limit", String.valueOf(DEFAULT_MUTATION_LIMIT))),
+            1);
+    this.maxBatchSize = Math.max(mutationLimit / (tableColumns.size() + indexedColumnsCount), 1);
```
pratickchokhani, line 188: Shouldn't this be 20K since spanner's max mutation limit is 20K?

olavloite: The 20,000 mutations limit in Spanner is unfortunately a little confusing. It does not mean a maximum of 20,000 mutation objects or rows, it means a maximum of 20,000 column mutations. And it gets a little bit more complex as every secondary index that includes a column that is inserted/updated is also counted as a mutation. So assume that we have the following table and index definition:

```sql
create table numbers (number bigint not null primary key, name varchar(1000));
create index idx_numbers on numbers (name);
```

Any insert into this table will count as 3 mutations:
1. 2 mutations because it the table has two columns.
2. +1 mutation because the `name` column is included in a secondary index.

See also https://cloud.google.com/spanner/quotas#note2

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
```diff
@@ -1387,72 +1379,60 @@ public void testCopyBatchSizeLimit() throws Exception {
...
-    Mockito.when(spannerType.getString("data_type")).thenReturn("bigint", "character varying");
-    Mockito.when(spannerType.next()).thenReturn(true, true, false);
-    Mockito.when(connection.executeQuery(any(Statement.class))).thenReturn(spannerType);
-
     CopyStatement copyStatement = new CopyStatement("COPY keyvalue FROM STDIN;", connection);
 
     Assert.assertFalse(copyStatement.isExecuted());
     copyStatement.execute();
-    Assert.assertTrue(copyStatement.isExecuted());
+    assertTrue(copyStatement.isExecuted());
```
Vizerai, line 1390: nit: This doesn't match the assertions below.  The class name is omitted here but not below and other references.

Assert.assertFalse
Assert.assertTrue

Vizerai (approved):
LGTM.  Thanks for adding this functionality!

src/test/java/com/google/cloud/spanner/pgadapter/ProtocolTest.java
olavloite: Yeah, I know this looks a little messy at this moment. I tried to minimise the number of merge conflicts, so I didn't change the existing assertions where it was not necessary. My plan was to do a general overhaul of the assertions in the different test classes in a separate PR so they are all in the same style.

olavloite:
> LGTM. Thanks for adding this functionality!

Thanks for looking into this. I know it was a large PR, and so not a very easy one to review.

PR #57, opened 2022-03-09
fix: Statements with no results would return an error

### LINKED ISSUE #56: DDL statements return an error as PGAdapter tries to get the update count of the result
DDL statements return no results, but PGAdapter tries to get the update count of these statements, which causes an error to be returned. The DDL statement is executed and persisted in Spangres, it is only the return value that is wrong.

olavloite:
Statements that return no results, such as DDL statements, would cause
an error as PGAdapter tried to get the update count of these.

Fixes #56


Vizerai (approved):
LGTM

pratickchokhani (approved):

gauravsnj (approved):
LGTM

PR #59, opened 2022-03-09
chore: model ProxyServer as an ApiService

olavloite:
Use a standard AbstractApiService to model the ProxyServer
instead of keeping track of the state of the server manually.
This reduces code complexity by relying on a standard service
that is included in the standard Google libraries.


Vizerai (approved):
LGTM

PR #61, opened 2022-03-10
build: restore the distribution management tags in pom.xml

olavloite:
The specific distribution management setup for PGAdapter was accidentally removed when switching to the public Java client. This causes the publishing of the jars in the artifact registry to fail.


Vizerai (approved):
LGTM

gauravsnj (approved):
LGTM

PR #67, opened 2022-03-19
fix: handle connection failures better by returning an error response

olavloite:
Improves error handling of connect request failures by returning an error instead of terminating the entire server.


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -169,12 +169,23 @@ public void run() {
   }
 
   /** Called when a Terminate message is received. This closes this {@link ConnectionHandler}. */
-  public void handleTerminate() throws Exception {
+  public void handleTerminate() {
```
gauravsnj, line 172: Was this "throws Exception" part unnecessary or did it become obsolete due to the new changes?

gauravsnj (approved):
LGTM

PR #69, opened 2022-03-19
chore: disable logging during tests by default

olavloite:
Reduce the noise during tests by turning off logging by default during unit tests.


gauravsnj (approved):
LGTM

PR #76, opened 2022-03-23
build: only run Go tests in profile test-all

olavloite:
Disables running Go unit tests by default, so environments without a Go development
environment do not fail tests. Enabling the profile test-all will execute the Go unit
tests along with the Java tests. This profile is activated on the CI build.


gauravsnj (approved):
LGTM

PR #77, opened 2022-03-23
perf: reduce flushing the output stream

olavloite:
Reduce the number of times that the output buffer is flushed to improve network
performance for (large) queries.


src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
```diff
@@ -115,13 +115,14 @@ public void handleQuery() throws Exception {
...
-            .send();
+            .send(false);
       }
       this.sendSpannerResult(this.statement, QueryMode.SIMPLE, 0L);
       boolean inTransaction = connection.getSpannerConnection().isInTransaction();
       new ReadyResponse(
               this.outputStream, inTransaction ? Status.TRANSACTION : ReadyResponse.Status.IDLE)
-          .send();
+          .send(false);
+      this.outputStream.flush();
```
gauravsnj: Is this change necessary? I mean if we would've left line 124 as it was and didn't add line 125, I guess then also it would have been the same.

gauravsnj (approved):
LGTM

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
olavloite: Good point, changed.

PR #78, opened 2022-03-23
perf: reduce logger overhead

olavloite:
Use lambda expressions for log statements instead of constructing the log
message every time, even when the statement is not logged.


src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePortalStatement.java
```diff
@@ -94,7 +94,7 @@ public DescribeMetadata describe() {
         proxy. As such, we send the user a descriptive message to help them understand the issue
         in case they misuse the method.
       */
-      logger.log(Level.SEVERE, e.toString());
+      logger.log(Level.SEVERE, e, e::toString);
```
gauravsnj, line 97: What's this change intending to do? AFAIK the string message should always be second parameter in the logger.log() method.

olavloite: A couple of new methods were added in Java 8 for java.util.Logger. One of them is [public void log](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html#log-java.util.logging.Level-java.lang.Throwable-java.util.function.Supplier-)([Level](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Level.html) level,
                [Throwable](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html) thrown,
                [Supplier](https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html)<[String](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html)> msgSupplier)

The advantage of using this method is that it will automatically include the error information in the log, but more importantly in this case: It takes a `Supplier<String>` as the message argument. This means that `e.toString()` will only be called if there is actually something to log, which can improve performance if `e.toString()` is constructs a large string, and the logger will not always actually log the statement, for example if logging has been turned off.

gauravsnj: Okay Thanks for the info

gauravsnj (approved):
LGTM

PR #81, opened 2022-03-25
refactor: move pgx test to separate package

olavloite:
Moves the `pgx` tests to a separate Go package, so we can easily add tests for other Go drivers and tools if needed.


gauravsnj (approved):
LGTM

PR #82, opened 2022-03-25
feat: support pgx in extended mode

olavloite:
Adds support and tests for PGX in extended query mode. This also fixes
some minor issues with sending bytea values in text format. bytea values
in text format should use the newer hex format, as some newer drivers
will not understand the older octal format. Also, any '\' characters in
a literal in an array must be escaped.

The Golang PGX driver will by default use extended mode in combination
with DescribeStatement. The problem with DescribeStatement is that we
currently do not have any way of really supporting it. We therefore need
a patch in PGX to be able to work with it. However, that patch is also
valuable for other users of PGX, and hopefully we will be able to get it
merged in the main line of PGX.


src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
```diff
@@ -118,6 +118,9 @@ private boolean stringEquivalence(Code arrayElementType) {
    */
   private String stringify(String value) {
     if (this.isStringEquivalent) {
+      if (value.indexOf('\\') > -1) {
+        value = value.replace("\\", "\\\\");
```
olavloite, line 126: As these values are stringifyed, they also need to escape any backslashes inside the strings.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/BinaryParser.java
```diff
@@ -63,7 +64,7 @@ public int getSqlType() {
 
   @Override
   protected String stringParse() {
-    return this.item == null ? null : PGbytea.toPGString(this.item.toByteArray());
+    return this.item == null ? null : "\\x" + Utils.toHexString(this.item.toByteArray());
```
olavloite, line 67: PGAdapter should use the newer hex format when sending bytea values as text: https://www.postgresql.org/docs/current/datatype-binary.html#id-1.5.7.12.9

pgx only supports this hex format, and will refuse the older escape format

src/main/java/com/google/cloud/spanner/pgadapter/parsers/Parser.java
```diff
@@ -209,7 +209,7 @@ public static int toOid(Type type) {
       case BYTES:
         return Oid.BYTEA;
       case TIMESTAMP:
-        return Oid.TIMESTAMP;
+        return Oid.TIMESTAMPTZ;
```
olavloite, line 216: The timestamps that we return do contain a timezone.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePortalStatement.java
```diff
@@ -62,7 +62,7 @@ public short getParameterFormatCode(int index) {
   @Override
   public short getResultFormatCode(int index) {
     if (this.resultFormatCodes == null || this.resultFormatCodes.isEmpty()) {
-      return 0;
+      return super.getResultFormatCode(index);
```
olavloite, line 65: Bubble up to the super implementation. This does nothing at the moment, but we should implement a smarter default than just return 0 (text) when the client did not specify what it wants. That gives us the freedom to choose the binary format when we think that is better.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -114,13 +114,16 @@ public IntermediatePortalStatement bind(
 
   @Override
   public DescribeMetadata describe() {
-    Statement statement = Statement.of(this.sql);
-    try (ResultSet resultSet = connection.analyzeQuery(statement, QueryAnalyzeMode.PLAN)) {
-      // TODO: Remove ResultSet.next() call once this is supported in the client library.
-      // See https://github.com/googleapis/java-spanner/pull/1691
-      resultSet.next();
-      return new DescribeStatementMetadata(getParameterTypes(), resultSet);
+    if (PARSER.isQuery(this.sql)) {
```
olavloite: We should not try to describe a DML using `analyzeQuery`. Instead, we should just return the parameter types.

src/test/golang/go.mod
```diff
@@ -11,7 +14,8 @@ require (
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgproto3/v2 v2.2.0 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20200714003250-2b9c44734f2b // indirect
-	github.com/jackc/pgtype v1.10.0 // indirect
 	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97 // indirect
 	golang.org/x/text v0.3.6 // indirect
 )
+
+replace github.com/jackc/pgx/v4 => github.com/olavloite/pgx/v4 v4.15.1
```
olavloite: This is the patch for the pgx driver that enables skipping the `DescribeStatement` message, as it is not needed. The JDBC driver already does this.

src/test/java/com/google/cloud/spanner/pgadapter/ParserTest.java
```diff
@@ -305,7 +305,7 @@ public void testTimestampParsingTimestampValidityChecks() {
   public void testBinaryParsing() {
     ByteArray value = ByteArray.copyFrom(new byte[] {(byte) 0b01010101, (byte) 0b10101010});
     byte[] byteResult = {(byte) 0b01010101, (byte) 0b10101010};
-    byte[] stringResult = {'U', '\\', '2', '5', '2'};
+    byte[] stringResult = {'\\', 'x', '5', '5', 'a', 'a'};
```
olavloite, line 308: Changed to hex format from the old escape format for bytea values in text format.

src/test/java/com/google/cloud/spanner/pgadapter/PgxMockServerTest.java
```diff
@@ -98,7 +106,9 @@ public static void compile() throws IOException, InterruptedException {
 
   private GoString createConnString() {
     return new GoString(
-        String.format("postgres://uid:pwd@localhost:%d/test-db", pgServer.getLocalPort()));
+        String.format(
+            "postgres://uid:pwd@localhost:%d/test-db?statement_cache_capacity=0",
```
olavloite: Do not use a local statement cache in the pgx driver. This is needed to enable skipping the `DescribeStatement` message.

src/test/java/com/google/cloud/spanner/pgadapter/PgxMockServerTest.java
```diff
@@ -130,15 +140,16 @@ public void testHelloWorld() {
...
-    // 1. DESCRIBE statement
-    // 2. DESCRIBE portal
-    // 3. EXECUTE portal
-    assertEquals(3, requests.size());
+    // The patched version of pgx sends the query two times.
+    // This will be one time when https://github.com/GoogleCloudPlatform/pgadapter/pull/80 has been
+    // merged.
+    // 1. DESCRIBE portal
+    // 2. EXECUTE portal
+    assertEquals(2, requests.size());
```
olavloite: Yay! One less round trip :-)

And when #80 is merged, it will be down to just one round trip.

libingye816:
Hey @olavloite, I observe two failures after merging this pr. Can you check those out?

Failures:
[ERROR]   ITPgxTest.testInsertNullsAllDataTypes:165 expected null, but was:<failed to execute insert statement: ERROR: INVALID_ARGUMENT: com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: [ERROR] there is no parameter $3 (SQLSTATE XX000)>
[ERROR]   ITPgxTest.testQueryAllDataTypes:145 expected null, but was:<Failed to execute query: number of field descriptions must equal number of destinations, got 8 and 9>

olavloite:
Where are you seeing this? I just merged the latest changes into [this PR](https://github.com/GoogleCloudPlatform/pgadapter/pull/112), and all tests were successful.

libingye816:
I cloned support-pgx-extended-mode branch and run ITPgxTest locally.

libingye816:
> Where are you seeing this? I just merged the latest changes into [this PR](https://github.com/GoogleCloudPlatform/pgadapter/pull/112), and all tests were successful.

I cloned support-pgx-extended-mode branch and run ITPgxTest locally.

PR #87, opened 2022-03-29
fix: handle null values in arrays

olavloite:
Null elements inside an array were not handled correctly. These were
either sent as the plain string null (all lower case) or as a quoted
string "null". Both were incorrect, and should in all cases be NULL (all
upper case letters, no quotes).


gauravsnj (approved):
LGTM

PR #92, opened 2022-03-30
build: simplify Docker build

olavloite:
Simplifies the Docker build and image by:
1. Using pre-built images that already contain Maven and a JRE.
2. Downloading and caching Maven dependencies for quicker rebuilds.
3. Building a thin jar + dependencies folder instead of a fat jar.

This also reduces the size of the Docker container from more than 800MB to 330MB.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -2,12 +2,7 @@
 #                                     BUILD                                    #
 ################################################################################
 
-FROM gcr.io/cloud-marketplace/google/ubuntu1804:latest AS build
-
-ENV CLOUD_SDK_VERSION=340.0.0
-
-RUN apt-get update                                                          && \
-    apt-get -qq -y install git maven python
+FROM maven:3.8.4-eclipse-temurin-17-alpine AS build
 
 # Copy over build files to docker image.
 COPY LICENSE ./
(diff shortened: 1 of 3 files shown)


build/Dockerfile
```diff
@@ -2,12 +2,7 @@
 #                                     BUILD                                    #
 ################################################################################
 
-FROM gcr.io/cloud-marketplace/google/ubuntu1804:latest AS build
-
-ENV CLOUD_SDK_VERSION=340.0.0
-
-RUN apt-get update                                                          && \
-    apt-get -qq -y install git maven python
+FROM maven:3.8.4-eclipse-temurin-17-alpine AS build
```
olavloite, line 5: - maven uses the Apache License v2
- eclipse-temurin uses the Apache License v2 and the GNU General Public License, version 2 (GPL2) for OpenJDK
- Alpine uses the MIT License

These are all already used by other components in this repository, and hence pre-approved.

build/Dockerfile
```diff
@@ -19,29 +14,19 @@ COPY src src/
...
 
 ################################################################################
 #                                   RELEASE                                    #
 ################################################################################
 
-FROM gcr.io/cloud-marketplace/google/ubuntu1804:latest
-
-RUN apt-get update                                                          && \
-    apt-get -qq -y install default-jre
+FROM eclipse-temurin:17-jre
```
olavloite, line 27: (See above, this also uses pre-approved licenses)

pom.xml
```diff
@@ -204,6 +204,62 @@
...
       </properties>
     </profile>
+
+    <profile>
+      <id>assembly</id>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-resources-plugin</artifactId>
+            <version>3.2.0</version>
```
snehashah16, line 199: do we need this divergence between fat-jar and docker builds?
eventually this list may grow and the divergence always makes it hard to maintain and track

build/Dockerfile
snehashah16, line 27: curious, do we need eclipse for JRE ?

olavloite: `eclipse` is just the name of the Docker image, it does not contain the `eclipse` IDE. It is maintained by the Eclipse Software Foundation: https://adoptium.net/docs/faq

pom.xml
olavloite: This is a safer way to build an executable Java distribution than building a fat jar, and I would like to try to convert the fat-jar into one that uses a similar approach as this.

The potential problem with a fat jar is that it places everything into one jar, which means that if for example two different dependencies define the same configuration files or other resources, only one of the versions will be included in the fat jar. This can cause silent breakage if a dependency adds a new  file in a new version and expects it to be added to the final build. Example:

* dependency-1.jar contains `META-INF/services/io.grpc.LoadBalancerProvider` with the load balancer provider that it wants. This file will automatically be copied to `META-INF/services/io.grpc.LoadBalancerProvider` in the fat jar. The classes in dependency-1.jar will still be able to find the file, as it is in the expected location.
* During an update of the dependencies of PGAdapter dependency-2.jar also adds a `META-INF/services/io.grpc.LoadBalancerProvider`. The fat jar will only contain one of these files (and it is random which it is).

The setup for the Docker image keeps all dependency jars as they are and place these in a `lib` directory that is then automatically added to the classpath when PGAdapter is started.

snehashah16 (approved):

PR #97, opened 2022-04-01
docs: update README with instructions for running

olavloite:
Updates README with additional instructions for how to run PGAdapter:
* Using a pre-built Docker image.
* Using a pre-built uber-jar.

Also adds separate documentation files for commonly used tools/drivers with specific instructions for those tools/drivers. The latter is to a certain degree a placeholder for more documentation to come.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,41 +1,18 @@
 # Google Cloud Spanner PostgreSQL Adapter
 
-## Introduction
-Spanner-PGAdapter is a simple, MITM, forward, non-transparent proxy, which 
-translates Postgres wire protocol into the Cloud Spanner equivalent. By running
-this proxy locally, any Postgres client (including the SQL command-line client
-PSQL) should function seamlessly by simply pointing its outbound port to the 
-this proxy's inbound port. For the time being, we only support PSQL versions 11, 12, 13, and 14.
+Google Cloud Spanner PGAdapter is a proxy which translates the PostgreSQL wire-protocol into the
+Cloud Spanner equivalent for Cloud Spanner databases [that use the PostgreSQL dialect](https://cloud.google.com/spanner/docs/postgresql-interface).
 
-Additionally to translation, this proxy also concerns itself with authentication
-and to some extent, connection pooling. Translation for the most part is simply
-a transformation of the [PostgreSQL wire protocol](https://www.postgresql.org/docs/8.2/protocol-message-formats.html)
-except for some cases concerning PSQL, wherein the query itself is translated.
+PGAdapter has been tested and can be used with the following clients:
+1. `psql`: Versions 11, 12, 13 and 14 are supported. See [psql support](docs/psql.md) for more details.
+2. `JDBC`: Versions 42.x and higher are partly supported. See [JDBC support](docs/jdbc.md) for more details.
+3. `pgx`: Version 4.15 and higher have limited support. See [pgx support](docs/pgx.md) for more details.
 
-Simple query mode and extended query mode are supported, and any data type
-supported by Spanner is also supported. Items, tables and language not native to
-Spanner are not supported, unless otherwise specified.
-
-Though the majority of functionality inherent in most PostgreSQL clients
-(including PSQL and JDBC) are included out of the box, the following items are
-not supported:
-* Functions
-* COPY <table_name> TO ...
-* COPY <table_name> FROM <filename | PROGRAM program>
-* Prepared Statement DESCRIBE
-* SSL
-* PSQL meta-commands not included in this list (i.e.: these are supported):
-  * `\d <table>` 
-  * `\dt <table>`
-  * `\dn <table>`
-  * `\di <table>`
-  * `\l`
-
-COPY <table_name> FROM STDIN is supported.
+Other drivers and tools have not been specifically tested and are not known to work with PGAdapter.
 
 ## Usage
-The PostgreSQL adapter can be started both as a Docker container, a standalone process as well as an in-process server.
-
+The PostgreSQL adapter can be started both as a Docker container, a standalone process as well as an
+in-process server (the latter is only supported for Java applications).
 
 ### Docker
 
@@ -57,12 +34,33 @@ The option `-v /local/path/to/credentials.json:/tmp/keys/key.json:ro` mounts the
 your local system to a file in the Docker container. The `-e GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/key.json`
 sets the file as the default credentials to use inside the container.
 
-See [Options](#Options) for an explanation of all further options. 
+The `-x` option tells PGAdapter to accept connections other hosts than `localhost`. This is required
+when running PGAdapter in a Docker container, as requests will not be seen as coming from `localhost`
+in the container. Choose a different host port than 5432 if you already have PostgreSQL running on
+your local system.
 
-### Standalone
+See [Options](#Options) for an explanation of all further options.
+
+### Standalone with pre-built jar
+
+A pre-built jar containing all dependencies can be downloaded from https://storage.googleapis.com/pgadapter-jar-releases/pgadapter-<VERSION>.jar
+
+Example (replace `v0.2.1` with the version you want to download):
+
+```shell
+wget https://storage.googleapis.com/pgadapter-jar-releases/pgadapter-v0.2.1.jar
+java -jar pgadapter-v0.2.1.jar -p my-project -i my-instance -d my-database
+```
+
+Use the `-s` option to specify a different local port than the default 5432 if you already have
+PostgreSQL running on your local system.
+
+See [Options](#Options) for an explanation of all further options.
+
+### Standalone with locally built jar
 1. Build a jar file containing all dependencies by running `mvn package -P shade`.
 2. Execute `java -jar <jar-file> <options>`.
-3. To get fine-grained logging messages, make sure that you have the logging.properties file and run the jar with `-Djava.util.logging.config.file=logging.properties`. You need to create one according to this sample if it's missing. 
+3. To get fine-grained logging messages, make sure that you have the logging.properties file and run the jar with `-Djava.util.logging.config.file=logging.properties`. You need to create one according to this sample if it's missing.
     ```
     handlers=java.util.logging.ConsoleHandler,java.util.logging.FileHandler
     com.google.cloud.spanner.pgadapter.level=FINEST
@@ -76,8 +74,12 @@ See [Options](#Options) for an explanation of all further options.
     java.util.logging.FileHandler.formatter=java.util.logging.SimpleFormatter
     ```
 
+See [Options](#Options) for an explanation of all further options.
+
 ### In-process
-1. Add google-cloud-spanner-pgadapter as a dependency to your project.
+This option is only available for Java/JVM-based applications.
+
+1. Add `google-cloud-spanner-pgadapter` as a dependency to your project.
 2. Build a server using the `com.google.cloud.spanner.pgadapter.ProxyServer`
    class:
 
@@ -86,9 +88,9 @@ class PGProxyRunner {
     public static void main() {
         ProxyServer server = new ProxyServer(
           new OptionsMetadata(
-                "jdbc:cloudspanner:/projects/my-project-name"
-                + "/instances/my-instance-id"
-                + "/databases/my-database-name"
+                "jdbc:cloudspanner:/projects/my-project"
+                + "/instances/my-instance"
+                + "/databases/my-database"
                 + ";credentials=/home/user/service-account-credentials.json",
                 portNumber,
                 textFormat,
@@ -108,7 +110,7 @@ path; All other items map directly to previously mentioned CLI options.
 
 ### Options
 The following options are required to run the proxy:
-  
+
 ```    
 -p <projectname>
   * The project name where the desired Spanner database is running.
@@ -120,16 +122,17 @@ The following options are required to run the proxy:
   * The Spanner database name.
 
 -c <credentialspath>
-  * The full path for the file containing the service account credentials in JSON 
-    format.
-  * Do remember to grant the service account sufficient credentials to access the
-    database.
+  * This is only required if you have not already set up default credentials on the system where you
+    are running PGAdapter.
+  * The full path for the file containing the service account credentials in JSON format.
+  * Do remember to grant the service account sufficient credentials to access the database.
 ```
 
 The following options are optional:
 ```    
 -s <port>
-  * The inbound port for the proxy. Defaults to 5432.
+  * The inbound port for the proxy. Defaults to 5432. Choose a different port if you already have
+    PostgreSQL running on your local system.
    
 -a
   * Use authentication when connecting. Currently authentication is not strictly
@@ -145,11 +148,18 @@ The following options are optional:
     performance penalties due to query matching and translation and as such is
     not recommended for production. It is further not guaranteed to perfectly
     match PSQL logic. Please only use this mode when using PSQL.
+    
+-jdbc
+  * JDBC Mode. Use this option when fronting the native PostgreSQL JDBC driver.
+    This option will inspect queries to look for native JDBC metadata queries and
+    replace these with queries that are suppported by Cloud Spanner PostgreSQL
+    databases.
 
 -f <POSTGRESQL|SPANNER>
   * The data result format coming back to the client from the proxy. By default,
     this is POSTGRESQL, but you can choose SPANNER format if you do not wish the
-    data to be modified and the client used can handle it.
+    data to be modified and the client used can handle it. This is generally only
+    useful when using `psql`.
     
 -c <multi-statementcommand>
     Runs a single command before exiting. This command can be comprised of multiple 
@@ -226,6 +236,44 @@ java -jar <jar-file> -p <project name> -i <instance id> -d <database name> -c
 <path to credentials file> -s 5432 
 ```
 
+## Details
+Google Cloud Spanner PGAdapter is a simple, MITM, forward, non-transparent proxy, which translates
+the PostgreSQL wire-protocol into the Cloud Spanner equivalent. It can only be used with Cloud
+Spanner databases that use the [PostgreSQL dialect](https://cloud.google.com/spanner/docs/postgresql-interface).
+By running this proxy locally, any PostgreSQL client (including the SQL command-line client PSQL)
+should function seamlessly by simply pointing its outbound port to this proxy's inbound port.
+The proxy does not support all parts of the PostgreSQL wire-protocol. See [Limitations](#Limitations)
+for a list of current limitations.
+
+In addition to translation, this proxy also concerns itself with authentication
+and to some extent, connection pooling. Translation for the most part is simply
+a transformation of the [PostgreSQL wire protocol](https://www.postgresql.org/docs/current/protocol-message-formats.html)
+except for some cases concerning PSQL, wherein the query itself is translated.
+
+Simple query mode and extended query mode are supported, and any data type
+supported by Spanner is also supported. Items, tables and language not native to
+Spanner are not supported, unless otherwise specified.
+
+Though the majority of functionality inherent in most PostgreSQL clients
+(including PSQL and JDBC) are included out of the box, the following items are
+not supported:
+* Prepared Statement DESCRIBE
+* SSL
+* Functions
+* COPY <table_name> TO ...
+* COPY <table_name> FROM <filename | PROGRAM program>
+
+COPY <table_name> FROM STDIN __is supported__.
+
+Only the following `psql` meta-commands are supported:
+  * `\d <table>` 
+  * `\dt <table>`
+  * `\dn <table>`
+  * `\di <table>`
+  * `\l`
+
+Other `psql` meta-commands are __not__ supported.
+
 ## COPY support
 `COPY <table-name> FROM STDIN` is supported. This option can be used to insert bulk data to a Cloud
 Spanner database. `COPY` operations are atomic by default, but the standard transaction limits of
@@ -290,14 +338,18 @@ psql -h localhost -p 5432 -d my-local-db \
   -c "copy numbers from stdin;"
 ```
 
+## Limitations
+
+PGAdapter has the following known limitations at this moment:
+- PostgreSQL prepared statements are not fully supported. It is recommended not to use this feature.
+  Note: This limitation relates specifically to server side PostgreSQL prepared statements.
+  The JDBC `java.sql.PreparedStatement` interface is supported.
+- DESCRIBE statement wire-protocol messages currently return `Oid.UNSPECIFIED` for all parameters in the query.
+- DESCRIBE statement wire-protocol messages return `NoData` as the row description of the statement.
+- The COPY protocol only supports COPY FROM STDIN.
 
 ## Support Level
 
 We are not currently accepting external code contributions to this project. 
 Please feel free to file feature requests using GitHub's issue tracker or 
 using the existing Cloud Spanner support channels.
-
-## Note
-
-Currently PGAdapter is in public preview. Please [sign-up](https://goo.gle/PostgreSQL-interface) 
-to get access to the Cloud Spanner PostgreSQL interface.
--- a/README.md
+++ b/README.md
@@ -116,6 +116,9 @@ information regarding project id, instance id, database name, credentials file
 path; All other items map directly to previously mentioned CLI options.
 
 ### Options
+
+#### Required
+
 The following options are required to run the proxy:
 
 ```    
@@ -135,7 +138,10 @@ The following options are required to run the proxy:
   * Do remember to grant the service account sufficient credentials to access the database.
 ```
 
+#### Optional
+
 The following options are optional:
+
 ```    
 -s <port>
   * The inbound port for the proxy. Defaults to 5432. Choose a different port if you already have
--- a/README.md
+++ b/README.md
@@ -1,20 +1,15 @@
-# Google Cloud Spanner PostgreSQL Adapter
+# Google Cloud Spanner PGAdapter
 
-Google Cloud Spanner PGAdapter is a proxy which translates the PostgreSQL wire-protocol into the
-Cloud Spanner equivalent for Cloud Spanner databases [that use the PostgreSQL dialect](https://cloud.google.com/spanner/docs/postgresql-interface).
+PGAdapter is a proxy which translates the PostgreSQL wire-protocol into the
+equivalent for Spanner databases [that use the PostgreSQL interface](https://cloud.google.com/spanner/docs/postgresql-interface).
 
-PGAdapter has been tested and can be used with the following clients:
+PGAdapter can be used with the following clients:
 1. `psql`: Versions 11, 12, 13 and 14 are supported. See [psql support](docs/psql.md) for more details.
 2. `JDBC`: Versions 42.x and higher have __limited support__. See [JDBC support](docs/jdbc.md) for more details.
 3. `pgx`: Version 4.15 and higher have __limited support__. See [pgx support](docs/pgx.md) for more details.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,68 +1,73 @@
...
-this proxy locally, any Postgres client (including the SQL command-line client
-PSQL) should function seamlessly by simply pointing its outbound port to the 
-this proxy's inbound port. For the time being, we only support PSQL versions 11, 12, 13, and 14.
-
-Additionally to translation, this proxy also concerns itself with authentication
-and to some extent, connection pooling. Translation for the most part is simply
-a transformation of the [PostgreSQL wire protocol](https://www.postgresql.org/docs/8.2/protocol-message-formats.html)
-except for some cases concerning PSQL, wherein the query itself is translated.
+Google Cloud Spanner PGAdapter is a proxy which translates the PostgreSQL wire-protocol into the
+Cloud Spanner equivalent for Cloud Spanner databases [that use the PostgreSQL dialect](https://cloud.google.com/spanner/docs/postgresql-interface).
```
snehashah16: s/PostgreSQL wire-protocol into the
Cloud Spanner equivalent for Cloud Spanner databases/PostgreSQL wire-protocol into the Cloud Spanner databases/

"Cloud Spanner" seems redundant, maybe remove ?

snehashah16: s/dialect/interface

dialect is only internal terminology to refer to the PG project

README.md
```diff
@@ -1,68 +1,73 @@
...
 
-Simple query mode and extended query mode are supported, and any data type
-supported by Spanner is also supported. Items, tables and language not native to
-Spanner are not supported, unless otherwise specified.
+PGAdapter has been tested and can be used with the following clients:
```
snehashah16: s/has been tested/

README.md
```diff
@@ -1,68 +1,73 @@
...
-* SSL
-* PSQL meta-commands not included in this list (i.e.: these are supported):
-  * `\d <table>` 
-  * `\dt <table>`
-  * `\dn <table>`
-  * `\di <table>`
-  * `\l`
-
-COPY <table_name> FROM STDIN is supported.
+Other drivers and tools have __not been specifically tested__ and are not known to work with PGAdapter
```
snehashah16: remove

README.md
```diff
@@ -1,68 +1,73 @@
...
+and are therefore __not supported__. Drivers and tools that use [simple query protocol](https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.5.7.4)
+are likely to work with PGAdapter. Drivers and tools that use the [extended query protocol](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)
+are more likely to at least have some issues when used with PGAdapter.
```
snehashah16:
```
Drivers and tools that use [simple query protocol](https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.5.7.4)
are likely to work with PGAdapter. Drivers and tools that use the [extended query protocol](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)
are more likely to at least have some issues when used with PGAdapter.
```
While I appreciate the detail, it sends a confusing message to our customers.. While these may be not supported, we still think they may work ?

README.md
```diff
@@ -1,68 +1,73 @@
...
-translates Postgres wire protocol into the Cloud Spanner equivalent. By running
-this proxy locally, any Postgres client (including the SQL command-line client
-PSQL) should function seamlessly by simply pointing its outbound port to the 
-this proxy's inbound port. For the time being, we only support PSQL versions 11, 12, 13, and 14.
-
-Additionally to translation, this proxy also concerns itself with authentication
-and to some extent, connection pooling. Translation for the most part is simply
-a transformation of the [PostgreSQL wire protocol](https://www.postgresql.org/docs/8.2/protocol-message-formats.html)
-except for some cases concerning PSQL, wherein the query itself is translated.
+Google Cloud Spanner PGAdapter is a proxy which translates the PostgreSQL wire-protocol into the
```
snehashah16: we are interchangeably using PostgreSQL Adapter and PGAdapter.. Should we make it explicit here that they refer to the same thing ?

s/PGAdapter/PostgreSQL Adapter (also referred to as PGAdapter)/    <--- or just PGAdapter in parenthesis

README.md
```diff
@@ -1,68 +1,73 @@
...
+Other drivers and tools have __not been specifically tested__ and are not known to work with PGAdapter
+and are therefore __not supported__. Drivers and tools that use [simple query protocol](https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.5.7.4)
+are likely to work with PGAdapter. Drivers and tools that use the [extended query protocol](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)
+are more likely to at least have some issues when used with PGAdapter.
 
 ## Usage
-The PostgreSQL adapter can be started both as a Docker container, a standalone process as well as an in-process server.
-
+The PostgreSQL adapter can be started both as a Docker container, a standalone process as well as an
+in-process server (the latter is only supported for Java applications).
```
snehashah16, line 13: `(the latter is only supported for Java applications)`
isnt this obvious because it is a java application ?

README.md
```diff
@@ -1,68 +1,73 @@
...
 ### Docker
 
+See [running PGAdapter using Docker](docs/docker.md) for more examples for running PGAdapter in Docker.
+
 Replace the project, instance and database names and the credentials file in the example below to
 run PGAdapter from a pre-built Docker image.
 
 ```shell
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json
 docker pull us-west1-docker.pkg.dev/cloud-spanner-pg-adapter/pgadapter-docker-images/pgadapter
```
snehashah16: could we have a gcr.io/<project>/image link like we have one for the emulator?
https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/blob/master/README.md#via-pre-built-docker-image

snehashah16: `pgadapter-docker-images/`
also not a fan of this, so if we can remove it and just have it in parent folder (since the project is restricted to PGAdapter), that would look easy/clean for our customers

README.md
```diff
@@ -1,68 +1,73 @@
...
 ```
 
-The option `-v /local/path/to/credentials.json:/tmp/keys/key.json:ro` mounts the credentials file on
-your local system to a file in the Docker container. The `-e GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/key.json`
-sets the file as the default credentials to use inside the container.
+The option `-v ${GOOGLE_APPLICATION_CREDENTIALS}:${GOOGLE_APPLICATION_CREDENTIALS}:ro` mounts the
+default credentials file on your local system to a virtual file with the same name in the Docker container.
+The `-e GOOGLE_APPLICATION_CREDENTIALS` sets the file as the default credentials to use inside the container.
+
+The `-x` option tells PGAdapter to accept connections other hosts than `localhost`. This is required
```
snehashah16: *to* other hosts than `localhost`

README.md
```diff
@@ -1,68 +1,73 @@
...
 
 ### Docker
 
+See [running PGAdapter using Docker](docs/docker.md) for more examples for running PGAdapter in Docker.
+
 Replace the project, instance and database names and the credentials file in the example below to
 run PGAdapter from a pre-built Docker image.
 
 ```shell
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json
```
snehashah16, line 23: stretch ask:

just thinking out loud, how would customers set this up in an automated release or CI/CD pipelines where all this is done programatically

is there a IAM command utility that can pull your credential file ? 
also, we should link to "Setup" docs for Cloud Spanner where they discuss this in detail

README.md
```diff
@@ -1,68 +1,73 @@
...
   -x
 ```
 
-The option `-v /local/path/to/credentials.json:/tmp/keys/key.json:ro` mounts the credentials file on
-your local system to a file in the Docker container. The `-e GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/key.json`
-sets the file as the default credentials to use inside the container.
+The option `-v ${GOOGLE_APPLICATION_CREDENTIALS}:${GOOGLE_APPLICATION_CREDENTIALS}:ro` mounts the
+default credentials file on your local system to a virtual file with the same name in the Docker container.
+The `-e GOOGLE_APPLICATION_CREDENTIALS` sets the file as the default credentials to use inside the container.
+
```
snehashah16: remove line since -v & -e dont have one between them

README.md
```diff
@@ -1,68 +1,73 @@
...
+See [Options](#Options) for an explanation of all further options.
+
+### Standalone with pre-built jar
+
+A pre-built jar containing all dependencies can be downloaded from https://storage.googleapis.com/pgadapter-jar-releases/pgadapter-VERSION.jar
+
+Example (replace `v0.2.1` with the version you want to download):
+
+```shell
+VERSION=v0.2.1
```
snehashah16, line 53: could there be an automation to label the most recent binary as "latest" and then run a wget command that always fetches "latest" unless a version is specified ?

It's tedious to go find the latest version unless u plan to keep it updated here

README.md
```diff
@@ -1,68 +1,73 @@
...
+wget https://storage.googleapis.com/pgadapter-jar-releases/pgadapter-${VERSION}.jar
+java -jar pgadapter-${VERSION}.jar -p my-project -i my-instance -d my-database
```
snehashah16, line 55: same applies, if we have a latest version without the version tag  in the binary name, this becomes super easy to script and run within an automated CI/CD pipeline or release

README.md
```diff
@@ -1,68 +1,73 @@
...
+```
+
+Use the `-s` option to specify a different local port than the default 5432 if you already have
```
snehashah16: doesnt this apply to all modes of PGAdapter , irrespective of how it is started.. since it exists under the Options, we should remove from here

README.md
```diff
@@ -1,68 +1,73 @@
...
+PostgreSQL running on your local system.
 
-See [Options](#Options) for an explanation of all further options. 
+See [Options](#Options) for an explanation of all further options.
 
-### Standalone
+### Standalone with locally built jar
 1. Build a jar file containing all dependencies by running `mvn package -P shade`.
```
snehashah16: stretch ask: `-P shade`
curious if we can get rid of this ? I understand why we do it, though could our maven build file hide this requirement from customers ?

README.md
```diff
@@ -76,8 +81,12 @@ See [Options](#Options) for an explanation of all further options.
     java.util.logging.FileHandler.formatter=java.util.logging.SimpleFormatter
```
snehashah16: seems generic to all modes of running PGAdapter, remove ?

README.md
```diff
@@ -1,68 +1,73 @@
...
-See [Options](#Options) for an explanation of all further options. 
+See [Options](#Options) for an explanation of all further options.
 
-### Standalone
+### Standalone with locally built jar
 1. Build a jar file containing all dependencies by running `mvn package -P shade`.
 2. Execute `java -jar <jar-file> <options>`.
-3. To get fine-grained logging messages, make sure that you have the logging.properties file and run the jar with `-Djava.util.logging.config.file=logging.properties`. You need to create one according to this sample if it's missing. 
+3. To get fine-grained logging messages, make sure that you have the logging.properties file and run the jar with `-Djava.util.logging.config.file=logging.properties`. You need to create one according to this sample if it's missing.
     ```
```
snehashah16: have all code changes in shell / java / etc, so they are highlighted uniformly


```shell

```

README.md
```diff
@@ -76,8 +81,12 @@ See [Options](#Options) for an explanation of all further options.
     java.util.logging.FileHandler.formatter=java.util.logging.SimpleFormatter
     ```
 
+See [Options](#Options) for an explanation of all further options.
+
 ### In-process
-1. Add google-cloud-spanner-pgadapter as a dependency to your project.
+This option is only available for Java/JVM-based applications.
+
+1. Add `google-cloud-spanner-pgadapter` as a dependency to your project.
```
snehashah16, line 77: is it valuable to demonstrate how to do this in mvn or gradle ?

README.md
```diff
@@ -120,16 +132,20 @@ The following options are required to run the proxy:
   * The Spanner database name.
 
 -c <credentialspath>
-  * The full path for the file containing the service account credentials in JSON 
-    format.
-  * Do remember to grant the service account sufficient credentials to access the
-    database.
+  * This is only required if you have not already set up default credentials on the system where you
```
snehashah16, line 123: [set up default credential](link to documentation)

README.md
```diff
@@ -120,16 +132,20 @@ The following options are required to run the proxy:
...
+    are running PGAdapter.
+  * The full path for the file containing the service account credentials in JSON format.
+  * Do remember to grant the service account sufficient credentials to access the database.
```
snehashah16: `service account sufficient credentials to access the database`
link to documentation

README.md
```diff
@@ -145,11 +161,18 @@ The following options are optional:
...
     not recommended for production. It is further not guaranteed to perfectly
     match PSQL logic. Please only use this mode when using PSQL.
+    
+-jdbc
+  * JDBC Mode. Use this option when fronting the native PostgreSQL JDBC driver.
+    This option will inspect queries to look for native JDBC metadata queries and
+    replace these with queries that are suppported by Cloud Spanner PostgreSQL
+    databases.
 
 -f <POSTGRESQL|SPANNER>
```
snehashah16: can we hide this ? I am not sure we want our customers to understand and use this correctly

README.md
```diff
@@ -226,6 +249,44 @@ java -jar <jar-file> -p <project name> -i <instance id> -d <database name> -c
 <path to credentials file> -s 5432 
 ```
 
+## Details
+Google Cloud Spanner PGAdapter is a simple, MITM, forward, non-transparent proxy, which translates
+the PostgreSQL wire-protocol into the Cloud Spanner equivalent. It can only be used with Cloud
+Spanner databases that use the [PostgreSQL dialect](https://cloud.google.com/spanner/docs/postgresql-interface).
```
snehashah16: s/dialect/interface/

README.md
```diff
@@ -226,6 +249,44 @@ java -jar <jar-file> -p <project name> -i <instance id> -d <database name> -c
...
+Though the majority of functionality inherent in most PostgreSQL clients
+(including PSQL and JDBC) are included out of the box, the following items are
+not supported:
+* Prepared Statement DESCRIBE
+* SSL
+* Functions
+* COPY <table_name> TO ...
+* COPY <table_name> FROM <filename | PROGRAM program>
+
+COPY <table_name> FROM STDIN __is supported__.
```
snehashah16: this seems misplaced..

README.md
```diff
@@ -290,14 +351,18 @@ psql -h localhost -p 5432 -d my-local-db \
   -c "copy numbers from stdin;"
 ```
 
+## Limitations
+
+PGAdapter has the following known limitations at this moment:
+- PostgreSQL prepared statements are not fully supported. It is recommended not to use this feature.
```
snehashah16: link to the postgresql docs

snehashah16: nit- s/recommended not/not recommended/

README.md
```diff
@@ -1,68 +1,73 @@
...
+  -e GOOGLE_APPLICATION_CREDENTIALS \
+  us-west1-docker.pkg.dev/cloud-spanner-pg-adapter/pgadapter-docker-images/pgadapter \
   -p my-project -i my-instance -d my-database \
   -x
 ```
 
-The option `-v /local/path/to/credentials.json:/tmp/keys/key.json:ro` mounts the credentials file on
-your local system to a file in the Docker container. The `-e GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/key.json`
-sets the file as the default credentials to use inside the container.
+The option `-v ${GOOGLE_APPLICATION_CREDENTIALS}:${GOOGLE_APPLICATION_CREDENTIALS}:ro` mounts the
```
snehashah16: reading further down, maybe remove this to keep this simple.. this level of detail is provided in the docker.md anyways

docs/jdbc.md
```diff
@@ -0,0 +1,32 @@
...
+version 42.0.0 and higher.
+
+## Usage
+
+First start PGAdapter in JDBC mode. This example uses the pre-built jar with dependencies to run
+PGAdapter as a standalone process. See [README](../README.md) for more possibilities on how to run
+PGAdapter:
+
+```shell
+wget https://storage.googleapis.com/pgadapter-jar-releases/pgadapter-v0.2.1.jar
```
snehashah16: either use a VERSION env-var or i am in the favor of using latest tag

docs/pgx.md
```diff
@@ -0,0 +1,44 @@
+# Google Cloud Spanner PostgreSQL Adapter - pgx Support
+
+PGAdapter has limited support for the [Go pgx driver](https://github.com/jackc/pgx) version 4.15.0
+and higher. 
+
+## Usage
+
+First start PGAdapter:
```
snehashah16, line 8: do not duplicate this across docs.. rather link to the main README so that there is a variety of ways to "start" adapter

docs/psql.md
```diff
@@ -0,0 +1,52 @@
...
+docker run \
+  -d -p 5432:5432 \
+  -v ${GOOGLE_APPLICATION_CREDENTIALS}:${GOOGLE_APPLICATION_CREDENTIALS}:ro \
+  -e GOOGLE_APPLICATION_CREDENTIALS \
+  us-west1-docker.pkg.dev/cloud-spanner-pg-adapter/pgadapter-docker-images/pgadapter \
+  -p my-project -i my-instance -d my-database \
+  -x -q
+```
+
+Note:
```
snehashah16: favor of de-duping wherever possible

docs/psql.md
```diff
@@ -0,0 +1,52 @@
...
+```shell
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json
+docker pull us-west1-docker.pkg.dev/cloud-spanner-pg-adapter/pgadapter-docker-images/pgadapter
+docker run \
+  -d -p 5433:5432 \
+  -v ${GOOGLE_APPLICATION_CREDENTIALS}:${GOOGLE_APPLICATION_CREDENTIALS}:ro \
+  -e GOOGLE_APPLICATION_CREDENTIALS \
+  us-west1-docker.pkg.dev/cloud-spanner-pg-adapter/pgadapter-docker-images/pgadapter \
+  -p my-project -i my-instance -d my-database \
+  -x -q
```
snehashah16: can u highlight -q 

all other options are common to all tools it seems and therefore should be de-duped

snehashah16 (commented):
a ton of `nit` comments.. 

generally, lets keep it clean and easy to maintain.

README.md
snehashah16: stretch ask: could we link to a sample demo app that can get running with the latest version of the adapter ?

README.md
olavloite: Will be added in this PR: https://github.com/GoogleCloudPlatform/pgadapter/pull/102 PTAL

README.md
olavloite: For anyone familiar with Java: Yes.
For someone not at all familiar with Java: Not necessarily, and that is the audience for this comment.

README.md
olavloite: Done

olavloite: Done

README.md
olavloite: Done

README.md
olavloite: Done

README.md
olavloite: I've removed the section. I agree that it might be a little confusing.

The situation is that the simple query protocol is, as the name already says, simple. This means that we already support a lot more of that protocol than the extended protocol, which means that tools that use that protocol by default are a lot more likely to work out of the box than tools that use the extended mode. But it's probably better not to advertise that here, as it could confuse users, especially those who do not know what the simple/extended protocol is. Those who do know the difference, might automatically be inclined to try it out.

README.md
olavloite: I've changed it to PGAdapter everywhere. I think that is clearer than using two names interchangeably.

README.md
olavloite: That is unfortunately not possible in Artifact Registry. It requires us to define a folder like this, as Artifact Registry can contain different types of artifacts, and this folder holds metadata about the types of artifacts inside it. In other words: Top-level artifacts are not supported by Artifact Registry.

README.md
olavloite: Good catch. But I think it should be _from_ other hosts than `localhost` in this case, as this relates to the incoming connection requests to PGAdapter.

README.md
olavloite: I've added a link to the getting started page for authentication setup.

Regarding CI/CD pipelines etc.: I think we need (a) separate document(s) for that, as there are a number of different options, and obviously also a lot of different CI/CD solutions.

> is there a IAM command utility that can pull your credential file ?
It must be executed inside the container in that case, and the container does not have access to anything on the host machine unless it has been explicitly given access to it when starting up (which is what this example does with the file that is pointed to by `GOOGLE_APPLICATION_CREDENTIALS`).

So the short version is: There is no generic solution that will work on all hosts for getting the standard credentials from the host environment to the Docker container. The current solution is AFAICT the closest that we can get to a generic solution, as GOOGLE_APPLICATION_CREDENTIALS will be an environment variable that is already set on many systems.

README.md
olavloite: I think we should keep the extra empty line here, as it creates a small section break. The reason that there is no break between the `-e` and `-v` options above is that they are logically tied together, as you need both together to specify the credentials that are used by the container.

README.md
olavloite: Good point. There's a PR here that will fix that: https://github.com/GoogleCloudPlatform/pgadapter/pull/103

olavloite: (I've already manually uploaded a -latest.jar, so it is available as of now)

README.md
olavloite: Modified this example to use `-latest.jar` and added a separate section for downloading a specific version.

README.md
olavloite: I think we should keep it here, as it is a very typical problem that (first time) users will run into when running the uber-jar. Many people who will try this out will already have a local PG installation, and I think that this section should be as complete as possible for anyone who wants to try it out, without having to go through the complete list of options to figure out how to use a custom port. Especially as PGAdapter uses `-s` instead of `-p` to specify the port to use.

README.md
olavloite: That is technically possible, but I'm not sure it is really what we want in this case. This option is realistically only going to be used by customers who are already in the Java ecosystem, or at least are familiar with it. It is customary to let `mvn package` build a thin jar, and have a specific profile for building a thick jar.

Building a fat jar directly in the default profile (so without the `-P shade` flag) would mean that we would have to choose between:
1. Name the fat jar something like `pgadapter-with-dependencies.jar` (`pgadapter.jar` will by default be the thin jar)
2. OR: Overwrite the thin jar with the fat jar, meaning that customers cannot get a thin jar from the default profile.
3. OR: Rename the thin jar to something like `pgadapter-thin-jar.jar` and let the thick jar be named `pgadapter.jar`.

I feel that the current solution with a specific profile for building a fat jar is the most idiomatic for the Java ecosystem.

README.md
olavloite: I've moved it to a separate section, including a specific section for Docker as that requires some volume mapping.

README.md
olavloite: Done

README.md
olavloite: Regarding mvn/gradle sample: Yes, that would certainly be valuable. But I'm awaiting approval of the CL that sets up the kokoro job config that publishes it to Maven central, so we have to wait for that before we can add an actual sample.

Regarding demo app: Agree, but also for that it needs to be in Maven central first.

README.md
olavloite: Added. Note that the [name of link](actual link) format is not supported inside code blocks. So either we need to accept that the link here is shown as the full URL, or we need to reformat the block as not being a code block.

README.md
olavloite: Done

README.md
olavloite: Good point, done.

README.md
olavloite: Done

README.md
olavloite: Any suggestions on how we should rephrase this? I'm reluctant to completely remove it, as I think many people will read the line above quickly and come to the conclusion that `COPY .. FROM` is not supported, while it is specifically `COPY .. FROM <filename | PROGRAM program` that is not supported.

README.md
olavloite: Done

olavloite: Done

README.md
olavloite: Agreed. I've removed all the additional details here from the README and added a link to the specific Running in Docker doc file.

docs/jdbc.md
olavloite: Replaced with -latest.

docs/pgx.md
olavloite: I've added this to all specific documents as it is likely that we at least temporarily will need specific options for specific drivers. For pgx we currently do not have that (although it is very likely that it will change). I've tried to use the most likely start option for each tool (e.g. Docker when using pgx, as it is outside the Java ecosystem, fat-jar when using JDBC)

docs/psql.md
olavloite: See above. I think it's better to have this here to be able to provide a full example of starting PGAdapter in psql-mode.

docs/psql.md
olavloite: I've reorganized the jdbc, pgx and psql documents a little to try to highlight the specific options that are needed for the specific tool. I don't think it's good to deduplicate the examples by removing the parameters that are common, as people generally want commands that they can copy-paste directly.

README.md
snehashah16: https://cloud.google.com/container-registry/docs/pushing-and-pulling - it demonstrates the process to create the gcr.io version (using cloud storage buckets)

README.md
snehashah16: ack

README.md
snehashah16: Thank u

snehashah16: i think this was resolved by PR#102

README.md
snehashah16: ack

README.md
snehashah16: sent an email .. we can remove the -latest suffix too.. all pgadapter.jar are pointing to the latest version and then there is also a versioned copy

README.md
snehashah16: leaving this open to remove -latest suffix

README.md
snehashah16: ack

README.md
snehashah16: ack

README.md
snehashah16: ack, closing this for u to track these requests internally

README.md
snehashah16: may be add a link to the supported versions of "COPY" below ?

snehashah16 (approved):
minor comments, LGTM otherwise

README.md
olavloite: > think this was resolved by PR#102

Yes, that's correct.

README.md
olavloite: I've removed the suffix in #103 and in the references in the README and jdbc documentation file.

README.md
olavloite: -latest.jar suffix removed.

README.md
olavloite: Agreed.

PR #102, opened 2022-04-06
build: add gcr.io alias for Docker images

olavloite:
Adds a `gcr.io` alias for Docker images. This will make the Docker image available at `gcr.io/cloud-spanner-pg-adapter/pgadapter`.

### DIFF
--- a/.github/workflows/docker-build-and-push.yaml
+++ b/.github/workflows/docker-build-and-push.yaml
@@ -12,8 +12,8 @@ env:
   GITHUB_SHA: ${{ github.sha }}
   GITHUB_REF: ${{ github.ref }}
   IMAGE: pgadapter
-  # REGISTRY_HOSTNAME: us-west1-docker.pkg.dev/cloud-spanner-pg-adapter/pgadapter-docker-images
-  REGISTRY_HOSTNAME: gcr.io/cloud-spanner-pg-adapter
+  ARTIFACT_REGISTRY_HOSTNAME: us-west1-docker.pkg.dev/cloud-spanner-pg-adapter/pgadapter-docker-images
+  GCR_HOSTNAME: gcr.io/cloud-spanner-pg-adapter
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build-and-push.yaml
```diff
@@ -48,6 +51,9 @@ jobs:
         run: |
           export TAG=`echo $GITHUB_REF | awk -F/ '{print $NF}'`
           echo $TAG
-          docker push "$REGISTRY_HOSTNAME"/"$IMAGE":"$TAG"
-          docker tag "$REGISTRY_HOSTNAME"/"$IMAGE":"$TAG" "$REGISTRY_HOSTNAME"/"$IMAGE":latest
-          docker push "$REGISTRY_HOSTNAME"/"$IMAGE":latest
+          docker push "$ARTIFACT_REGISTRY_HOSTNAME"/"$IMAGE":"$TAG"
```
snehashah16, line 54: docker push "$ARTIFACT_REGISTRY_HOSTNAME"/"$IMAGE":"$TAG"
          docker tag "$ARTIFACT_REGISTRY_HOSTNAME"/"$IMAGE":"$TAG" "$ARTIFACT_REGISTRY_HOSTNAME"/"$IMAGE":latest
          docker push "$ARTIFACT_REGISTRY_HOSTNAME"/"$IMAGE":latest



do we still need these ?

snehashah16 (approved):

.github/workflows/docker-build-and-push.yaml
olavloite: I'm afraid that someone might already have sent those links around, so I'm a little reluctant to remove them right away.

I'll open a separate PR for it right away, so we can remove it quickly if we want to. (Note: A new image is only pushed if we do a release. We don't push a new version for each commit that is done.)

PR #103, opened 2022-04-06
build: upload jar with version suffix + plain pgadapter.jar

olavloite:
The most recent uber-jar will be available at https://storage.googleapis.com/pgadapter-jar-releases/pgadapter.jar


snehashah16:
could we remove the -latest suffix ?

olavloite:
> could we remove the -latest suffix ?

Done

snehashah16 (approved):
Thank u!

PR #106, opened 2022-04-07
test: use simple DatabaseAdminServiceImpl for tests

olavloite:
Use a simple `DatabaseAdminServiceImpl` instance for tests to allow easy adding of exceptions for invalid DDL statements.

### DIFF
--- a/src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
+++ b/src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
@@ -17,6 +17,7 @@ package com.google.cloud.spanner.pgadapter;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
```diff
@@ -405,4 +417,18 @@ public void testDdl() throws SQLException {
     assertEquals(1, updateDatabaseDdlRequests.get(0).getStatementsCount());
     assertEquals(sql, updateDatabaseDdlRequests.get(0).getStatements(0));
   }
+
+  @Test
+  public void testInvalidDdl() throws SQLException {
+    String sql = "CREATE TABLE foo (id int64 primary key)";
+    mockDatabaseAdmin.addException(
```
olavloite, line 424: @libingye816 The mock server that we were using for database admin operations (DDL) did not support adding exceptions (and also did a lot of other things that we don't need). This PR replaces that mock server with a much simpler one that also allows us to just add an exception when we want it to fail.

olavloite:
Closing as this is superseded by #108

PR #108, opened 2022-04-08
feat: all statement batching

libingye816:
Support batching of all statement types. Response messages are sent for each statement in the batch. 
Cover most of the cases (not only the short-term solution to python driver) with the exception:
1. The remaining statements in a dml/ddl batch are skipped and throw exceptions if the dml/ddl batch fails in the middle. The preceding statements in dml/ddl batch may not take effect even though they succeed. The whole batch resumes after the dml/ddl batch.


.ci/e2e-expected/ddl-dml-batching.txt
```diff
@@ -1 +1 @@
-ERROR: FAILED_PRECONDITION: Executing updates is not allowed for DDL batches. "CREATE TABLE testtable (id integer PRIMARY KEY, data integer); INSERT INTO testtable (id, data) VALUES (1, 2); INSERT INTO testtable (id, data) VALUES (3, 4)"
+INSERT 0 1
```
olavloite, line 1: Shouldn't the output of this batch be:
```
CREATE
INSERT 0 1
INSERT 0 1
```

.ci/e2e-expected/dml-batching.txt
```diff
@@ -1 +1 @@
-INSERT 0 5
+INSERT 0 1
```
olavloite, line 1: The output of this batch should be:
```
INSERT 0 1
INSERT 0 1
INSERT 0 1
INSERT 0 1
INSERT 0 1
```

.ci/e2e-expected/select-ddl-batching.txt
```diff
@@ -1 +1,4 @@
-ERROR: FAILED_PRECONDITION: Executing queries is not allowed for DDL batches. "CREATE TABLE test (id integer PRIMARY KEY); SELECT * FROM test; SELECT * FROM test"
```
olavloite, line 1: Also here: It seems that the current implementation only returns the result of the last statement in the batch.

(And in this case it also misses the `SELECT <rowcount>` that should also follow after a SELECT statement, but I'm not sure we do that at all for non-batched statements)

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -386,7 +388,8 @@ public synchronized void setStatus(ConnectionStatus status) {
     UNAUTHENTICATED,
     IDLE,
     COPY_IN,
-    TERMINATED
+    TERMINATED,
+    ABORTED
```
olavloite: nit: Could we rename this to `TRANSACTION_ABORTED`. This sounds more like the connection itself is in an aborted state, which would normally be something that you cannot recover from.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
```diff
@@ -288,15 +286,14 @@ private int queryIndexedColumnsCount(Set<String> columnNames) {
     return 0;
   }
 
-  @Override
```
olavloite, line 291: This worries me a bit: This was a method that was defined on one of the super classes of this class. This means that it is possible that the method was called from generic error handling code, for example in `ConnectionHandler`. That will no longer be the case, and that might mean that the `executor.shutdownNow()` will not be called. Can we add an implementation of `handleExecutionException(int index, SpannerException exception)` to this class that first calls `executor.shutdownNow()` and then calls the super implementation?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -95,24 +95,37 @@ protected ParsedStatement replaceKnownUnsupportedQueries(ParsedStatement parsedS
...
-      return ResultType.UPDATE_COUNT;
-    } else if (parsedStatement.isQuery()) {
-      return ResultType.RESULT_SET;
-    } else {
-      return ResultType.NO_RESULT;
+  protected static ImmutableList<StatementType> determineStatementTypes(
+      ImmutableList<String> statements) {
+    ImmutableList.Builder<StatementType> builder = ImmutableList.builder();
+    for (String sql : statements) {
+      if (PARSER.isUpdateStatement(sql)) {
```
olavloite: It's more efficient to call `PARSER.parse(String sql)` and then call the `ParsedStatement#isQuery()` etc. methods. The current implementation will repeatedly parse the SQL string to check whether it is an update, query or DDL statement.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -95,24 +95,37 @@ protected ParsedStatement replaceKnownUnsupportedQueries(ParsedStatement parsedS
...
+        builder.add(StatementType.DDL);
+      } else {
+        builder.add(StatementType.CLIENT_SIDE);
+      }
     }
+    return builder.build();
   }
 
   // Split statements by ';' delimiter, but ignore anything that is nested with '' or "".
+
```
olavloite: nit: remove

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -168,45 +181,66 @@ public boolean isBound() {
...
   }
 
   /** @return The number of items that were modified by this execution. */
-  public Long getUpdateCount() {
-    return this.updateCount;
+  public long getUpdateCount(int index) {
+    if (this.updateCounts == null) {
+      return 0;
+    }
+    return this.updateCounts[index];
```
olavloite, line 227: I think that this is incorrect. This assumes that if `this.updateCounts != null` that each statement has an update count and will return 0 (zero) instead of `null` for statements in the batch that are not an update when the first statement was an update. I think you need to change the type of `updateCounts` from `long[]` to `Long[]`.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -168,45 +181,66 @@ public boolean isBound() {
...
   }
 
-  public void addUpdateCount(long count) {
-    if (this.updateCount == null) {
-      this.updateCount = 0L;
+  public void addUpdateCount(int index, long count) {
```
olavloite: nit: rename this to `setUpdateCount`. The `addUpdateCount` was reasonable when we were only returning one result for the entire batch, but that should not be the case anymore.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -168,45 +181,66 @@ public boolean isBound() {
...
+    if (this.updateCounts == null) {
+      this.updateCounts = new long[getStatementCount()];
     }
-    this.updateCount += count;
+    this.updateCounts[index] += count;
```
olavloite:
```suggestion
    this.updateCounts[index] = count;
```

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: I just noticed the `setUpdateCount` below. Maybe we even need to remove this entire method. It should at least somehow be clear that `addUpdateCount` does not actually add a count to an existing update count.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -168,45 +181,66 @@ public boolean isBound() {
...
+  }
+
+  private void setUpdateCount(int index, long updateCount) {
```
olavloite: I think this should be the only method here (so it should be made public). We only __set__ the update count for a specific index once.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +285,195 @@ public Exception getException() {
...
-          connection.startBatchDdl();
-        } else {
-          throw SpannerExceptionFactory.newSpannerException(
-              ErrorCode.INVALID_ARGUMENT, "Statement type is not supported for batching");
-        }
-        for (String stmt : statements) {
-          connection.execute(Statement.of(stmt));
+    if (connectionHandler.getStatus() == ConnectionStatus.ABORTED
+        && !"ROLLBACK".equals(getCommand(0))) {
+      lastExecutedIndex = 0;
```
olavloite: This is good for now, considering the note in the PR description, so no need to take any action at this point. But we probably need to change this later, as we should not return early if a batch is received while the connection is in an aborted transaction:

1. Assume that the connection is in an aborted transaction.
2. We receive the following batch: `update table1 set value=1; commit;`
3. The return values of this batch should be `ERROR:  current transaction is aborted, commands ignored until end of transaction block`, `ROLLBACK`
4. The state of the connection after executing the above batch should be `Idle`

See also the discussion in the design doc regarding what should happen if we receive a `COMMIT` while in an aborted transaction. `COMMIT` should be accepted, but should `ROLLBACK`  the transaction and return `ROLLBACK` as the command tag.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +285,195 @@ public Exception getException() {
...
+      handleExecutionException(
+          0,
+          SpannerExceptionFactory.newSpannerException(
+              ErrorCode.INVALID_ARGUMENT,
+              "Transaction is aborted and must be rolled back prior to further execution."));
```
olavloite: This is the literal error message that is returned by PostgreSQL

```suggestion
              "ERROR: current transaction is aborted, commands ignored until end of transaction block"));
```

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +285,195 @@ public Exception getException() {
...
+    lastExecutedIndex = index - 1;
+  }
+
+  private void executeSingleStatement(int index) {
+    // Before executing the statement, handle specific statements that change the transaction status
+    String command = getCommand(index);
+    if ("BEGIN".equals(command)) {
+      if (executionStatus == ExecutionStatus.EXPLICIT_TRANSACTION) {
+        // Executing a BEGIN statement when an explicit transaction is already active is a no-op
+        return;
```
olavloite, line 395: Although it is a no-op, we should return a command tag to indicate that it was 'executed'. So we should return `BEGIN` to the client in this case.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +285,195 @@ public Exception getException() {
...
+        // Executing a BEGIN statement when an implicit transaction is active causes the implicit
+        // transaction to be committed and a new explicit transaction to be started
+        connection.commit();
       }
+      executionStatus = ExecutionStatus.EXPLICIT_TRANSACTION;
+    }
+    if ("COMMIT".equals(command) || "ROLLBACK".equals(command)) {
+      executionStatus = ExecutionStatus.AUTOCOMMIT;
+    }
+    if ("ROLLBACK".equals(command) && connectionHandler.getStatus() == ConnectionStatus.ABORTED) {
```
olavloite: See also discussion in design doc: We need to add `COMMIT` to this check. And we need to make sure that we return `ROLLBACK` as the command tag to the client, and not `COMMIT`.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +285,195 @@ public Exception getException() {
...
+      updateResultCount(index, result);
     } catch (SpannerException e) {
-      if (statements.size() > 1) {
-        SpannerException exception =
-            SpannerExceptionFactory.newSpannerException(
-                e.getErrorCode(),
-                e.getMessage() + " \"" + this.parsedStatement.getSqlWithoutComments() + "\"",
-                e);
-        handleExecutionException(exception);
+      handleExecutionExceptionInTransaction(index, e);
```
olavloite: Don't we need to check if the connection is in autocommit here? Or is that for a later PR?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +285,195 @@ public Exception getException() {
...
+      updateBatchResultCount(fromIndex, counts);
+      handleExecutionExceptionInTransaction(fromIndex + counts.length, e);
+      // TODO: May want to continue executing from the failure in order to be compatible with the
+      // behavior of PG when there is no transaction.
+      executionStatus = ExecutionStatus.HALTED;
+    }
+    return index - fromIndex;
+  }
+
+  private boolean canBeBatchedTogether(StatementType statement1, StatementType statement2) {
```
olavloite: nit: 
```suggestion
  private boolean canBeBatchedTogether(StatementType statementType1, StatementType statementType2) {
```

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/DescribeMessage.java
```diff
@@ -84,15 +84,16 @@ protected int getHeaderLength() {
...
+    if (this.statement.hasException(0)) {
+      this.handleError(this.statement.getException(0));
     } else {
-      switch (this.statement.getResultType()) {
-        case UPDATE_COUNT:
-        case NO_RESULT:
+      switch (this.statement.getStatementType(0)) {
+        case UPDATE:
+        case DDL:
+        case CLIENT_SIDE:
```
olavloite, line 93: Client side statements can have a result. For example the client side statement `SHOW SPANNER.COMMIT_TIMESTAMP`  will return a result set containing the commit timestamp of the last transaction. So I think you might need to refactor it back to keeping track of the result type instead of the statement type.

There is also another (future) case where the statement type will not necessarily always be enough to determine the result type. DML statements may in the future allow a `RETURNING` clause, which again will make the DML statement return a result set.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
```diff
@@ -116,21 +119,29 @@ public void handleQuery() throws Exception {
...
+              .send(false);
+        }
+        this.sendSpannerResult(index, this.statement, QueryMode.SIMPLE, 0L);
+      }
+    }
+    boolean inTransaction = connection.getSpannerConnection().isInTransaction();
+    Status transactionStatus = Status.IDLE;
+    if (inTransaction) {
+      if (connection.getStatus() == ConnectionStatus.ABORTED) {
+        transactionStatus = Status.FAILED;
```
olavloite, line 139: Note: I'm not sure exactly where (and if) this is currently handled in this PR, but if the connection status is set to `ConnectionStatus.ABORTED`, then the actual read/write transaction in Spanner should be rollbacked. We will not be using that transaction anymore, and we should therefore also not hold on to it until we receive a `ROLLBACK` from the client, but actively execute a rollback before returning `ConnectionStatus.ABORTED` to the client.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
```diff
@@ -527,4 +527,80 @@ public void testCopyIn_Large_SucceedsWhenNonAtomic() throws SQLException, IOExce
...
   }
+
+  @Test
+  public void testTwoDmlStatementsInSimpleMode() throws SQLException {
+    // This connection prefers the simple query mode. This makes the JDBC driver behave very similar
+    // to psql, and enables us to send multiple simple unparameterized statements in one batch.
+    try (Connection connection =
+        DriverManager.getConnection(
+            String.format(
+                "jdbc:postgresql://localhost:%d/?preferQueryMode=simple", server.getLocalPort()))) {
```
olavloite: `ITJdbcTest` is parameterized and will  execute all test cases twice; once in simple mode and once in extended mode. This implementation causes this test to be executed twice in simple mode (which is what the test name also indicates). However; the result that we receive from the PostgreSQL JDBC driver should (in most cases) be the same regardless whether the connection is in simple or extended mode, so we should change this test case so that it either:

* Is executed in both modes, and ensure that the result is the same regardless of the mode that is used.
* OR: Only execute it is simple mode once. Do this by calling `assumeTrue("simple".equals(preferQueryMode), "Batching is only supported in simple mode")` at the first line of the test case. This will make JUnit skip the test for extended mode.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
```diff
@@ -527,4 +527,80 @@ public void testCopyIn_Large_SucceedsWhenNonAtomic() throws SQLException, IOExce
...
+          assertEquals("Two - Updated", resultSet.getString("name"));
+          assertFalse(resultSet.next());
+        }
+      }
+    }
+  }
+
+  @Test
+  public void testDmlAndQueryInBatchInSimpleMode() throws SQLException {
+    // This connection prefers the simple query mode. This makes the JDBC driver behave very similar
```
olavloite: Same as above: Either execute this in both simple and extended mode, or add an `assumeTrue(..)` statement at the first line to skip it in extended mode.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -172,83 +177,327 @@ public void testQueryWithParameters() throws SQLException {
...
+    }
+
+    // Verify that the DML statements were batched together by PgAdapter.
+    List<ExecuteBatchDmlRequest> requests =
+        mockSpanner.getRequestsOfType(ExecuteBatchDmlRequest.class);
+    assertEquals(1, requests.size());
+    ExecuteBatchDmlRequest request = requests.get(0);
+    assertEquals(2, request.getStatementsCount());
+    assertEquals(INSERT_STATEMENT.getSql(), request.getStatements(0).getSql());
+    assertEquals(UPDATE_STATEMENT.getSql(), request.getStatements(1).getSql());
```
olavloite, line 214: nit: add an assertion that we receive exactly one `CommitRequest`.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -172,83 +177,327 @@ public void testQueryWithParameters() throws SQLException {
...
+        try (ResultSet resultSet = statement.getResultSet()) {
+          assertTrue(resultSet.next());
+          assertEquals(1L, resultSet.getLong(1));
+          assertFalse(resultSet.next());
+        }
+
+        assertFalse(statement.getMoreResults());
+        assertEquals(-1, statement.getUpdateCount());
+
+        // Start Batch
```
olavloite: nit:
```suggestion
        // Execute batch
```

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -172,83 +177,327 @@ public void testQueryWithParameters() throws SQLException {
...
+        }
+
+        assertFalse(statement.getMoreResults());
+        assertEquals(-1, statement.getUpdateCount());
+      }
+    }
+
+    List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+    assertEquals(3, requests.size());
+    assertEquals(SELECT1.getSql(), requests.get(0).getSql());
```
olavloite, line 260: nit: add an assertion that this `SELECT1` statement carries a `BeginTransaction` option (i.e. request.getTransaction().hasBegin()`)

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -172,83 +177,327 @@ public void testQueryWithParameters() throws SQLException {
...
+    List<ExecuteBatchDmlRequest> batchDmlRequests =
+        mockSpanner.getRequestsOfType(ExecuteBatchDmlRequest.class);
+    assertEquals(1, requests.size());
+    ExecuteBatchDmlRequest request = batchDmlRequests.get(0);
+    assertEquals(2, request.getStatementsCount());
+    assertEquals(UPDATE_STATEMENT.getSql(), request.getStatements(0).getSql());
+    assertEquals(INVALID_DML.getSql(), request.getStatements(1).getSql());
+
+    List<RollbackRequest> rollbackRequests = mockSpanner.getRequestsOfType(RollbackRequest.class);
+    assertEquals(0, rollbackRequests.size());
```
olavloite: There should be 1 rollback request. The `COMMIT` statement in the batch should not be ignored by PGAdapter, but should be changed into a `ROLLBACK`.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -172,83 +177,327 @@ public void testQueryWithParameters() throws SQLException {
...
```
olavloite: We should rollback the transaction when it enters the aborted state. Keeping the transaction alive on Cloud Spanner will unnecessarily hold on to locks, and we won't be using the transaction anymore (PGAdapter itself keeps track of the fact that the transaction is aborted)

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -172,83 +177,327 @@ public void testQueryWithParameters() throws SQLException {
...
+
+        // getMoreResults() should now return false. We should also check getUpdateCount() as that
+        // method should return -1 to indicate that there is also no update count available.
+        assertFalse(statement.getMoreResults());
+        assertEquals(-1, statement.getUpdateCount());
+      }
+    }
 
+    // The server will also receive a 'SELECT 1' statement from PgAdapter, as PgAdapter calls
+    // connection.isAlive() before using it.
```
olavloite: This is no longer the case
```suggestion
```

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -172,83 +177,327 @@ public void testQueryWithParameters() throws SQLException {
...
+    List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+    assertEquals(2, requests.size());
+    assertEquals(SELECT1.getSql(), requests.get(0).getSql());
+    assertTrue(requests.get(0).getTransaction().hasBegin());
```
olavloite: This turns out to be incorrect (my fault, I had written that wrong in the design doc): A batch that is executed without an active transaction should execute each statement in auto-commit mode. So this select statement (and also the next one) should be executed using a single-use read-only transaction (or more simply: The `com.google.cloud.spanner.Connection` should be in auto-commit mode, and will handle that automatically for you)

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -259,6 +508,366 @@ public void testDdl() throws SQLException {
...
+
+    // Verify that the DML statements were batched together by PgAdapter.
+    List<ExecuteBatchDmlRequest> requests =
+        mockSpanner.getRequestsOfType(ExecuteBatchDmlRequest.class);
+    assertEquals(1, requests.size());
+    ExecuteBatchDmlRequest request = requests.get(0);
+    assertEquals(2, request.getStatementsCount());
+    assertEquals(INSERT_STATEMENT.getSql(), request.getStatements(0).getSql());
+    assertEquals(UPDATE_STATEMENT.getSql(), request.getStatements(1).getSql());
+  }
```
olavloite, line 512: nit: Add an assertion that we also received one `CommitRequest`.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -259,6 +508,366 @@ public void testDdl() throws SQLException {
...
+            exception.getMessage(), containsString("INVALID_ARGUMENT: Statement is invalid."));
+      }
+    }
+
+    // Verify that the execution does not halt and the last statement is executed.
+    List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+    assertEquals(2, requests.size());
+    assertEquals(INSERT_STATEMENT.getSql(), requests.get(0).getSql());
+    assertEquals(UPDATE_STATEMENT.getSql(), requests.get(1).getSql());
+  }
```
olavloite, line 534: nit: add an assertion that we received two `CommitRequest`s.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -259,6 +508,366 @@ public void testDdl() throws SQLException {
...
+
+        assertFalse(statement.getMoreResults());
+        assertEquals(-1, statement.getUpdateCount());
+
+        statement.execute("COMMIT");
+      }
+    }
+
+    // The server will also receive a 'SELECT 1' statement from PgAdapter, as PgAdapter calls
+    // connection.isAlive() before using it. This should happen before the transaction is started.
```
olavloite: Remove, this is no longer true
```suggestion
```

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -259,6 +508,366 @@ public void testDdl() throws SQLException {
...
+    // Verify that the BEGIN statement is no-op
+    assertFalse(batchDmlRequest.getTransaction().hasBegin());
+    assertTrue(batchDmlRequest.getTransaction().hasId());
+    assertEquals(INSERT_STATEMENT.getSql(), batchDmlRequest.getStatements(0).getSql());
+    assertEquals(UPDATE_STATEMENT.getSql(), batchDmlRequest.getStatements(1).getSql());
+
+    List<CommitRequest> commitRequests = mockSpanner.getRequestsOfType(CommitRequest.class);
+    assertEquals(1, commitRequests.size());
+  }
+
```
olavloite, line 730: (I've reviewed everything up until here. I'll try to continue later today with the rest of these tests)

olavloite (commented):
Thanks, this looks great! I've temporarily stopped reviewing at line 730 of JdbcSimpleModeMockServerTest.java, but I'll try to continue reviewing later today to give you a full review of everything.

Sorry for the confusion regarding error handling of implicit transactions. It seems that I got confused by the current implementation of batching when writing the first part of the design doc. The rule for what to do in case of an error in a batch is actually a lot simpler than I thought:
1. Always continue executing statements in the batch.
2. In case of implicit transaction: Execute each statement in the batch using auto-commit. This means that each statement that succeeds should be committed, regardless of whether other statements fail.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -259,6 +508,366 @@ public void testDdl() throws SQLException {
...
+        // getMoreResults() should now return false. We should also check getUpdateCount() as that
+        // method should return -1 to indicate that there is also no update count available.
+        assertFalse(statement.getMoreResults());
+        assertEquals(-1, statement.getUpdateCount());
+      }
+
+      List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+      assertEquals(3, requests.size());
+      assertEquals(SELECT1.getSql(), requests.get(0).getSql());
```
olavloite, line 753: nit: verify that:
1. This and the following SELECT both use a single-use read-only transaction
2. The INSERT statement uses a read/write transaction, and that the transaction is committed.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -100,21 +100,22 @@ public void testBasicSelectStatement() throws Exception {
...
 
     assertFalse(intermediateStatement.isExecuted());
-    assertEquals("SELECT", intermediateStatement.getCommand());
+    assertEquals("SELECT", intermediateStatement.getCommand(0));
 
     intermediateStatement.execute();
 
-    verify(connection, Mockito.times(1)).execute(Statement.of("SELECT * FROM users"));
-    assertTrue(intermediateStatement.containsResultSet());
+    Mockito.verify(connection, Mockito.times(1)).execute(Statement.of("SELECT * FROM users"));
```
olavloite: nit: use a static import for `Mockito` calls and remove the `Mockito.` prefix. Also remove the `times(1)` call. That is implied for `verify(..)`, so it is only necessary to include `times(x)` if x > 1.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -191,34 +195,38 @@ public void testBasicNoResultStatement() throws Exception {
...
     assertFalse(intermediateStatement.isExecuted());
-    assertEquals("CREATE", intermediateStatement.getCommand());
+    assertEquals("CREATE", intermediateStatement.getCommand(0));
 
     intermediateStatement.execute();
 
-    verify(connection).execute(Statement.of("CREATE TABLE users (name varchar(100) primary key)"));
-    assertFalse(intermediateStatement.containsResultSet());
-    assertNull(intermediateStatement.getUpdateCount());
+    Mockito.verify(connection, Mockito.times(1))
```
olavloite: nit: Here also, remove `Mockito.` and use a static import, and remove `times(1)`

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
```diff
@@ -191,34 +195,38 @@ public void testBasicNoResultStatement() throws Exception {
...
+    assertEquals(StatementType.DDL, intermediateStatement.getStatementType(0));
+    assertNull(intermediateStatement.getStatementResult(0));
+    assertFalse(intermediateStatement.isHasMoreData(0));
+    assertFalse(intermediateStatement.hasException(0));
     assertEquals(0, intermediateStatement.getResultFormatCode(0));
 
     intermediateStatement.close();
 
-    verify(resultSet, never()).close();
+    Mockito.verify(resultSet, Mockito.times(0)).close();
```
olavloite:
```suggestion
    verify(resultSet, never()).close();
```

.ci/e2e-expected/ddl-dml-batching.txt
libingye816: According to psql documentation, psql only prints the result of the last SQL command in the string. 
"Each SQL command string passed to -c is sent to the server as a single request. Because of this, the server executes it as a single transaction even if the string contains multiple SQL commands, unless there are explicit BEGIN/COMMIT commands included in the string to divide it into multiple transactions. (See [Section 53.2.2.1](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-MULTI-STATEMENT) for more details about how the server handles multi-query strings.) Also, psql only prints the result of the last SQL command in the string. This is different from the behavior when the same string is read from a file or fed to psql's standard input, because then psql sends each SQL command separately." [(link)](https://www.postgresql.org/docs/current/app-psql.html)

I would also like to have a further discussion on this because I notice that in our design doc, the test cases have outputs for each statement in the batch and I would like to know how you get them.

olavloite: > According to psql documentation, psql only prints the result of the last SQL command in the string. "Each SQL command string passed to -c is sent to the server as a single request. Because of this, the server executes it as a single transaction even if the string contains multiple SQL commands, unless there are explicit BEGIN/COMMIT commands included in the string to divide it into multiple transactions. (See [Section 53.2.2.1](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-MULTI-STATEMENT) for more details about how the server handles multi-query strings.) Also, psql only prints the result of the last SQL command in the string. This is different from the behavior when the same string is read from a file or fed to psql's standard input, because then psql sends each SQL command separately." [(link)](https://www.postgresql.org/docs/current/app-psql.html)
> 
> I would also like to have a further discussion on this because I notice that in our design doc, the test cases have outputs for each statement in the batch and I would like to know how you get them.

The trick to get `psql` to print out the response for each command is to send the commands to the standard input of `psql`. For this you have to options:
1. Just open `psql` in interactive mode and enter the commands directly into `psql`:

```shell
loite@penguin:~$ psql -h localhost -d knut-test-db
Password for user loite: 
psql (14.1 (Debian 14.1-1.pgdg110+1))
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bits: 256, compression: off)
Type "help" for help.

knut-test-db=# begin; insert into t values (3); insert into t values (4); commit;
BEGIN
INSERT 0 1
INSERT 0 1
COMMIT
knut-test-db=# 
```

2. Use echo to input a batch of commands into the standard input of `psql`:

```shell
loite@penguin:~$ echo "begin; insert into t values (1); insert into t values (2); commit;" | psql -h localhost -d knut-test-db
Password for user loite: 
BEGIN
INSERT 0 1
INSERT 0 1
COMMIT
loite@penguin:~$ 
```

If you repeat the above command to force errors:

```shell
loite@penguin:~$ echo "begin; insert into t values (1); insert into t values (2); commit;" | psql -h localhost -d knut-test-db
BEGIN
ERROR:  duplicate key value violates unique constraint "t_pkey"
DETAIL:  Key (v)=(1) already exists.
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ROLLBACK
loite@penguin:~$ 
```

Note: You can make `psql` stop asking for a password by setting the environment variable PGPASSWORD to the password it should use before running `psql`.

olavloite: Note: If our current `psql` test setup is created in such a way that it only outputs the last command tag (that is; the commands are fed into `psql` using the `-c` flag), then we should leave that as it is for now, and change it in a separate PR.

libingye816: Okay, will leave it as it is.

.ci/e2e-expected/dml-batching.txt
libingye816: This is expected as the test was set up using -c.

.ci/e2e-expected/select-ddl-batching.txt
libingye816: I don't see "SELECT \<rowcount\>" using psql on my local postgres db.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
libingye816: Done

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done, but doesn't PARSER.parse do the same thing? 
```
ParsedStatement parse(Statement statement, QueryOptions defaultQueryOptions) {
    String sql = removeCommentsAndTrim(statement.getSql());
    ClientSideStatementImpl client = parseClientSideStatement(sql);
    if (client != null) {
      return ParsedStatement.clientSideStatement(client, statement, sql);
    } else if (isQuery(sql)) {
      return ParsedStatement.query(statement, sql, defaultQueryOptions);
    } else if (isUpdateStatement(sql)) {
      return ParsedStatement.update(statement, sql);
    } else if (isDdlStatement(sql)) {
      return ParsedStatement.ddl(statement, sql);
    }
    return ParsedStatement.unknown(statement, sql);
  }
```

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: This will not be exposed to customers through psql or pg drivers. For internal implementation, it is fine because we can know ahead that this is a query/ddl/client_side through getStatementTypes() and we never call this method in those cases.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Removed.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: As an alternative, I can return -1, which complies to what users will observe through jdbc's getUpdateCount()

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: ACK. Will construct a follow-up PR to address it.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Yes, it is returned to the client. See sendSpannerResult method defined in ControlMessage

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: The name might be misleading. It simply degrades to the handleExecutionException() if no transaction.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/DescribeMessage.java
libingye816: I don't know a way of determining the result type of a statement directly. Do we need to massage all the special cases ourselves? If so, can you list out all those cases?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
libingye816: Done. By the way, how do you run IT tests? I tried `mvn -DskipTests=false integration-test` but none was run.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: This may be removed again in the future once we change implicit transaction to autocommit.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: Yes, sorry, I should have been a little more elaborate (and also; we will probably try to optimize the `PARSER.parse` method to do that a little more intelligently). The reason that we should call `PARSER.parse(String)` is that it will return a `ParsedStatement`, and that is what we should keep in our list of statements. The reason for that is that:
1. `ParsedStatement` contains helper methods like `isQuery()` and `isUpdate()` that do not need to re-parse the statement. And as we potentially need to know what type of statement it is multiple times, it is better to only parse it once.
2. `ParsedStatement.isQuery()` will also handle client side statements correctly, that is; it knows whether a client side statement is a query or statement without a result.

So we should change the type of `this.statements` from `ImmutableList<String>` to `ImmutableList<ParsedStatement>`.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: If this is guaranteed not to be returned to clients when it is not the result, then I'm fine with this being 0 (zero). I think it would be good to add it to the documentation of the method that the result is only valid if the statement type at the given index is `UPDATE`.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: Ok, thanks.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
olavloite: Ok, thanks. Would you mind changing the name?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/DescribeMessage.java
olavloite: See also my answer further up: We should change the list of statements from `ImmutableList<String>` to `ImmutableList<ParsedStatement>`. `ParsedStatement` contains a method `isQuery()` that will handle that correctly for both 'normal' statements and client side statements.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
olavloite: You need to use `-DskipITs=false` instead of `-DskipTests=false`.

`mvn -DskipITs=false integration-test` should work.
`mvn -DskipITs=false verify` should also work.

I normally run them directly from IntelliJ, as it gives me more control over which tests are run. Running all integration tests can take a long time.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Why two commit requests? Two is the correct answer but I can only think of one before INVALID_DDL.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Under current implementation, all three statements are included in the same implicit transaction?

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/StatementTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
olavloite: An invalid DDL statement should not cause the batch to be halted. So we get one commit request from the insert statement, then an invalid DDL statement, and then the update statement, which is also committed.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
olavloite: > Under current implementation, all three statements are included in the same implicit transaction?

Ok. I'm open to changing that in a following PR. In the end we need to make this just use `autocommit=true` on the underlying Spanner connection. That will automatically pick the best transaction for each statement that we are executing.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/DescribeMessage.java
libingye816: Okay, done.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
libingye816: Thanks.

libingye816:
@olavloite Do you have an idea why clirr and e2e-psql-xxx are failing?

olavloite:
The e2e psql tests are failing because some of those tests check whether they get an expected error when executing a mixed batch. That is no longer the case, as this PR introduces support for that. The psql tests work with text files that contain the expected output. If you take a look at the original draft PR, you will see that I changed the expected output of some of those (see for example dml-ddl-batching.txt: https://github.com/GoogleCloudPlatform/pgadapter/pull/30/files)

Clirr is failing because this PR introduces breaking changes in the public API of PGAdapter. I think we should disable clirr for this repository. It does not really make very much sense, as this is not a library that is imported as a dependency into other applications.

EDIT: I've removed the clirr check as a required check for this repo. So it is still there and will still fail for this PR, but it won't block merging this PR.

libingye816:
> The e2e psql tests are failing because some of those tests check whether they get an expected error when executing a mixed batch. That is no longer the case, as this PR introduces support for that. The psql tests work with text files that contain the expected output. If you take a look at the original draft PR, you will see that I changed the expected output of some of those (see for example dml-ddl-batching.txt: https://github.com/GoogleCloudPlatform/pgadapter/pull/30/files)

I have already updated those expected output files. All tests were passed at last Friday, so I am surprised that they fail again. The psql receives error while expecting some outputs. Looks like the batch support doesn't take effect.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done. The change is now available in this PR.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
libingye816: Done.

libingye816:
I fixed the logic of statement split to ignore escaped quotes. It should work now. @olavloite

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -42,45 +48,40 @@ public class IntermediateStatement {
...
-  protected boolean executed;
-  protected final Connection connection;
-  protected Long updateCount;
-  protected final ImmutableList<String> statements;
+  protected List<String> commands;
+  protected boolean executed = false;
+  protected Connection connection;
+  protected long[] updateCounts;
+  protected List<ParsedStatement> statements;
+  private ConnectionHandler connectionHandler;
```
olavloite: nit: I think some of these could be declared final.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -147,9 +167,14 @@ private static ImmutableList<String> splitStatements(String sql) {
     return builder.build();
   }
 
-  protected static ImmutableList<String> parseStatements(String sql) {
+  protected static List<ParsedStatement> parseStatements(ParsedStatement stmt) {
```
olavloite: nit: Prefer `ImmutableList` over something like `ArrayList` whenever possible (I don't think the contents of this list can/should change after the initial creation).

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +297,190 @@ public Exception getException() {
...
+      if (executionStatus == ExecutionStatus.AUTOCOMMIT) {
+        // Executing a COMMIT/ROLLBACK statement in autocommit mode is a no-op
+        return;
       }
+      executionStatus = ExecutionStatus.AUTOCOMMIT;
+      if (connectionHandler.getStatus() == ConnectionStatus.TRANSACTION_ABORTED) {
+        connectionHandler.setStatus(ConnectionStatus.IDLE);
+        // COMMIT rollbacks aborted transaction
+        statements.set(index, PARSER.parse(Statement.of("ROLLBACK")));
+        commands.set(index, "ROLLBACK");
```
olavloite: Ah, this explains why `statements` is not an `ImmutableList`. I think it would be cleaner if we keep both `statements` and `commands` immutable, and instead create two variables:
1. `List<String> commandTag` that contains the command tags that we will be returning (instead of modifying the `command` list). This variable should be defined on class level (so next to `commands`).
2. `Statement statement` that is initially initialized to `statements.get(index)`. Then the value is changed here if we need to switch from `COMMIT` to `ROLLBACK`. So this variable is only defined for this code block.

(I'm perfectly fine with this change being done in a follow-up PR, by the way. This PR is already a large change.)

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +297,190 @@ public Exception getException() {
...
+      // Skip remaining statements in DML/DDL batch
+      // TODO: May want to start off from the failure in order to be compatible with the
+      // behavior of PG.
+      for (int i = fromIndex + counts.length + 1; i < index; i++) {
+        handleExecutionException(
+            i,
+            SpannerExceptionFactory.newSpannerException(
+                ErrorCode.INTERNAL,
+                "ERROR: Statement is not executed due to the preceding error in DML/DDL batch."));
       }
```
olavloite: nice solution for this for now 👍

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/ControlMessage.java
```diff
@@ -150,34 +150,37 @@ protected void handleError(Exception e) throws Exception {
...
+      int resultIndex, IntermediateStatement statement, QueryMode mode, long maxRows)
       throws Exception {
-    String command = statement.getCommand();
-    switch (statement.getResultType()) {
-      case NO_RESULT:
+    String command = statement.getCommand(resultIndex);
+    switch (statement.getStatementType(resultIndex)) {
+      case DDL:
+      case CLIENT_SIDE:
         new CommandCompleteResponse(this.outputStream, command).send(false);
```
olavloite, line 160: We need to do a check here whether the client side statement is a query or a no-result statement. You can do this by calling `statements.get(index).isQuery()`. (Also fine for a follow-up PR, but in that case add a TODO here)

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
```diff
@@ -116,21 +119,32 @@ public void handleQuery() throws Exception {
...
+      }
+    }
+    boolean inTransaction = connection.getSpannerConnection().isInTransaction();
+    Status transactionStatus = Status.IDLE;
+    if (inTransaction) {
+      if (connection.getStatus() == ConnectionStatus.TRANSACTION_ABORTED) {
+        transactionStatus = Status.FAILED;
+        // Actively rollback the aborted transaction
+        connection.getSpannerConnection().rollback();
+        connection.setStatus(ConnectionStatus.IDLE);
```
olavloite: I'm not sure this is correct. The connection status should remain `TRANSACTION_ABORTED` until we receive a `COMMIT`/`ROLLBACK` from the client. And instead of returning `ConnectionStatus.IDLE` 'I' we should return 'E' (error). See https://www.postgresql.org/docs/14/protocol-message-formats.html and search for `Current backend transaction status indicator`

I'm also fine with this being fixed in a follow-up PR, as this is not a regression from our current behavior. But please add a TODO for it so we don't forget.

src/test/java/com/google/cloud/spanner/pgadapter/EmulatedPsqlMockServerTest.java
```diff
@@ -0,0 +1,80 @@
...
+  public static void loadPgJdbcDriver() throws Exception {
+    // Make sure the PG JDBC driver is loaded.
+    Class.forName("org.postgresql.Driver");
+  }
+
+  @BeforeClass
+  public static void startMockSpannerAndPgAdapterServers() throws Exception {
+    // This test case fails because batching does not work properly when the server is started in
+    // psql mode (i.e. with -q).
+    // Comment out the next line and enable the second and the test case will succeed.
```
olavloite: Remove; you fixed this ;-)

src/test/java/com/google/cloud/spanner/pgadapter/EmulatedPsqlMockServerTest.java
```diff
@@ -0,0 +1,80 @@
...
+    doStartMockSpannerAndPgAdapterServers(ImmutableList.of("-q"));
+    // doStartMockSpannerAndPgAdapterServers(Collections.emptyList());
```
olavloite: nit: remove

olavloite (approved):
Thanks! This looks good to me. I have some TODOs / nits in my comments, but feel free to defer those to a follow-up PR.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: I did this on purpose. It may change from COMMIT to ROLLBACK for aborted transaction.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/ControlMessage.java
libingye816: I believe this is done in determineStatementType()? We call isQuery on a parsedStatement there.

libingye816: The CLIENT_SIDE here should be the true CLIENT_SIDE which has no result. Otherwise, it is categorized as QUERY.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/QueryMessage.java
libingye816: I see, so we still need to block clients until they rollback the transaction but we just silently rollback it in advance. Then, we only need to remove `connection.setStatus(ConnectionStatus.IDLE);`, 'E' is already returned by setting status to Status.FAILED.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/EmulatedPsqlMockServerTest.java
libingye816: Done.

src/test/java/com/google/cloud/spanner/pgadapter/EmulatedPsqlMockServerTest.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Done.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: thx :-)

libingye816:
Hi @olavloite Can you give a last round of quick review?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +299,191 @@ public Exception getException() {
...
+    executed = true;
+    if (connection.isInTransaction()) {
+      executionStatus = ExecutionStatus.EXPLICIT_TRANSACTION;
+    } else {
+      executionStatus = ExecutionStatus.AUTOCOMMIT;
+    }
+    int index = 0;
+    while (index < getStatementCount()) {
+      if (connectionHandler.getStatus() == ConnectionStatus.TRANSACTION_ABORTED
+          && !"ROLLBACK".equals(getCommand(index))
```
olavloite, line 353: The checks for rollback/commit must be case-insensitive (the best way is probably to fill the command list with all upper-case commands, as PostgreSQL will always return the command tags in upper case).

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
```diff
@@ -240,85 +299,191 @@ public Exception getException() {
...
+        executionStatus = ExecutionStatus.EXPLICIT_TRANSACTION;
       }
+    }
+    if ("COMMIT".equals(command) || "ROLLBACK".equals(command)) {
+      if (executionStatus == ExecutionStatus.AUTOCOMMIT) {
+        // Executing a COMMIT/ROLLBACK statement in autocommit mode is a no-op
+        return;
+      }
+      executionStatus = ExecutionStatus.AUTOCOMMIT;
+      if (connectionHandler.getStatus() == ConnectionStatus.TRANSACTION_ABORTED) {
```
olavloite, line 401: I'm not exactly sure how/why, but I'm not able to get out of an aborted state after a transaction has failed. Executing for example a batch like this (assuming invalid_table does not exist):

`BEGIN; INSERT INTO invalid_table VALUES (1); COMMIT;`

This will bring the connection into an aborted state, and it is impossible to get out of that state. The result of this batch should be that the transaction is rolled back by the commit, and that the connection is idle. Executing a separate rollback after the batch also does not get the connection out of the aborted state. Could you add a couple of tests for this so we know that getting out of the aborted state works.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: The commands are forced to be upper-case when they are parsed.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediateStatement.java
libingye816: Ok, so the reason is that the psql interactive mode splits the statements and sends them one by one rather than as a batch. The current logic assumes that the aborted transaction cannot happen when the execution status is autocommit, which is indeed incorrect. I reordered those lines and it works out.

PR #117, opened 2022-04-19
chore: mark public classes with @InternalApi

olavloite:
- Marks all public classes that are only intended for internal use with
  @InternalApi
- Makes public classes package private where possible


src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -137,17 +137,6 @@ public void stopServer() {
     awaitTerminated();
   }
 
-  public void run() {
```
pratickchokhani, line 140: Not sure why this change is in this PR.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/DateParser.java
```diff
@@ -74,7 +73,7 @@ public DateParser(byte[] item, FormatCode formatCode) {
    * @param value The value to check. May not be <code>null</code>.
    * @return <code>true</code> if the text contains a valid date.
    */
-  public static boolean isDate(String value) {
+  static boolean isDate(String value) {
```
pratickchokhani, line 76: Are there any concerns in exposing these methods publicly?

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
olavloite: Oh, this was more of a general cleanup PR, and I noticed that this method is not used anymore.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/DateParser.java
olavloite: Customers could add PGAdapter as a dependency to their own application and reference it like a library. That means that they could call any public method in any of these classes. It is therefore good practice to keep as many of them package private as possible, unless they are explicitly intended for public usage. As PGAdapter is set up in such a way that many classes are in separate packages, we still need a lot of public classes and methods. I've marked those as `@InternalApi` to clearly indicate that using them is not advisable, and that we may change the signature of those methods without declaring it as a breaking change.

pratickchokhani (approved):

PR #125, opened 2022-04-25
feat: add support for describe statement

olavloite:
Adds support for DescribeStatement by transforming sql statements into a select statement that queries the (types of) the parameters in the original query. This can be used to temporarily support prepared statements until a DescribeStatement endpoint is available in the backend.

With this change the entire extended query protocol is supported by PGAdapter.


src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -159,34 +176,283 @@ public IntermediatePortalStatement bind(
...
+        return null;
+      }
+    }
+
+    parser.skipWhitespaces();
+    int potentialSelectStart = parser.getPos();
+    if (parser.eat("select")) {
+      // This is an `insert into <table> [(...)] select ...` statement. Then we can just use the
+      // select statement as the result.
+      return Statement.of(
```
pratickchokhani: Can we make the Select statement method reusable?

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -159,34 +176,283 @@ public IntermediatePortalStatement bind(
...
+        return transformDmlToSelectParams(parameters);
+      case CLIENT_SIDE:
+      case DDL:
+      case UNKNOWN:
+      default:
+        return Statement.of(this.parsedStatement.getSqlWithoutComments());
+    }
+  }
+
+  private Statement transformSelectToSelectParams(Set<String> parameters) {
```
pratickchokhani: Can we add examples as comment? Same for other parsers

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
```diff
@@ -159,34 +176,283 @@ public IntermediatePortalStatement bind(
...
+      return new DescribeStatementMetadata(this.parameterDataTypes, null);
+    } finally {
+      if (columnsResultSet != null) {
+        columnsResultSet.close();
+      }
+    }
+  }
+
+  @VisibleForTesting
+  Statement transformToSelectParams(Set<String> parameters) {
```
pratickchokhani, line 257: Can this be made more simple by just fetching columns name and the respective table name? Because from what I understand the resultant query can be made from just those?
If yes, we can use regex to fetch the respective values. Which will be more readable and be less error prone.

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
olavloite: Done

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
olavloite: Done

src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePreparedStatement.java
olavloite: No, that is not possible, as the type of a parameter might be different from the type of the column that it is used to for example filter. Consider for example the following statement:

```sql
select *
from my_table
where created_at > to_timestamp($1, 'YYYY-MM-DD HH:MI')
```

The type of the column `created_at` is `TIMESTAMP`, but the type of the parameter is `text`.

pratickchokhani (approved):

PR #149, opened 2022-05-05
fix: error handling for extended query protocol

olavloite:
Fixes the error handling for the extended query protocol, so it correctly returns the `Transaction aborted` state after an error in a transaction.


src/main/java/com/google/cloud/spanner/pgadapter/statements/IntermediatePortalStatement.java
```diff
@@ -94,9 +93,9 @@ public DescribeMetadata describe() {
       ResultSet statementResult = connection.executeQuery(this.statement);
       setStatementResult(0, statementResult);
       return new DescribePortalMetadata(statementResult);
-    } catch (SpannerException e) {
-      logger.log(Level.SEVERE, e, e::getMessage);
```
olavloite, line 98: Removed logging, as an invalid user query is not something that we should be logging (it's not a system fault).

PR #166, opened 2022-06-01
test: enable domain sockets for pgx tests

olavloite:
Adds tests using Unix domain sockets for pgx.


olavloite:
Integration test failure due to upload failure to codecov.

PR #167, opened 2022-06-01
test: Add tests for Python psycopg2 driver

gauravsnj:
Added tests for Python psycopg2 driver


src/test/python/pgAdapterBasicTests.py
```diff
@@ -0,0 +1,23 @@
+import psycopg2 as pg
+import sys
+
+try:
+  connection = pg.connect(user = "postgres",
+                          database = "postgres",
+                          host = "localhost",
+                          port = "5432")
```
olavloite: I think the problem that you are having is caused by this line. The `AbstractMockServerTest` will not start PGAdapter on port 5432, but on a dynamically assigned port. This is done to prevent problems on systems that already have PostgreSQL or anything else running on port 5432, as well as to make it possible to run these tests in parallel (which would mean that there are multiple instances of PGAdapter running at the same time).

If you take a look here:
https://github.com/GoogleCloudPlatform/pgadapter/blob/9be1c3c3955e041c9fb075a5ce6578cba6506202/src/test/java/com/google/cloud/spanner/pgadapter/csharp/AbstractNpgsqlMockServerTest.java#L322

you can see that the .NET tests create the connection string (including the port number) in Java code, and then passes that connection string to the .NET program, so it knows which port number to use.

olavloite (approved):

PR #174, opened 2022-06-10
feat: enable authentication

olavloite:
Enables authentication on PGAdapter so the client can supply the credentials for a specific connection, instead of setting a fixed set of credentials on PGAdapter at startup. This also enables connecting to different Google Cloud projects and instances. The database property in the connection parameters may be a simple database id (e.g. `my-database`) or a fully qualified database name (e.g. `projects/my-project/instances/my-instance/databases/my-database`).

The credentials must be one of the following:
1. The private key string of a service account. This string **must** include the `-----BEGIN PRIVATE KEY-----` and `-----END PRIVATE KEY-----` headers. The username must be the email address of the service account.
2. The JSON contents of a Google credentials file. This could be a service account file or a user account file. The credentials file is sent in the password message. The username is ignored.


aseering (approved):
@guangylegend I wonder if this change to PGAdapter would be sufficient to enable running multiple Spangres tests against multiple databases (in the same local instance) using a single PGAdapter, like we just talked about earlier today?

Anyway -- I am by no means a qualified Java reviewer, but overall this change seems reasonable.

olavloite:
> @guangylegend I wonder if this change to PGAdapter would be sufficient to enable running multiple Spangres tests against multiple databases (in the same local instance) using a single PGAdapter, like we just talked about earlier today?

You actually don't need this change for that, as that is already supported. If you start PGAdapter __without__ a `-d` (database) command line argument, PGAdapter will connect to the database that is given in the connection request. Note that this does mean that you **must** include a database in the connection request when you start for example `psql`.

This change will however also add support for connecting to different databases on different instances or even in different projects, as it will add support for fully-qualified database names in the connection request.

PR #176, opened 2022-06-12
build: skip Python tests in default build

olavloite:
Removes the Python tests from the default build to prevent build failures on systems without Python installed.The Python tests are still executed when the project is built in the `test-all` profile.


gauravsnj (approved):
LGTM

PR #184, opened 2022-06-14
chore: refactor ProxyServer to use channels instead of sockets

olavloite:
Refactors the ProxyServer to use the SocketChannel interfaces instead of Sockets. This makes the sockets replaceable for the new Unix domain socket implementation in Java16, which again means that we can conditionally load that when we know that we have Java16 or higher as the target language.

Also adds a Maven profile for building a .jar that can be used for a native image. This profile loads a Socket factory that uses the default Java Unix domain sockets implementation instead of a third-party dependency that requires a native library.


.github/workflows/native-image.yaml
```diff
@@ -0,0 +1,65 @@
...
+  workflow_dispatch:
+  pull_request:
+  schedule:
+    - cron: '0 2 * * 1,2,3,4,5'
+name: native-image
+env:
+  GOOGLE_CLOUD_PROJECT: "span-cloud-testing"
+  GOOGLE_CLOUD_INSTANCE: "pgadapter-testing"
+  GOOGLE_CLOUD_DATABASE: "testdb_integration"
+  GOOGLE_CLOUD_ENDPOINT: "spanner.googleapis.com"
```
pratickchokhani, line 12: Is it fine to to put these in public repo?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -80,7 +83,9 @@ public class ConnectionHandler extends Thread {
   private static final String CHANNEL_PROVIDER_PROPERTY = "CHANNEL_PROVIDER";
 
   private final ProxyServer server;
-  private final Socket socket;
+  private final ByteChannel socketChannel;
+  private final InetSocketAddress remoteAddress;
```
pratickchokhani: Should we add a comment that this is nullable?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -100,18 +105,26 @@ public class ConnectionHandler extends Thread {
   private WellKnownClient wellKnownClient;
   private ExtendedQueryProtocolHandler extendedQueryProtocolHandler;
 
-  ConnectionHandler(ProxyServer server, Socket socket) {
+  ConnectionHandler(ProxyServer server, ByteChannel byteChannel) throws IOException {
     super("ConnectionHandler-" + CONNECTION_HANDLER_ID_GENERATOR.incrementAndGet());
     this.server = server;
-    this.socket = socket;
+    this.socketChannel = byteChannel;
+    if (byteChannel instanceof SocketChannel
```
pratickchokhani, line 125: What are the cases when byteChannel is not of SocketChannel type?

.github/workflows/native-image.yaml
olavloite: The endpoint is public information. It's for example also included in the generated client for each programming language.
The project, instance and database names are also included in many other repositories.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Good point, added a `@Nullable` annotation and a comment.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: I've added a comment to explain. `SocketChannel` is only used by TCP connections. Unix domain sockets use other implementations.

olavloite:
There's a good chance that this will not be needed, as the Unix domain socket factory that we are currently using has added support for GraalVM, which was the initial reason to add this.

818Nawaf (approved):

PR #192, opened 2022-06-16
chore: remove guessType for unknown parameter types

olavloite:
Remove the guessType method as the backend is now able to deduce the
type of an unspecified parameter, and PGAdapter supports the
DescribeStatement message.


olavloite:
Closing as this still fails for timestamps

PR #231, opened 2022-06-23
test: Add tests related to named parameters with different data types for Python psycopg2 driver 

gauravsnj:


pom.xml
```diff
@@ -26,7 +26,7 @@
       com.google.cloud.spanner.pgadapter.IntegrationTest,
       com.google.cloud.spanner.pgadapter.golang.GolangTest,
       com.google.cloud.spanner.pgadapter.csharp.DotnetTest,
-      com.google.cloud.spanner.pgadapter.python.PythonTest,
+      <!--com.google.cloud.spanner.pgadapter.python.PythonTest,-->
```
olavloite: nit: remove commented code. The Python tests should be excluded for default builds, so the build does not break on systems that have only Java installed. Running `mvn test -P test-all` will run all tests, including those that are excluded here.

pom.xml
```diff
@@ -207,7 +207,8 @@
       <id>test-all</id>
       <properties>
         <excludedTests>
-          com.google.cloud.spanner.pgadapter.IntegrationTest,
+          com.google.cloud.spanner.pgadapter.IntegrationTest
+          <!--com.google.cloud.spanner.pgadapter.python.PythonTest-->
```
olavloite: Here also remove commented code

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonBasicTests.java
```diff
@@ -190,4 +190,243 @@ public void testBasicDeleteWithParameters() throws IOException, InterruptedExcep
...
+    String sql2 =
+        "SELECT * FROM some_table where COLUMN_NAME1 = 'VALUE1' and COLUMN_NAME2 = 'VALUE2'";
+    mockSpanner.putStatementResult(StatementResult.query(Statement.of(sql2), createResultSet()));
+
+    String actualOutput =
+        executeWithNamedParameters(pgServer.getLocalPort(), sql, "query", parameters);
+    String expectedOutput = "(1, 'abcd')\n";
+    assertEquals(expectedOutput, actualOutput);
+
+    assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
```
olavloite, line 222: Could we add an assertion that the `ExecuteSqlRequest` that the server received is equal to `sql2`? So something like `assertEquals(sql2, mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0).getSql());`

src/test/python/StatementsWithNamedParameters.py
```diff
@@ -0,0 +1,106 @@
...
+  i = 0
+  while i < len(parameters):
+    parameters_dic[parameters[i]] = eval(parameters[i+1]+'('+ parameters[i + 2]+')')
+    i += 3
+  return parameters_dic
+
+
+def parse(statement_type, parameters):
+  if 'data_type' in statement_type:
+    return statement_type[10:], parse_named_data_type_parameters(parameters)
```
olavloite, line 86: What's the `10` in `statement_type[10:]`? Is that because we never have queries with more than 10 types?

Would it be possible to make this more generic? (E.g. get the last element in the array, or last X elements of the array) Or is this generic enough, and will it work if we for example add tests for more data types in the future?

olavloite (approved):
Looks generally good to me, with some nits and suggestions.

It surprised me that psycopg2 does not really support real prepared statements (i.e. statements that use actual parameters). Instead, it just replaces the parameter strings with literals client side before sending it to PGAdapter.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonBasicTests.java
gauravsnj: Sure. I've added this line to all the tests

pom.xml
gauravsnj: Done

pom.xml
gauravsnj: Done

src/test/python/StatementsWithNamedParameters.py
gauravsnj: Actually, whenever I am sending all the parameters as a specific data_type I am giving the StatementType string (in the cmd line arguments) as "data_type_query" or "data_type_update". So, the first 10 characters are "data_type_". That's why I'm removing those 10 characters to get the actual statementType. It has nothing to do with types of queries.

Also, if we need to add more tests related to new data types, we can do so without actually modifying the Python code. We just need a function that can convert string to that particular data type.

gauravsnj:
Yes Psycopg2 doesn't provide any support for the prepared statement. But one can use Prepare statement with the execute function of the Cursor class like `cur.execute("PREPARE stmt AS big and complex SQL with $1 $2 params")` and then `cur.execute("stmt(%s, %s)", (param1, param2))`

Starting from version 3, Psycopg will automatically prepare statements based on if frequency of execution of a particular statement crosses `prepare_threshold`. You can check this [link](https://www.psycopg.org/psycopg3/docs/advanced/prepare.html)

Also, one can emulate prepared statements by overriding the methods. Check this [blog](https://www.psycopg.org/articles/2012/10/01/prepared-statements-psycopg/) and [example](https://gist.github.com/dvarrazzo/3797445)

PR #251, opened 2022-07-01
test: added some integration tests for the Python Psycopg2 driver

gauravsnj:


src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
```diff
@@ -20,16 +20,20 @@
...
+import org.junit.FixMethodOrder;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+import org.junit.runners.MethodSorters;
 
 @RunWith(JUnit4.class)
 @Category(IntegrationTest.class)
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
olavloite: I know that we use this some places in the Cloud Spanner JDBC driver, but generally speaking, we should try to avoid this in new code. This makes each test case potentially dependent on all the previous test cases, which can make it very difficult to debug when something goes wrong. It also makes it more difficult to add new test cases or modify existing ones.

It is therefore better to insert/update the data that is needed for a specific test as part of the setup of that specific test, instead of relying on data from a previous test.

Note: Re-using the same table for all tests in a test class is OK, as creating a table can take a little while on Cloud Spanner.

src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
```diff
@@ -90,4 +94,246 @@ public void testBasicInsert() throws IOException, InterruptedException {
     String expectedOutput = "1\n";
     assertEquals(expectedOutput, actualOutput);
   }
+
+  @Test
+  public void test2Select() throws IOException, InterruptedException {
+    String sql =
```
olavloite, line 131: For this test it would be better to insert the data that it needs as part of the setup for this test, instead of relying on the data from the previous test. You can use the standard features in Spanner to insert the data, that does not need to use Python. See for example [this setup method](https://github.com/GoogleCloudPlatform/pgadapter/blob/a006dec4454d875d2224a1f2268a83612ce0e06d/src/test/java/com/google/cloud/spanner/pgadapter/golang/ITPgxTest.java#L123).

src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
```diff
@@ -90,4 +94,246 @@ public void testBasicInsert() throws IOException, InterruptedException {
...
+    String sql = "DELETE FROM all_types where col_bigint = 2";
+
+    String actualOutput = executeWithoutParameters(testEnv.getPGAdapterPort(), sql, "update");
+    String expectedOutput = "1\n";
+    assertEquals(expectedOutput, actualOutput);
+
+    sql = "Select COUNT(*) from all_types where col_bigint = 2";
+
+    actualOutput = executeWithoutParameters(testEnv.getPGAdapterPort(), sql, "query");
+    expectedOutput = "(0,)\n";
```
olavloite, line 194: Out of curiosity: Why is the output `(0,)\n` here, and not `(0)\n`?

src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
```diff
@@ -90,4 +94,246 @@ public void testBasicInsert() throws IOException, InterruptedException {
...
+    assertEquals(expectedOutput, actualOutput);
+  }
+
+  @Test
+  public void test1InsertParametrized() throws IOException, InterruptedException {
```
olavloite: nit:
```suggestion
  public void test1InsertParameterized() throws IOException, InterruptedException {
```

src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
gauravsnj: Okay. I've changed the code in such a way that the order of tests will not matter now.

src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
gauravsnj: Each row in the result is returned as a tuple in Psycopg2. Since in this case there is only one column, the size of the tuple is 1.

In Python, tuples are the objects that are delimited by comma(,). So, even if the size of tuple is 1, we need to put a comma after the element. Parenthesis in tuple are optional.

For example, `t = (1, 2, 3)` will create a tuple and so will `t = 1, 2, 3`

For creating a tuple of size 1, `t = (1,)` and `t = 1,` will work but `t = (1)` will not work because tuples are identified by delimitation by comma and not by parenthesis, hence we need to put one comma after `1`

olavloite: Thanks for the elaborate explanation. That makes sense to me.

olavloite (approved):

src/test/java/com/google/cloud/spanner/pgadapter/python/ITPsycopg2Test.java
olavloite: Thanks, LGTM

PR #275, opened 2022-07-11
feat: support current_database and server_version

olavloite:
Adds support for local handling of select current_database and show server_version
statements. These statements are known to be used by at least TypeORM.


src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
```diff
@@ -108,15 +122,16 @@ public ImmutableList<LocalStatement> getLocalStatements(ConnectionHandler connec
...
-  public static WellKnownClient detectClient(
+  public static @Nonnull WellKnownClient detectClient(
       List<String> orderParameterKeys, Map<String, String> parameters) {
     for (WellKnownClient client : WellKnownClient.values()) {
       if (client.isClient(orderParameterKeys, parameters)) {
         return client;
       }
     }
-    return null;
+    throw new IllegalStateException("UNSPECIFIED.isClient() should have returned true");
```
gauravsnj, line 135: Is throwing this exception covered in the tests?

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/ShowServerVersionStatement.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  public StatementResult execute(BackendConnection backendConnection) {
+    ResultSet resultSet =
+        ResultSets.forRows(
+            Type.struct(StructField.of("server_version", Type.string())),
+            ImmutableList.of(
+                Struct.newBuilder()
+                    .set("server_version")
+                    .to(backendConnection.getOptionsMetadata().getServerVersion())
```
gauravsnj, line 47: Is it sure that getServerVersion() will not return null? If not, do we need to throw an error message instead of showing the version as null?

src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
gauravsnj, line 138: Is this line reachable through any condition?

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/SelectCurrentDatabaseStatement.java
```diff
@@ -0,0 +1,63 @@
...
+  public String[] getSql() {
+    return new String[] {
+      "select current_database()",
+      "SELECT current_database()",
+      "Select current_database()",
+      "select current_database",
+      "SELECT current_database",
+      "Select current_database",
+      "select * from current_database()",
+      "SELECT * FROM current_database()",
```
gauravsnj, line 45: Why do we need to write `"select * from current_database()"` and `"SELECT * from current_database()"` as two different entries here?

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/SelectCurrentDatabaseStatement.java
```diff
@@ -0,0 +1,63 @@
...
+      "select * from current_database",
+      "SELECT * FROM current_database"
+    };
```
gauravsnj, line 50: I think `select * from current_catalog` and `select current_catalog` also return database name. Any specific reason for excluding them?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
```diff
@@ -138,6 +138,39 @@ public void testSelectCurrentSchema() throws SQLException {
     assertEquals(0, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
   }
 
+  @Test
+  public void testSelectCurrentDatabase() throws SQLException {
+    String sql = "SELECT current_database";
```
gauravsnj: Please correct me if I'm wrong but I don't think it should work for `SELECT current_database` because `current_database` is a function and the statement should be like `select current_database()`

gauravsnj: Have you tested with the scenarios like `Select current_database()`, `SeLect cUrrent_DATabase`, etc? Should we add such tests here also?

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/ShowServerVersionStatement.java
gauravsnj, line 47: What does `getServerVersion()` exactly returns? I mean does it return the version of the Spanner or the details of OS on which the instance is running because in PostgreSQL, `show server_version` doesn't returns the version of the PostgreSQL but the details about the OS like `11.16 (Debian 11.16-0+deb10u1)
`

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
```diff
@@ -138,6 +138,39 @@ public void testSelectCurrentSchema() throws SQLException {
...
+      }
+    }
+
+    // The statement is handled locally and not sent to Cloud Spanner.
+    assertEquals(0, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+  }
+
+  @Test
+  public void testShowServerVersion() throws SQLException {
+    String sql = "show server_version";
```
gauravsnj, line 185: Have you tested with the scenarios like `Show server_version`, `ShoW servER_verSIoN`, etc? Should we add such tests here also?

gauravsnj (commented):
Looks mostly good to me. Just need info on some points.

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/ShowServerVersionStatement.java
olavloite: This value is guaranteed to be not null. It is set as one of the options when starting PGAdapter, and when no value has been set, it will default to the string "1.0.0".

src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
olavloite: No, it is not reachable. Added a comment to the code to clarify that.

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/SelectCurrentDatabaseStatement.java
olavloite: These statements are intended to support auto-generated queries from frameworks (e.g. ORMs). These could use different casings, although something like `sElEcT ...` is very unlikely to be used by any framework. So to keep it as efficient as possible, this has been created using a simple `Map` where the same statement is mapped to a number of different keys (one key for each casing that is reasonable). This has a much quicker lookup time than having to do a case-insensitive comparison of SQL strings. So you're right that this will not work for all possible ways to write the statement (it will for example also fail if the statement uses tabs instead of normal whitespaces to separate keywords), but that is an intended trade-off for efficiency in this case.

I've added a comment to the `LocalStatement` interface to clarify this.

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/SelectCurrentDatabaseStatement.java
olavloite: Good point, thanks. The simple reason was 'because it was not used by any framework that I've tested so far'. I've added mappings for `current_catalog` as well.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
olavloite: You're right, I was assuming that `current_database` had the same status as `current_schema`. `current_schema` is a special case, that is allowed to be used both with and without `()`. `current_database()` must always include the parentheses.

olavloite: No, they are intentionally not all supported (see above)

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
olavloite: No, they are intentionally not all supported for performance reasons (see above)

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/ShowServerVersionStatement.java
olavloite: It currently returns the version that has been set in the options. I think that exactly what it returns in real PostgreSQL will depend on the PostgreSQL build. On my local (MacOS) system, it simply returns `13.4`, which corresponds with the PostgreSQL version I have installed.

gauravsnj (approved):
Thanks for providing clarification to my doubts. LGTM

PR #290, opened 2022-07-19
test: added tests for the transactions in the Psycopg2 

gauravsnj:


src/test/java/com/google/cloud/spanner/pgadapter/AbstractMockServerTest.java
```diff
@@ -88,7 +88,7 @@
  * server received the requests that the test expected.
  */
 public abstract class AbstractMockServerTest {
-  private static final Logger logger = Logger.getLogger(AbstractMockServerTest.class.getName());
+  protected static final Logger logger = Logger.getLogger(AbstractMockServerTest.class.getName());
```
olavloite: nit: loggers should normally be kept private and specific to one class, and a new logger should be recreated in the subclass. This makes it easier to configure loggers per class.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    for (ExecuteSqlRequest request : requests) {
+      sql.add(request.getSql());
+    }
+
+    Assert.assertEquals(true, sql.contains(sql1));
+    Assert.assertEquals(true, sql.contains(sql2));
+    Assert.assertEquals(true, sql.contains(sql3));
+    Assert.assertEquals(true, sql.contains(sql4));
+    Assert.assertEquals(true, sql.contains(sql5));
+    Assert.assertEquals(true, sql.contains(sql6));
```
olavloite: nit: remove the for loop and replace the `assertEquals(true, ...)` comparisons with an actual comparison of the request at for example index 2 with `sql3`. This then also ensures that the requests were executed in the expected order.

Additional nits here and elsewhere: Prefer `assertTrue(..)` over `assertEquals(true, ...)` and use a static import for `assertEquals` to make the code less verbose.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+            .getId()
+            .toStringUtf8();
+    String transactionIdForSql3 =
+        requests.stream()
+            .filter(req -> req.getSql().equals(sql3))
+            .findAny()
+            .get()
+            .getTransaction()
+            .getId()
+            .toStringUtf8();
```
olavloite: nit: you don't need to convert this to a string. You can just compare the `ByteString` instances.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+
+    Assert.assertEquals(true, transactionIdForSql2.equals(transactionIdForSql3));
```
olavloite: nit: just compare the transaction ids directly (`ByteString` has an `equals(Object)` implementation)

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+        StatementResult.query(Statement.of(sql1), createResultSet(1, "abcd")));
+    mockSpanner.putStatementResult(StatementResult.update(Statement.of(sql2), 1));
+    mockSpanner.putStatementResult(
+        StatementResult.query(Statement.of(sql3), createResultSet(2, "pqrs")));
+    mockSpanner.putStatementResult(StatementResult.update(Statement.of(sql4), 2));
+
+    String expectedOutput = "(1, 'abcd')\n" + "1\n" + "2\n" + "(2, 'pqrs')\n" + "1\n" + "2\n";
+    String actualOutput = executeTransactions(pgServer.getLocalPort(), statements);
+    for (AbstractMessage request : mockSpanner.getRequests())
+      System.out.println(request.getClass());
```
olavloite: nit: I think this should be removed

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    List<String> sql = new ArrayList<>();
+
+    for (ExecuteSqlRequest request : mockSpanner.getRequestsOfType(ExecuteSqlRequest.class)) {
+      sql.add(request.getSql());
+    }
+
+    Assert.assertEquals(1, Collections.frequency(sql, sql1));
+    Assert.assertEquals(2, Collections.frequency(sql, sql2));
+    Assert.assertEquals(1, Collections.frequency(sql, sql3));
+    Assert.assertEquals(2, Collections.frequency(sql, sql4));
```
olavloite: This needs to verify the order of the execution of the different statements and commit requests. Now we don't really know which statements were auto-committed and which were part of a transaction.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    statements.add(sql4);
+
+    // 1 commit request will be sent
+    statements.add("transaction");
+    statements.add("commit");
+
+    statements.add("transaction");
+    statements.add("set autocommit True");
+
+    // no commit request after this query although autocommit is true because it is a query sql
```
olavloite, line 206: Could we also add an assertion below that not only is there no commit request after this statement, but that it also uses a single-use read-only transaction? (So something like `assertTrue(request.getTransaction().hasSingleUse())` and `assertTrue(request.getTransaction().getSingleUse().hasReadOnly())`)

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    List<String> sql = new ArrayList<>();
+
+    for (ExecuteSqlRequest request : mockSpanner.getRequestsOfType(ExecuteSqlRequest.class)) {
+      sql.add(request.getSql());
+    }
+
+    Assert.assertEquals(1, Collections.frequency(sql, sql1));
+    Assert.assertEquals(2, Collections.frequency(sql, sql2));
+    Assert.assertEquals(1, Collections.frequency(sql, sql3));
+    Assert.assertEquals(2, Collections.frequency(sql, sql4));
```
olavloite: Can we also here verify the order of the requests (and most importantly: which requests carry a single-use read-only transaction and which ones use a read/write transaction)

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    statements.add("transaction");
+    statements.add("set readonly True");
+
+    // query will be executed as expected
+    statements.add("query");
+    statements.add(sql3);
+
+    // update will throw an error because the readonly is activated
+    statements.add("update");
+    statements.add(sql2);
```
olavloite, line 458: It's good that we test this, but we should also test a successful read-only transaction that is committed by the application. Committing a read-only transaction in psycopg2 is allowed (and required to release all resources associated with the transaction). It should however not lead to a `CommitRequest` being sent to Cloud Spanner, as Cloud Spanner does not require read-only transactions to be committed or rolled back.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+
+    // 1 commit request will be sent
+    statements.add("transaction");
+    statements.add("commit");
+
+    statements.add("transaction");
+    statements.add("set session readonly True");
+
+    // query will be executed as expected
+    statements.add("query");
```
olavloite, line 550: Can we verify that this starts a multi-use read-only transaction. That is; the session (connection) is now in read-only mode, but not in autocommit mode. This means that the first statement after a commit should start a new read-only transaction, and all subsequent statements should use that read-only transaction. The transaction should last until the next commit/rollback. The following statement should start a new read-only transaction.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    statements.add(sql1);
+
+    // update will be executed as expected
+    statements.add("update");
+    statements.add(sql2);
+
+    // update will be executed as expected
+    statements.add("update");
+    statements.add(sql4);
```
olavloite: nit: Can we remove this from the test? I don't think it adds anything to this test

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    for (ExecuteSqlRequest request : mockSpanner.getRequestsOfType(ExecuteSqlRequest.class)) {
+      sql.add(request.getSql());
+    }
+
+    Assert.assertEquals(1, Collections.frequency(sql, sql1));
+    Assert.assertEquals(1, Collections.frequency(sql, sql2));
+    Assert.assertEquals(0, Collections.frequency(sql, sql3));
+    Assert.assertEquals(1, Collections.frequency(sql, sql4));
+
+    mockSpanner.clearRequests();
```
olavloite: Can we split this test into three separate test cases:
1. A test for the isolation levels that we don't support (so 2 through 4) and try each of them, possibly just in a loop.
2. A test for the isolation level(s) that we do support (currently only 1)
3. A separate test for setting it to level 0. I wonder exactly what SQL statement the driver is generating if you try to set autocommit by setting the isolation level to 0.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+
+    Assert.assertEquals(1, Collections.frequency(sql, sql1));
+    Assert.assertEquals(1, Collections.frequency(sql, sql2));
+    Assert.assertEquals(0, Collections.frequency(sql, sql3));
+    Assert.assertEquals(1, Collections.frequency(sql, sql4));
+
+    Assert.assertEquals(
+        2,
+        getWireMessagesOfType(QueryMessage.class).stream()
+            .filter(qm -> qm.toString().contains("READ COMMITTED"))
```
olavloite: The specific behavior for `READ COMMITTED` for versions before/after 9.1 should be in a separate test case, and needs a little documentation so later readers understand what is going on.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+      sql.add(request.getSql());
+    }
+
+    Assert.assertEquals(1, Collections.frequency(sql, sql1));
+    Assert.assertEquals(1, Collections.frequency(sql, sql2));
+    Assert.assertEquals(0, Collections.frequency(sql, sql3));
+    Assert.assertEquals(1, Collections.frequency(sql, sql4));
+
+    mockSpanner.clearRequests();
+    statements.set(9, "set session isolation_level 4");
```
olavloite: This should also be split into multiple test cases.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+
+    actualOutput = executeTransactions(pgServer.getLocalPort(), statements);
+    Assert.assertEquals(expectedOutput, actualOutput);
+
+    for (AbstractMessage request : mockSpanner.getRequests())
+      System.out.println(request.getClass());
```
olavloite: Replace this with assertions that we receive the requests in the expected order

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    statements.add(sql2);
+
+    statements.add("update");
+    statements.add(sql4);
+
+    statements.add("transaction");
+    statements.add("commit");
+
+    // This will not cause error because the server version is 1.0,
+    // so the psycopg2 will send SERIALIZABLE instead of REPEATABLE READ as the isolation level
```
olavloite, line 977: It would be good to add a more elaborate comment on this. For someone who is not familiar with PostgreSQL / psycopg2, this comment does not explain why and when there is a difference.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+      sql.add(request.getSql());
+    }
+
+    Assert.assertEquals(1, Collections.frequency(sql, sql1));
+    Assert.assertEquals(1, Collections.frequency(sql, sql2));
+    Assert.assertEquals(1, Collections.frequency(sql, sql3));
+    Assert.assertEquals(1, Collections.frequency(sql, sql4));
+
+    mockSpanner.clearRequests();
+    statements.set(9, "set isolation_level 3");
```
olavloite: Try to split these tests into separate test cases. Ideally, each test case should only test one scenario.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    statements.add("update");
+    statements.add(sql2);
+
+    // no commit request will be sent after this update because autocommit is false
+    statements.add("update");
+    statements.add(sql4);
+
+    // 1 commit request will be sent
+    statements.add("transaction");
+    statements.add("commit");
```
olavloite: I think we can remove this, as it is not related to what is being tested here.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    Assert.assertEquals(
+        1,
+        getWireMessagesOfType(QueryMessage.class).stream()
+            .filter(qm -> qm.toString().contains("REPEATABLE READ"))
+            .count());
+
+    restartServerWithDifferentVersion("1.0");
+  }
+
+  @Ignore("To be Removed when changes in Client Library are live")
```
olavloite: nit: Make this comment more specific. That is; add a reference to the specific PR that is required for this to be enabled.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+
+    Assert.assertEquals(0, mockSpanner.countRequestsOfType(CommitRequest.class));
+    Assert.assertEquals(0, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+
+    statements.set(
+        1, "set session isolation_level 2 readonly True autocommit False deferrable True");
+
+    mockSpanner.clearRequests();
+
+    // TODO
```
olavloite: nit (here and elsewhere in this test): Remove, or specify what the TODO is

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1095 @@
...
+    String actualOutput = executeTransactions(pgServer.getLocalPort(), statements);
+
+    Assert.assertEquals(expectedOutput, actualOutput);
+
+    Assert.assertEquals(1, mockSpanner.countRequestsOfType(CommitRequest.class));
+    Assert.assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+
+    statements.set(1, "set session isolation_level 3 readonly True autocommit False");
+
+    mockSpanner.clearRequests();
```
olavloite: Try to split the different scenarios that are being tested into separate test cases. That makes it easier to debug if a test fails.

src/test/python/StatementsWithTransactions.py
```diff
@@ -0,0 +1,107 @@
...
+def set_property(connection, property, value):
+  if property == 'autocommit':
+    connection.autocommit = bool(value)
+  elif property == 'deferrable':
+    connection.deferrable = bool(value)
+  elif property == 'isolation_level':
+    if value == '0':
+      connection.set_isolation_level(0)
+    else:
+      connection.isolation_level = int(value)
```
olavloite, line 39: nit: do we need this if-else?

src/test/python/StatementsWithTransactions.py
```diff
@@ -0,0 +1,107 @@
...
+  elif property == 'readonly':
+    connection.readonly = bool(value)
+  else:
+    raise Exception('Trying to set Invalid Property '+property)
+
+def set_session_property(connection, sql):
+  properties = sql.split()
+  i = 0
+  arguements_string = '('
```
olavloite:
```suggestion
  arguments_string = '('
```

src/test/java/com/google/cloud/spanner/pgadapter/AbstractMockServerTest.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Thank for suggesting this. Done the change

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Thank for suggesting this. Done the change

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Thanks for pointing this out. Removed it

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: I've added this test also in the same test case.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done. 
Actually, when setting it to 0, it will behave just like setting `autocommit = True`. So, no particular statement will be sent.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/python/StatementsWithTransactions.py
gauravsnj: Yeah, because we aren't allowed to do `connection.isolation_level = 0` or `set_session(isolation_level = 0)`. The only way to set it to 0 is by using `set_isolation_level()` function

src/test/python/StatementsWithTransactions.py
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1514 @@
...
+    String actualOutput = executeTransactions(pgServer.getLocalPort(), statements);
+    assertEquals(expectedOutput, actualOutput);
+
+    assertEquals(1, mockSpanner.countRequestsOfType(CommitRequest.class));
+    assertEquals(1, mockSpanner.countRequestsOfType(RollbackRequest.class));
+    assertEquals(6, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+
+    List<ExecuteSqlRequest> requests = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class);
+
+    assertTrue(requests.get(0).getSql().equals(sql1));
```
olavloite: Prefer the use of `assertEquals(..)` over `assertTrue(..)` when possible. `assertEquals(..)` will give a more meaningful error message if the comparison fails.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1514 @@
...
+            .getId();
+    ByteString transactionIdForSql3 =
+        requests.stream()
+            .filter(req -> req.getSql().equals(sql3))
+            .findAny()
+            .get()
+            .getTransaction()
+            .getId();
+
+    assertTrue(transactionIdForSql2.equals(transactionIdForSql3));
```
olavloite: Also here and elsewhere in this file: use `assertEquals` instead of `assertTrue` when possible

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1514 @@
...
+    statements.add("query");
+    statements.add(sql3);
+
+    // commit request will not be sent to the Cloud Spanner
+    // because Cloud Spanner does not require read-only transactions
+    // to be committed or rolled back
+    statements.add("transaction");
+    statements.add("commit");
+
+    // update will throw an error because the readonly is activated
```
olavloite, line 456: This surprises me, because I thought that the `set readonly true` statement above only set the current transaction to read-only, not the entire session. And as the read-only transaction was just committed above, this statement would start a new read/write transaction.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1514 @@
...
+    // update will be executed as expected
+    statements.add("update");
+    statements.add(sql4);
+
+    // 1 commit request will be sent
+    statements.add("transaction");
+    statements.add("commit");
+
+    statements.add("transaction");
+    statements.add("set readonly True");
```
olavloite, line 440: Does this set the current __transaction__ to read-only, or the entire __session__ (connection)? (See also comment below)

Or should this be `set transaction readonly True`?

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1514 @@
...
+    assertEquals(
+        1,
+        getWireMessagesOfType(QueryMessage.class).stream()
+            .filter(qm -> qm.toString().contains("SERIALIZABLE"))
+            .count());
+  }
+
+  // The default version returned by PG Adapter is 1.0.
+  // Psycopg2 doesn't support Repeatable Read with the versions of Postgres lower than 9.1
+  // So, if the version of the Postgres (or PG Adapter in or case) is lower than 9.1,
```
olavloite: nit:
```suggestion
  // So, if the version of the Postgres (or PGAdapter in our case) is lower than 9.1,
```

src/test/python/StatementsWithTransactions.py
olavloite: Ah, makes sense. Thanks!

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: This here is only an alias. It is converted in the Python code to `connection.readonly = True` which sets the current session as Read Only. Quoted from the Psycopg2 documentation for [readonly](https://www.psycopg.org/docs/connection.html#connection.readonly):

> readonly
>
>Return or set the read-only status for the current session. Available values are True (new transactions will be in read-only mode), False (new transactions will be writable), None (use the default configured for the server by default_transaction_read_only).
>
>New in version 2.7.

Please note they mentioned "...(new **transactions** will be in read-only mode)...". Transaction(s) is plural here, so we can safely assume that all the transactions after `connection.readonly = True` will be started as readonly transactions.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: According to the Psycopg2 documentation, it will set the whole session as readonly. It doesn't send any statement like `set default_transaction_read_only = true` or something but all the transactions started after `set readonly true` will be in the read only mode. You can take a look at the following Wire Messages received by the PGAdapter to check more on this.

> Received Message: (Q) Query, with Payload: {Length: 10, SQL: BEGIN}
> Received Message: (Q) Query, with Payload: {Length: 29, SQL: Select * from some_table}
> Received Message: (Q) Query, with Payload: {Length: 62, SQL: insert into some_table(col1, col2) values(value1, value2)}
> Received Message: (Q) Query, with Payload: {Length: 63, SQL: insert into some_table3(col1, col2) values(value1, value2)}
> Received Message: (Q) Query, with Payload: {Length: 11, SQL: COMMIT}
> Received Message: (Q) Query, with Payload: {Length: 36, SQL: BEGIN;SET TRANSACTION READ ONLY}
> Received Message: (Q) Query, with Payload: {Length: 29, SQL: Select * from some_table}
> Received Message: (Q) Query, with Payload: {Length: 30, SQL: Select * from some_table3}
> Received Message: (Q) Query, with Payload: {Length: 11, SQL: COMMIT}
> Received Message: (Q) Query, with Payload: {Length: 36, SQL: BEGIN;SET TRANSACTION READ ONLY}
> Received Message: (Q) Query, with Payload: {Length: 62, SQL: insert into some_table(col1, col2) values(value1, value2)}

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTransactionTests.java
```diff
@@ -0,0 +1,1519 @@
...
+    assertEquals(ExecuteSqlRequest.class, requests.get(6).getClass());
+    assertEquals(sql3, ((ExecuteSqlRequest) (requests.get(6))).getSql());
+    ByteString transactionIdForRequest6 =
+        ((ExecuteSqlRequest) (requests.get(6))).getTransaction().getId();
+
+    assertEquals(transactionIdForRequest6, transactionIdForRequest5);
+
+    for (WireMessage wm : getWireMessagesOfType(QueryMessage.class)) {
+      System.out.println(wm);
+    }
```
olavloite, line 518: nit: remove

olavloite (approved):

PR #303, opened 2022-07-29
feat: set default PG version reported by PGAdapter to 14.1

olavloite:
Updates the default version that is reported by PGAdapter from `1.0` (which causes some drivers and tools to issue warnings about an old version of PG) to `14.1`. This also leads drivers to understand that it is OK to set an `application_name`, which again will make it easier to auto-detect more clients and drivers.


src/main/java/com/google/cloud/spanner/pgadapter/Server.java
```diff
@@ -16,31 +16,43 @@
...
 
+  static OptionsMetadata extractMetadata(String[] args, PrintStream out) {
+    out.printf("-- Starting PGAdapter version %s --\n", getVersion());
+    OptionsMetadata optionsMetadata = new OptionsMetadata(args);
+    out.printf("-- PostgreSQL version: %s -- \n", optionsMetadata.getServerVersion());
+
+    return optionsMetadata;
+  }
+
+  static void printError(Exception exception, PrintStream err, PrintStream out) {
```
pratickchokhani, line 48: Should we export it to a Utils for other classes to be used?

pratickchokhani (approved):
Have not validated the PG_JDBC_GET_COLUMNS_PREFIX statements.

src/main/java/com/google/cloud/spanner/pgadapter/Server.java
olavloite: I don't think that is necessary, as this is very specifically for errors during startup, which only happens here. The only reason I put it into a separate method was to be able to test it separately from the `main` method.

PR #306, opened 2022-07-31
fix: remove Spanner error prefixes

olavloite:
Removes the Spanner error code prefixes like INVALID_ARGUMENT: from all error
messages. Also introduces a specific PGException class that can be used to add
more information to error messages that are returned by PGAdapter, such as hints
or specific error codes.


src/main/java/com/google/cloud/spanner/pgadapter/error/SQLState.java
```diff
@@ -0,0 +1,85 @@
...
+  ProtocolViolation("08P01"),
+  TriggeredActionException("09000"),
+  FeatureNotSupported("0A000"),
+  InvalidTransactionInitiation("0B000"),
+  LocatorException("0F0000"),
+  InvalidLocatorSpecification("0F100"),
+  InvalidGrantor("0L000"),
+  InvalidGrantOperation("0LP01"),
+  InvalidRoleSpecification("0P000"),
+  // Skipping Query exceptions for now as we cannot easily identify them here
```
olavloite, line 47: This is copied from the existing SQLState enum. These will be added at a later moment.

PR #314, opened 2022-08-11
build: make unit tests skippable without skipping integration tests

olavloite:
Makes the unit tests skippable without also automatically skipping the integration tests.


olavloite:
Coverage changes are caused by the fact that the tests are logging less, meaning that they enter less log methods.

PR #344, opened 2022-09-02
docs: document psycopg2 usage

olavloite:
- Documents how to use psycopg2 with PGAdapter
- Adds tests for using Unix Domain Sockets with psycopg2
- Adds tests for DML/DDL semi-colon separated batches


gauravsnj (approved):
LGTM

PR #361, opened 2022-09-20
feat: added support for the set and show time zone

gauravsnj:


src/main/java/com/google/cloud/spanner/pgadapter/statements/SessionStatementParser.java
```diff
@@ -285,12 +285,23 @@ static SetStatement parseSetStatement(SimpleParser parser) {
           ErrorCode.INVALID_ARGUMENT,
           "Invalid SET statement: " + parser.getSql() + ". Expected configuration parameter name.");
     }
-    builder.setName(name);
-
-    if (!(parser.eatKeyword("to") || parser.eatToken("="))) {
-      throw SpannerExceptionFactory.newSpannerException(
-          ErrorCode.INVALID_ARGUMENT,
-          "Invalid SET statement: " + parser.getSql() + ". Expected TO or =.");
+    else if(name.toString().equalsIgnoreCase("TIME")){
```
olavloite: Instead of this, it would be better to add a `if (parser.eat("time", "zone"))` before the statement that reads the variable name (the `parser.readTableOrIndexName()` statement on line 282). If you do that, then you can enter a separate branch that specifically handles this type of statement, instead of having to mix it with the normal `set ...` statements.

src/main/java/com/google/cloud/spanner/pgadapter/statements/SessionStatementParser.java
```diff
@@ -279,20 +279,45 @@ static SetStatement parseSetStatement(SimpleParser parser) {
...
+    if (parser.eatKeyword("time", "zone")) {
+      builder.setName(new TableOrIndexName("TIMEZONE"));
+
+      String value = parser.parseExpression();
+      if (value == null) {
+        throw SpannerExceptionFactory.newSpannerException(
+            ErrorCode.INVALID_ARGUMENT,
+            "Invalid SET statement: " + parser.getSql() + ". Expected value.");
+      }
+      if ("default".equalsIgnoreCase(value) || "local".equalsIgnoreCase(value)) {
```
olavloite: `default` actually has a special meaning for all `SET` statements, and should not be translated here from `default` to `UTC`, but should cause the value to be set to `null`. The `SetStatement` class will then handle it from there.

The reason that it should be handled by the `SetStatement` is that the default for a connection can be different than the default of for example the server, as it can be set in the connection request.

src/main/java/com/google/cloud/spanner/pgadapter/statements/SessionStatementParser.java
```diff
@@ -279,20 +279,45 @@ static SetStatement parseSetStatement(SimpleParser parser) {
       // Ignore, this is the default.
       parser.eatKeyword("session");
     }
+    if (parser.eatKeyword("time", "zone")) {
```
olavloite, line 284: Hmmm... Now that I see this, I think having a completely separate branch for the special `set time zone` is a little too much. Could we keep the if statement here, and set the name as you do here now, and then add an else block that handles reading the variable name and the `to|=` for all other set statements? And then after that keep the parsing of the two statements in a single branch. That means that you also need to set a flag here to remember that it is the special `set time zone` statement, so you can handle the special `local` value. The rest should be the same for both types of statements, as far as I can see, so that should reduce the amount of code duplication.

src/main/java/com/google/cloud/spanner/pgadapter/statements/SessionStatementParser.java
```diff
@@ -361,6 +386,21 @@ static ShowStatement parseShowStatement(SimpleParser parser) {
       }
       return ShowStatement.createShowAll();
     }
+
+    if (parser.eatKeyword("time", "zone")) {
```
olavloite, line 371: Same here: I think it is good that we detect the special `show time zone` like this, but we can probably just add an `else` to read the variable name + check for `all` for all other statements, and after that keep the two branches together.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
```diff
@@ -2101,6 +2102,60 @@ public void testSetToEmpty() throws SQLException {
...
+  public void testSetTimeZone() throws SQLException {
+    try (Connection connection = DriverManager.getConnection(createUrl())) {
+      connection.createStatement().execute("set time zone 'IST'");
+      verifySettingValue(connection, "timezone", "IST");
+    }
+  }
+
+  @Test
+  public void testSetTimeZoneToDefault() throws SQLException {
+    try (Connection connection = DriverManager.getConnection(createUrl())) {
```
olavloite: If you were to change this to the following:

```suggestion
    try (Connection connection = DriverManager.getConnection(createUrl()
                + "?options=-c%20timezone=IST")) {
```

The the `set time zone default` should set the time zone to `IST` instead of `UTC`.

olavloite (approved):

PR #366, opened 2022-09-23
feat: replace sequences query with empty table

pratickchokhani:
Replace sequences query with empty table, as sequences are currently not supported in Cloud Spanner.


src/main/java/com/google/cloud/spanner/pgadapter/utils/PgJdbcCatalog.java
```diff
@@ -957,4 +957,9 @@ public class PgJdbcCatalog {
           + "inner join pg_catalog.pg_class rel on a.attrelid=rel.oid\n"
           + "inner join pg_catalog.pg_namespace n on rel.relnamespace=n.oid\n"
           + "inner join information_schema.columns c on n.nspname=c.table_schema and rel.relname=c.table_name and a.attname=c.column_name\n";
+
+  public static final String PG_JDBC_GET_SEQUENCES =
+      "select relname from pg_class where relkind='S'";
+  public static final String PG_JDBC_GET_SEQUENCES_REPLACEMENT =
+      "select 1 as relname LIMIT 0";
```
olavloite: nit: Make sure the column that is returned is of type `varchar` to correspond with the return type of the actual query.

```suggestion
      "select ''::varchar as relname LIMIT 0";
```

olavloite (approved):
Looks generally good to me, with a small nit on the replacement query + you need to run the code formatter to satisfy the lint check. You can run the code formatter by either:
1. mvn com.coveo:fmt-maven-plugin:format
2. mvn test

(`mvn test` will automatically invoke the code formatter and then run the tests, the former will run only the code formatter)

src/main/java/com/google/cloud/spanner/pgadapter/utils/PgJdbcCatalog.java
pratickchokhani: Done

PR #376, opened 2022-10-05
test: added the Django test setup and some simple django tests

gauravsnj:


src/test/java/com/google/cloud/spanner/pgadapter/python/Django/DjangoMockServerTest.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.cloud.spanner.pgadapter.AbstractMockServerTest;
+import java.util.Arrays;
+import org.junit.BeforeClass;
+
+public class DjangoMockServerTest extends AbstractMockServerTest {
+
+  @BeforeClass
+  public static void startMockSpannerAndPgAdapterServers() throws Exception {
+    doStartMockSpannerAndPgAdapterServers(
+        new MockSpannerServiceImpl(), "d", Arrays.asList("--server-version", "11.1"));
```
olavloite: This seems like a very specific version. Is there a reason for that?

src/test/java/com/google/cloud/spanner/pgadapter/python/Psycopg2/ITPsycopg2Test.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.cloud.spanner.pgadapter.python;
+package com.google.cloud.spanner.pgadapter.python.Psycopg2;
```
olavloite: nit: Java package names are always all lower-case

src/test/python/django/basic_test.py
```diff
@@ -0,0 +1,111 @@
...
+if __name__ == '__main__':
+  if len(sys.argv) < 4:
+    print('Invalid command line arguments')
+    sys.exit()
+  host = sys.argv[1]
+  port = sys.argv[2]
+
+  try:
+    create_django_setup(host, port)
+    flush()
```
olavloite: nit: could we call this `clearCache()`?

Also; why is this needed?

olavloite (approved):
Thanks, this looks good! I had a couple of minor nits and questions, but nothing blocking.

src/test/java/com/google/cloud/spanner/pgadapter/python/Django/DjangoBasicTest.java
```diff
@@ -0,0 +1,191 @@
...
+public class DjangoBasicTest extends DjangoTestSetup {
+
+  @Parameter public String host;
+
+  @Parameters(name = "host = {0}")
+  public static List<Object[]> data() {
+    return ImmutableList.of(new Object[] {"localhost"}, new Object[] {"/tmp"});
+  }
+
+  private ResultSet createResultSet(List<String> rows) {
```
olavloite: (You can leave this as is for now, but as a tip for further development): This works well for now as the sample is still very small, but I think that as you increase the number of classes (models) and the number of columns, it will be better to create a replica of your model classes in PHP in Java as well, and use that here.

src/test/java/com/google/cloud/spanner/pgadapter/python/Django/DjangoBasicTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+  @Test
+  public void testInsert() throws IOException, InterruptedException {
+    String sqlUpdate =
+        "UPDATE \"singers\" SET \"firstname\" = 'john', \"lastname\" = 'doe' WHERE \"singers\".\"singerid\" = 4";
+    String sqlInsert =
+        "INSERT INTO \"singers\" (\"singerid\", \"firstname\", \"lastname\") VALUES (4, 'john', 'doe')";
+
+    // Django first sends an update statement
+    // If the updated_rowcount returned by the server is 0, then only it sends the insert statement
```
olavloite: Is this the default behavior of Django, or do you need to configure something specific to get this behavior? The reason that I ask is that PostgreSQL supports [insert-or-update](https://www.postgresql.org/docs/15/sql-merge.html) statements, and I would have expected Django to use that by default instead of two separate DML statements.

src/test/java/com/google/cloud/spanner/pgadapter/python/Django/DjangoMockServerTest.java
gauravsnj: Django needs version 11 or later. Added the comment for the same

src/test/java/com/google/cloud/spanner/pgadapter/python/Psycopg2/ITPsycopg2Test.java
gauravsnj: done

src/test/python/django/basic_test.py
gauravsnj: This was not supposed to be here. Actually, Django was sending the UPDATE statement when calling the save method. So, I thought that maybe its something related to the cache as I read somewhere in the documentation that Django sometimes tries to optimize the statements by storing some results in the cache. But it was not the case. Django send the update statement every time.

src/test/java/com/google/cloud/spanner/pgadapter/python/Django/DjangoBasicTest.java
gauravsnj: Noted. Just one thing, you meant Python instead of PHP, right?

src/test/java/com/google/cloud/spanner/pgadapter/python/Django/DjangoBasicTest.java
gauravsnj: Yeah, it is the default behaviour of the Django and I was surprised too. There are some fields like `force_insert` and `force_update` in the `save` function to enforce the insert or update but by default, it works like this only.

PR #382, opened 2022-10-10
feat: added support for Django specific statement

gauravsnj:


src/main/java/com/google/cloud/spanner/pgadapter/statements/local/DjangoSpecificStatement.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.cloud.spanner.ResultSets;
+import com.google.cloud.spanner.Type;
+import com.google.cloud.spanner.Type.StructField;
+import com.google.cloud.spanner.connection.StatementResult;
+import com.google.cloud.spanner.pgadapter.statements.BackendConnection;
+import com.google.cloud.spanner.pgadapter.statements.BackendConnection.QueryResult;
+import com.google.common.collect.ImmutableList;
+
+@InternalApi
+public class DjangoSpecificStatement implements LocalStatement {
```
olavloite: nit: Could we rename this to something like `DjangoGetTableNamesStatement` and add a comment that this statement is used by Django to check whether the `migrations` tables exists. If it does not, then Django will assume that migrations are disabled, which makes it safe to mock this statement with an empty result set.

olavloite (approved):

src/main/java/com/google/cloud/spanner/pgadapter/statements/local/DjangoSpecificStatement.java
gauravsnj: done

PR #383, opened 2022-10-10
perf: do not create a new parser for each column

olavloite:
Convert a Spanner value directly to the PostgreSQL equivalent without creating a Parser object. This reduces the amount of memory that is allocated and must be garbage collected for each row that is sent to the PostgreSQL client.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/parsers/BooleanParser.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/parsers/BooleanParser.java
@@ -43,7 +43,7 @@ public class BooleanParser extends Parser<Boolean> {
 
   static {
     ByteConverter.bool(TRUE_VALUE_BYTES_BINARY, 0, true);
-    ByteConverter.bool(FALSE_VALUE_BYTES_BINARY, 0, true);
+    ByteConverter.bool(FALSE_VALUE_BYTES_BINARY, 0, false);
   }
   // See https://www.postgresql.org/docs/current/datatype-boolean.html
   private static final Set<String> TRUE_VALUES =
--- a/src/main/java/com/google/cloud/spanner/pgadapter/parsers/BooleanParser.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/parsers/BooleanParser.java
@@ -19,8 +19,10 @@ import com.google.cloud.spanner.ErrorCode;
 import com.google.cloud.spanner.ResultSet;
 import com.google.cloud.spanner.SpannerExceptionFactory;
 import com.google.cloud.spanner.Statement;
+import com.google.cloud.spanner.pgadapter.ProxyServer.DataFormat;
 import com.google.cloud.spanner.pgadapter.error.PGExceptionFactory;
 import com.google.common.collect.ImmutableSet;
+import java.nio.charset.StandardCharsets;
 import java.util.Locale;
 import java.util.Set;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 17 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/parsers/BooleanParser.java
```diff
@@ -34,6 +36,15 @@
...
   private static final String TRUE_VALUE = "t";
   private static final String FALSE_VALUE = "f";
+  private static final byte[] TRUE_VALUE_BYTES = new byte[] {'t'};
+  private static final byte[] FALSE_VALUE_BYTES = new byte[] {'f'};
+  private static final byte[] TRUE_VALUE_BYTES_BINARY = new byte[1];
+  private static final byte[] FALSE_VALUE_BYTES_BINARY = new byte[1];
+
+  static {
+    ByteConverter.bool(TRUE_VALUE_BYTES_BINARY, 0, true);
+    ByteConverter.bool(FALSE_VALUE_BYTES_BINARY, 0, true);
```
gauravsnj: Why are we passing `true` as the third parameter here? Shouldn't it be `false`?

src/main/java/com/google/cloud/spanner/pgadapter/parsers/DateParser.java
```diff
@@ -114,13 +116,28 @@ protected byte[] binaryParse() {
...
+  static byte[] convertToPG(Date value) {
+    LocalDate localDate = LocalDate.of(value.getYear(), value.getMonth(), value.getDayOfMonth());
     long days = localDate.toEpochDay() - PG_EPOCH_DAYS;
     int daysAsInt = validateRange(days);
     return IntegerParser.binaryParse(daysAsInt);
   }
 
+  public static byte[] convertToPG(ResultSet resultSet, int position, DataFormat format) {
+    switch (format) {
+      case SPANNER:
```
gauravsnj, line 131: Why haven't we added the conversion logic for Cloud Spanner format for some of the data types?

gauravsnj, line 131: If the conversion to Spanner format is not possible for these data types, can we add some more specific error message rather than falling back to `unknown data format`?

gauravsnj (approved):
LGTM with some small queries

src/main/java/com/google/cloud/spanner/pgadapter/parsers/DateParser.java
olavloite: In Java, case statements automatically fall through, so here the same logic is used for both when the format is `SPANNER` and `POSTGRESQL_TEXT`. This is different from some other programming languages where this would mean that there is no conversion for `SPANNER`.

olavloite: (See above)
The default case should not be reachable, but is required for the Java compiler, as it otherwise thinks that there are code paths possible that will not return a value for this method.

src/main/java/com/google/cloud/spanner/pgadapter/parsers/BooleanParser.java
olavloite: Great catch, thanks! That's absolutely right, and it was not caught by our tests for two reasons:
1. Most drivers will never use the binary format for booleans, because it does not really save any network traffic anyways (boolean values in text format are represented by `t` and `f`, so they take up just one byte, just as the binary format of 1 and 0)
2. Our tests mostly used `true` as the test value, which meant that they did not catch the fact that `false` was encoded as true.

I've added additional tests that use randomly generated values so we can catch this kind of errors in the future. That testing also turned out to detect a bug in the PostgreSQL JDBC driver, as that driver does not correctly decode binary boolean values. I've filed an issue and a PR for the JDBC driver in their repository for it.

PR #384, opened 2022-10-10
perf: write converted values directly to a byte buffer

olavloite:
Write the converted values directly to a byte buffer that can be reused for multiple rows. This reduces the amount of objects that are created and must be garbage collected for each result set that is returned by PGAdapter.


src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java
```diff
@@ -27,11 +29,62 @@
...
+
+  public ResultSet getResultSet() {
+    return resultSet;
+  }
+
+  public int convertResultSetRowToDataRowResponse() throws IOException {
+    buffer.reset();
+    outputStream.writeShort(resultSet.getColumnCount());
+    for (int column_index = 0; /* column indices start at 0 */
+        column_index < resultSet.getColumnCount();
```
gauravsnj, line 66: Can we use the `getColumnCount` method created above here or remove that method as its not used anywhere?

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/DataRowResponse.java
```diff
@@ -15,64 +15,29 @@
...
+  private final Converter converter;
 
-  public DataRowResponse(
-      DataOutputStream output,
-      IntermediateStatement statement,
-      ResultSet resultSet,
-      OptionsMetadata options,
-      QueryMode mode) {
+  public DataRowResponse(DataOutputStream output, Converter converter) {
     super(output, 0);
```
gauravsnj, line 29: Instead of moving the whole logic of converting the resultset to byte array and then copying the data from buffer present in the converter object, can we directly write into the outputStream in this class? Maybe something like below:
Instead of 
`columns.add(column);`
we can do
`this.outputStream.write(column, 0, column.length);`

src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java
```diff
@@ -27,11 +29,62 @@
...
+        outputStream.writeInt(column.length);
+        outputStream.write(column);
+      }
+    }
+    return buffer.size();
+  }
+
+  public void writeBuffer(DataOutputStream outputStream) throws IOException {
+    buffer.writeTo(outputStream);
+  }
```
gauravsnj, line 82: Why are we using `buffer` while we are not setting it anywhere? Do we mean `this.outputStream.writeTo(outputStream);` ?

gauravsnj (approved):
LGTM with some queries

src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java
olavloite: Good point. Removed the redundant `getColumnCount()` method.

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/DataRowResponse.java
olavloite: That would have been my preference as well, but it is unfortunately not possible. The reason is that the PG wire-protocol requires us to write the length of the message first. In order to get the length of the message, we first need to convert the Spanner row into the PG wire-protocol representation to determine the number of bytes. Then we can write the length, and then we can write the row data.

So to reduce the amount of memory that we need to reserve for this, we do the following:
1. We create a `ByteArrayOutputStream` that serves as a buffer in the `Converter`.
2. The buffer is used to place the converted row. This will allocate as much memory as is needed to convert a given row.
3. The data in the buffer indicates how large the PG wire-protocol representation of the row is, so we can then write the length of the message, and then write the contents of the buffer to the output stream
4. The buffer is reused for each row, meaning that we only need to allocate more memory if a following row is larger than all preceding rows.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java
olavloite: See also explanation above.

`buffer` is created when a `Converter` is created (see https://github.com/GoogleCloudPlatform/pgadapter/blob/cfb7cd2925e8d5d828b3e65216400edb7e3c7889/src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java#L40). This buffer is reused for each row that we convert, and automatically allocates more space as is needed to convert a given row.

gauravsnj: Got it. Thanks for the detailed explanation. But I am still not getting one thing that this `buffer`  is initialized on constructor but it is never filled anywhere. In the method `convertResultSetRowToDataRowResponse`, we are not writing anything to the `buffer` but the `outputStream`. Is there something internal which is passing the bytes from `outputStream` to `buffer`?

olavloite: Ah, yes, good question. Yes, what is happening here is the following:
1. We initialize `buffer` on [line 40](https://github.com/GoogleCloudPlatform/pgadapter/blob/cfb7cd2925e8d5d828b3e65216400edb7e3c7889/src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java#L40) with an initial buffer capacity of 256 bytes. That will grow dynamically if needed.
2. We initialize `outputStream` on [line 41](https://github.com/GoogleCloudPlatform/pgadapter/blob/cfb7cd2925e8d5d828b3e65216400edb7e3c7889/src/main/java/com/google/cloud/spanner/pgadapter/utils/Converter.java#L41) with `buffer` as its input argument. `outputStream` is of type `DataOutputStream` and requires another output stream where it can write its data. A `DataOutputStream` is just a convenient way of writing data using native Java types (e.g. writing something like an `int`). The `DataOutputStream` will convert this into bytes and write it to the underlying output stream. In this case, that is our previously defined `buffer`.

So the above means that when we write to `outputStream` in the `convertResultSetRowToDataResponse`, it will automatically convert whatever Java type that we write to it into the corresponding stream of bytes. Those bytes are then written to the underlying `buffer`.

PR #385, opened 2022-10-10
perf: copy binary can write directly to a byte buffer

olavloite:
Let binary copy operations write the converted data directly to a byte buffer instead of creating a parser that puts it in a temporary byte array, only to write it to the output stream later. This reduces the number of objects that are created for each row that is copied.


gauravsnj (approved):
LGTM

PR #398, opened 2022-10-14
build: double fork count for integration tests

olavloite:
Try to double the number of forks used for parallel integration tests to see if that can reduce build time.


olavloite:
Closing as it did not improve build time.

PR #426, opened 2022-10-24
feat: add setting for copy_commit_priority

olavloite:
Adds the setting 'spanner.copy_commit_priority' that can be used to set the Rpc priority for Commit requests for COPY operations. The default is MEDIUM to be consistent with Dataflow.

Fixes b/254330479

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/session/CopySettings.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/session/CopySettings.java
@@ -14,6 +14,8 @@
 
 package com.google.cloud.spanner.pgadapter.session;
 
+import com.google.cloud.spanner.Options.RpcPriority;
+import java.util.Locale;
 import java.util.Map;
 
 public class CopySettings {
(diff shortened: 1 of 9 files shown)


aseering (approved):
I'm once again a bit out of my depth in this codebase, but I like the objective and the UX of its implementation, and the code seems good insofar as I have the expertise to gauge such things.

src/main/java/com/google/cloud/spanner/pgadapter/session/CopySettings.java
```diff
@@ -86,6 +88,19 @@ public int getCommitTimeoutSeconds() {
     return sessionState.getIntegerSetting("spanner", "copy_commit_timeout", 300);
   }
 
+  /** Returns the request priority for commits executed by COPY operations. */
+  public RpcPriority getCommitPriority() {
```
thiagotnunes, line 92: nit: `getCopyCommitPriority()`

thiagotnunes (commented):
lgtm

src/main/java/com/google/cloud/spanner/pgadapter/session/CopySettings.java
olavloite: I don't think I agree with that suggestion. The class is already called `CopySettings`, so the additional `Copy` in the method name feels redundant. It's also not in line with the other method names in the class.

thiagotnunes: I missed the class context, please ignore

thiagotnunes (approved):

PR #427, opened 2022-10-24
fix: use copy commit timeout for all RPCs

olavloite:
Apply the spanner.copy_commit_timeout value for all RPCs that are executed as part of a commit during a COPY operation. This means that the timeout will also be applied to BeginTransaction and BatchCreateSessions if any of those are also needed during the commit operation. The timeout should also be applied to those, as those RPCs could also respond slowly if the COPY operation is causing the server to be overloaded.

### DIFF
--- a/src/test/java/com/google/cloud/spanner/pgadapter/nodejs/TypeORMMockServerTest.java
+++ b/src/test/java/com/google/cloud/spanner/pgadapter/nodejs/TypeORMMockServerTest.java
@@ -502,7 +502,8 @@ public class TypeORMMockServerTest extends AbstractMockServerTest {
     assertEquals(
         "some random string", insertRequest.getParams().getFieldsMap().get("p9").getStringValue());
     assertEquals(
-        "{\"key\":\"value\"}", insertRequest.getParams().getFieldsMap().get("p10").getStringValue());
+        "{\"key\":\"value\"}",
+        insertRequest.getParams().getFieldsMap().get("p10").getStringValue());
   }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/cloud/spanner/pgadapter/nodejs/TypeORMMockServerTest.java
```diff
@@ -502,7 +502,8 @@ public void testCreateAllTypes() throws IOException, InterruptedException {
     assertEquals(
         "some random string", insertRequest.getParams().getFieldsMap().get("p9").getStringValue());
     assertEquals(
-        "{\"key\":\"value\"}", insertRequest.getParams().getFieldsMap().get("p10").getStringValue());
+        "{\"key\":\"value\"}",
+        insertRequest.getParams().getFieldsMap().get("p10").getStringValue());
```
aseering, line 506: Not sure what this change is for?

aseering (approved):
I don't really know enough about the Java Client to comment on whether this does what it says on the tin, but I approve of what the PR says it's trying to accomplish, and the code at least looks like a plausible implementation of that objective.

src/test/java/com/google/cloud/spanner/pgadapter/nodejs/TypeORMMockServerTest.java
olavloite: Unrelated, but unfortunately gets forced in in this case: We use an automated code formatter, and it automatically changed this line.

thiagotnunes (approved):
lgtm

PR #468, opened 2022-11-08
test: Added some more Django tests

gauravsnj:


src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoTestSetup.java
```diff
@@ -61,4 +61,26 @@ public String executeBasicTests(int port, String host, List<String> options)
 
     return output.toString();
   }
+
+  public String executeTransactionTests(int port, String host, List<String> options)
```
olavloite: nit: Could we combine this method with the one above. As far as I can see, the only difference between the two is the Python script file that we are calling.

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoTransactionsTest.java
```diff
@@ -0,0 +1,166 @@
...
+    mockSpanner.putStatementResult(StatementResult.update(Statement.of(insertSQL2), 1));
+
+    List<String> options = new ArrayList<String>();
+    options.add("rollback");
+
+    String actualOutput = executeTransactionTests(pgServer.getLocalPort(), host, options);
+    String expectedOutput = "Transaction Rollbacked\n";
+
+    assertEquals(expectedOutput, actualOutput);
+    assertEquals(1, mockSpanner.countRequestsOfType(RollbackRequest.class));
```
olavloite, line 141: nit: Could we also add an assertion that there are no commit requests?

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoTransactionsTest.java
```diff
@@ -0,0 +1,166 @@
...
+    String actualOutput = executeTransactionTests(pgServer.getLocalPort(), host, options);
+    String expectedOutput = "Atomic Successful\n";
+
+    assertEquals(expectedOutput, actualOutput);
+    assertEquals(1, mockSpanner.countRequestsOfType(CommitRequest.class));
+
+  }
+
+
+
```
olavloite: nit: remove empty lines. It seems that you need to run the code formatter for this PR. This is done automatically when you run `mvn test`, or you can run only the code formatter with `mvn com.coveo:fmt-maven-plugin:format`.

olavloite (approved):

PR #478, opened 2022-11-11
refactor: use analyze to get statement params

### LINKED ISSUE #460: Preparing an invalid statement during a read/write transaction does not invalidate the transaction
Preparing an invalid statement during a transaction should put that transaction in the Aborted state. PGAdapter does not do that:

### PGAdapter

```
knut-test-db=> begin;
BEGIN
knut-test-db=*> prepare s as select * from foo;
ERROR:  relation "foo" does not exist - Statement: 'select * from foo'
knut-test-db=*> select 1;
 ?column? 
----------
        1
(1 row)

knut-test-db=*> rollback;
ROLLBACK
knut-test-db=> 
```

### PostgreSQL

```
knut-test-db=# select * from foo;
 id | colfoo  
----+---------
  1 | (1,One)
(1 row)

knut-test-db=# begin;
BEGIN
knut-test-db=*# prepare s as select * from bar;
ERROR:  relation "bar" does not exist
LINE 1: prepare s as select * from bar;
                                   ^
knut-test-db=!# select 1;
ERROR:  current transaction is aborted, commands ignored until end of transaction block
knut-test-db=!# 
```

olavloite:
Execute statements in `PLAN` mode to get parameter metadata instead of re-writing the SQL statements to a `SELECT`.

Fixes #460


src/main/java/com/google/cloud/spanner/pgadapter/metadata/DescribeResult.java
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+    return result;
+  }
+
+  static int maxParamNumber(StructType parameters) {
+    int max = 0;
+    for (int i = 0; i < parameters.getFieldsCount(); i++) {
+      try {
+        int paramIndex = Integer.parseInt(parameters.getFields(i).getName().substring(1));
```
pratickchokhani, line 66: Why is this required? What are the extra fields that are coming?

pratickchokhani (approved):

src/main/java/com/google/cloud/spanner/pgadapter/metadata/DescribeResult.java
olavloite: The parameters are all named `p1`, `p2`, ... `p10`, ...

So we need to parse the parameter index from the name by skipping the `p`, and we cannot use the position in the list as the index, as the order is textual, meaning that they are actually returned as (`p1`, `p10`, `p2`, ...)

PR #500, opened 2022-11-21
test: SQLAlchemy

olavloite:
Adds samples and tests for SQLAlchemy 1.4.


samples/python/sqlalchemy-sample/README.md
```diff
@@ -115,6 +116,12 @@ read/write transactions.
 
 ## Performance Considerations
 
+### Parameterized Queries
+`psycopg2` does not use parameterized queries. Instead, `psycopg2` will replace query parameters with
```
bvandiver: link to parameterized query doc in spanner docs?

olavloite: Added link to https://cloud.google.com/spanner/docs/sql-best-practices#postgresql_1

.github/workflows/units.yaml
```diff
@@ -26,6 +26,7 @@ jobs:
           python-version: '3.9'
       - run: python --version
       - run: pip install -r ./src/test/python/requirements.txt
+      - run: pip install -r ./src/test/python/sqlalchemy/requirements.txt
```
aseering, line 29: Probably out of scope for this PR, but do we have a strategy for dealing with potentially conflicting versions of dependencies?  Python's default behavior is to simply evaluate evaluate dependency statements in the order that they are seen, so it effectively resolves conflicts in favor of the most recently seen dependency declaration.

samples/python/sqlalchemy-sample/README.md
```diff
@@ -0,0 +1,169 @@
...
+| metadata.reflect()           | Cloud Spanner does not support all PostgreSQL `pg_catalog` tables. Using `metadata.reflect()` to get the current objects in the database is therefore not supported.                                                                                                |
+| DDL Transactions             | Cloud Spanner does not support DDL statements in a transaction. Add `?options=-c spanner.ddl_transaction_mode=AutocommitExplicitTransaction` to your connection string to automatically convert DDL transactions to [non-atomic DDL batches](../../../docs/ddl.md). |
+| Generated primary keys       | Manually assign a value to the primary key column in your code. The recommended primary key type is a random UUID. Sequences / SERIAL / IDENTITY columns are currently not supported.                                                                               |
+| INSERT ... ON CONFLICT       | `INSERT ... ON CONFLICT` is not supported.                                                                                                                                                                                                                          |
```
aseering, line 79: For Limitations that are described using SQL, could this section include "which common SQLALchemy function call or calls will cause it to generate this SQL?"  I'm not sure customers would necessarily know this offhand.  (_ping_ @bvandiver , we were discussing this a bit.)

samples/python/sqlalchemy-sample/README.md
```diff
@@ -0,0 +1,169 @@
...
+for workloads that are known to only execute read operations. Read-only transactions do not take any
+locks. You can create a separate database engine that can be used for read-only transactions from
+your default database engine by adding the `postgresql_readonly=True` execution option.
+
+```python
+read_only_engine = engine.execution_options(postgresql_readonly=True)
+```
+
+### Autocommit
+Using isolation level `AUTOCOMMIT` will supress the use of (read/write) transactions for each
```
aseering: Nit: `suppress` (two P's)

samples/python/sqlalchemy-sample/connect.py
```diff
@@ -0,0 +1,56 @@
...
+
+from sqlalchemy import create_engine, event
+from model import *
+import sys
+
+
+# Creates a database engine that can be used with the sample.py script in this
+# directory. This function assumes that the host, port and database name was
+# given as command line arguments.
+def create_test_engine(autocommit=False, options=""):
```
aseering: Nit:  I assume this code requires Python 3+?  (Python 2 has long since been deprecated, though it still sees significant use.)

If so, it's at least not a bad practice to use [type hints](https://docs.python.org/3/library/typing.html) in place of proper static typing.  There are differing opinions about when and how it's worth doing this.  My personal take is that it's a really good idea for library functions, and it's a minor hassle but never hurts to do elsewhere since it might enable a linter to catch a bug with static analysis.

samples/python/sqlalchemy-sample/connect.py
```diff
@@ -0,0 +1,56 @@
...
+
+def register_event_listener_for_prepared_statements(engine):
+  # Register an event listener for this engine that creates prepared statements
+  # for each connection that is created. The prepared statement definitions can
+  # be added to the model classes.
+  @event.listens_for(engine, "connect")
+  def connect(dbapi_connection, connection_record):
+    cursor_obj = dbapi_connection.cursor()
+    for model in BaseMixin.__subclasses__():
+      if not model.__prepare_statements__ is None:
```
aseering: Nit:  I would personally probably do `<x> is not None` rather than `not <x> is None`.  Functionally equivalent; just personal preference (reads more like English).

samples/python/sqlalchemy-sample/run_sample.py
```diff
@@ -0,0 +1,46 @@
...
+import sys
+
+
+def print_usage():
+  print()
+  print("Usage: `python run_sample.py <host> <port> <database>`")
+  print()
+
+
+if len(sys.argv) != 4:
```
aseering: Nit: The "do it right" solution here would be to use the `argparse` built-in [library](https://docs.python.org/3/library/argparse.html), and I bet it would be more concise and more maintainable too.  This should work ok, though.

samples/python/sqlalchemy-sample/sample.py
```diff
@@ -0,0 +1,329 @@
...
+    albums=create_random_albums(num_albums)
+  )
+
+
+# Creates `num_albums` random album rows.
+def create_random_albums(num_albums):
+  if num_albums == 0:
+    return []
+  albums = [None] * num_albums
+  for i in range(num_albums):
```
aseering, line 245: I don't often see lists being initialized to [None, None, ...] in Python and then populated.  Too easy to have an off-by-one and get an out of bounds exception or leave a None behind in the list.

I think it's more common to either initialize as `[]` and then `.append()` the new elements into the list, or (alternatively) to do a list comprehension:

```
albums = [
  Album(...)
  for i in range(num_albums)
]
```

The `.append()` has the disadvantage that Python has no good static way to preallocate storage for the list.  But that cost is low as compared to overheads inherent in the Python interpreter, so I usually don't see folks worrying about it...  IIRC the list comprehension approach doesn't bother to preallocate either, even though it could in this particular case (not sure if recent interpreters are more clever); it at least used to just use `.append()` under the hood.

samples/python/sqlalchemy-sample/sample.py
```diff
@@ -0,0 +1,329 @@
...
+    albums[i] = Album(
+      id="{}".format(uuid4()),
```
aseering: Nit:  Fine but I think `id=str(uuid4())` would work?, and would be more concise.

samples/python/sqlalchemy-sample/sample.py
```diff
@@ -0,0 +1,329 @@
...
+    print("Created Venue and Concert")
+
+
+# Prints the concerts currently in the database.
+def print_concerts():
+  with Session(read_only_engine) as session:
+    # Query all concerts and join both Singer and Venue, so we can directly
+    # access the properties of these as well without having to execute
+    # additional queries.
+    concerts = session.query(Concert) \
```
aseering: In my experience, Python folks try to avoid backslash-escaped newlines when feasible.  Not a hard rule but a preference.

In many cases, parens can serve to allow a giant long expression to actually be multi-line, rather than using newline-escaping to pretend that it's single line.  The body of a parenthetical is required to be a valid expression and is allowed to be multi-line.  (The Python parser knows the body of the parens is an expression so there's no requirement that newline mean "end of statement" at least until it sees the closing paren.)  So you can do things like:

```
  concerts = (
    session.query(Concert)
      .options(joinedload(Concert.venue))
      .options(...)
  )
```

Similarly, the large multi-line strings above can be wrapped in parens instead of using escaped newlines:

```
  a = (
    "abc"
    "def"
  )  # a == "abcdef"
```

I've even seen that written as lists or tuples, in code written by some Python "purists" who don't like the C behavior of magically concatenating consecutive string literals (to be fair the above does look confusingly like a tuple definition if you don't notice the missing commas):

```
  string = [
    f"abc",
    f"def {} {}".format(...),
    f"ghi",
  ].join('')
```

But unlike the first and second examples, that one actually has a semantically different meaning, so I'm not sure I'm personally a fan.

Bit of a tangent:  In Python, `(1,2,3)` is a _tuple_, `(1,2,3,)` is also a tuple (trailing commas are allowed), `(1,)` is _also_ a tuple, but `(1)` is an _expression_ and not a tuple.  It's not hard to accidentally turn a tuple into an expression when editing existing Python code.

samples/python/sqlalchemy-sample/model.py
```diff
@@ -0,0 +1,162 @@
...
+
+  name = Column(String(200))
+  description = Column(JSONB)
+
+  def __repr__(self):
+    return f"venues(" \
+           f"id={self.id!r}," \
+           f"name={self.name!r}," \
+           f"description={self.description!r}," \
+           f"created_at={self.created_at.astimezone(timezone.utc) if self.created_at else None!r}," \
```
aseering: Nit:  Perfectly valid Python, but I think it's at least mildly preferred to evaluate the expression outside of the string and store it in a local variable, and just reference variables inside the format string.  If nothing else, it keeps the string simple.  Especially when the format-string is already kinda long and doing a bunch of stuff.

samples/python/sqlalchemy-sample/test_add_singer.py
```diff
@@ -0,0 +1,28 @@
...
+
+singer = Singer()
+singer.id = "123-456-789"
+singer.first_name = "Myfirstname"
+singer.last_name = "Mylastname"
+singer.active = True
+# Manually set a created_at value, as we otherwise do not know which value to
+# add to the mock server.
+singer.created_at = datetime.fromisoformat("2011-11-04T00:05:23.123456+00:00"),
+add_singer(singer)
```
aseering, line 28: Are these test using a framework that I'm unfamiliar with (so something will actually run them as a test and print out results, etc)?

If so, cool!  Do you have a pointer?

If not, I'd encourage making them actual proper runnable tests.  [pytest](https://docs.pytest.org/en/7.2.x/) is probably the most straightforward way to do that with this code.  You'd have to wrap the logic above in a function `def test_<something>():` (for whatever value of `<something>` you prefer.  (Also the file needs to be named as `test_*.py`, but it already is.)  Then you just run `pytest` in the top-level directory, and it should find and run this and all other similarly-constructed tests in this folder, and produce output that's human-readable by default but xUnit-compatible with a flag.  (Note that none of these steps actually require importing pytest in the code in this repository so you don't need to add pytest as a dependency.)

If you want to be able to use pytest but also to be able to run these tests individually without having pytest installed, you could add the following to the bottom of each file:

```
if __name__ == '__main__':
  test_<something>()
```

Or, if you want a little more automation/hackery and you have a bunch of test functions in one file, maybe something like:
```
if __name__ == '__main__':
  for key in globals():
    if key.startswith("test_") and callable(globals()[key]):
      globals()[key]()
```

That way you just have a bunch of files containing functions named `test_*` (which is fine), and if the file is passed directly to the Python interpreter, it will be called.  But pytest will import these files (in which case `__name__` is not `"__main__"`, it's the name of the module), scan them for functions named `test_*`, and run the functions within its framework.

samples/python/sqlalchemy-sample/test_create_model.py
```diff
@@ -0,0 +1,23 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+from connect import create_test_engine
+from model import mapper_registry
+
+
+engine = create_test_engine(options="?options=-c spanner.ddl_transaction_mode"
+                                    "=AutocommitExplicitTransaction")
```
aseering, line 21: If you used pytest, create_test_engine() could possibly be a fixture or set of fixtures?

samples/python/sqlalchemy-sample/test_update_singer.py
```diff
@@ -0,0 +1,18 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+from sample import update_singer
+
+update_singer("123-456-789", "Newfirstname", "Newlastname")
```
aseering, line 18: If you used pytest (or Python unittest, etc), I'd recommend consolidating some of these tests.  Have several functions per file rather than just one test case per file.  You already have separate functions so I think the educational/readability value of having examples separated out by file is somewhat limited in this case.

src/test/python/sqlalchemy/all_types.py
```diff
@@ -0,0 +1,36 @@
...
+  col_bytea = Column(LargeBinary)
+  col_float8 = Column(Float)
+  col_int = Column(Integer)
+  col_numeric = Column(Numeric)
+  col_timestamptz = Column(DateTime(timezone=True))
+  col_date = Column(Date)
+  col_varchar = Column(String)
+  col_jsonb = Column(JSONB)
+
+  def __repr__(self):
```
aseering, line 24: Does SQLAlchemy really not have a builtin that does this?

... apparently maybe it doesn't?  I see an extension that adds it:  https://pypi.org/project/sqlalchemy-repr/

src/test/python/sqlalchemy/orm_error_in_read_write_transaction_continue.py
```diff
@@ -0,0 +1,35 @@
...
+      col_jsonb={"key1": "value1", "key2": "value2"}
+  )
+  session.add(row)
+  try:
+    session.commit()
+  except exc.SQLAlchemyError as e:
+    try:
+      # Trying to use the same transaction after an error is not possible.
+      session.get(AllTypes, 1)
+      # The following line should never be reached.
```
aseering, line 29: Worth an `assert`?

aseering (commented):
Reviewed; left a bunch of comments.  No major concerns about the code itself.  Though I'd like to better understand (and didn't have time to try to dig into) the test execution strategy here.

.github/workflows/units.yaml
olavloite: Currently we don't do it, but what we could do in the future is create separate workflows for testing different (Python) modules. That would isolate the dependencies that are installed to only those that are required for the specific module that we are testing.

samples/python/sqlalchemy-sample/README.md
olavloite: We explain that further in the sections below, but the entry for `SAVEPOINT` was missing. I've added that.

samples/python/sqlalchemy-sample/README.md
olavloite: Done

samples/python/sqlalchemy-sample/connect.py
olavloite: Done

samples/python/sqlalchemy-sample/connect.py
olavloite: Done

samples/python/sqlalchemy-sample/model.py
olavloite: I've moved the main logic to a separate function that can be used by each of these methods.

samples/python/sqlalchemy-sample/run_sample.py
olavloite: Thanks for the tip. That does clean it up quite a bit. Fixed.

samples/python/sqlalchemy-sample/sample.py
olavloite: Fixed (here and elsewhere in this file)

samples/python/sqlalchemy-sample/sample.py
olavloite: Changed to using `.append()`. That seems slightly more readable to me (as a non-Python expert)

samples/python/sqlalchemy-sample/sample.py
olavloite: Fixed

samples/python/sqlalchemy-sample/test_add_singer.py
olavloite: Yes, we have a separate test runner for this. We try to use Java for actually running all test cases, so we can then also do all assertions in Java and have one common test framework for all ecosystems. This also has the big advantage that we can use an in-memory mock Spanner server for testing features in all ecosystems.

The sample is for example tested twice:
1. A test runner that uses an in-memory mock Spanner server. We use this server to verify that we see the RPC calls that we actually expect, including the parameters / values that we expect to be passed in. The specific `test_add_singer.py` script is tested [from here](https://github.com/GoogleCloudPlatform/pgadapter/blob/ef00fb99a7a95f0385c0cf5ce788daf12100c2ae/src/test/java/com/google/cloud/spanner/pgadapter/python/sqlalchemy/SqlAlchemySampleTest.java#L517).
2. In addition, we [run the entire sample](https://github.com/GoogleCloudPlatform/pgadapter/blob/ef00fb99a7a95f0385c0cf5ce788daf12100c2ae/src/test/java/com/google/cloud/spanner/pgadapter/python/sqlalchemy/ITSQLAlchemySampleTest.java#L54) against a real Cloud Spanner instance to verify that the queries that are sent are actually supported by Cloud Spanner.

samples/python/sqlalchemy-sample/test_create_model.py
olavloite: See also above.

Our test framework spins up a mocked Cloud Spanner instance and a PGAdapter in-process, and then pass the network addresses and port numbers of those to Python as a command line argument. This makes the test framework reusable across multiple ecosystems. We currently use the same base setup for Python, Go, .NET and NodeJS.

samples/python/sqlalchemy-sample/test_update_singer.py
olavloite: See also above. Splitting the tests out across multiple files make them easier to run separately. Currently, each of these files correspond to one test case in Java, which can then be executed as a single unit. An alternative could be to combine them into one file and include the function name in the command line arguments that we send to Python to look up the function that should be called.

src/test/python/sqlalchemy/all_types.py
olavloite: I couldn't find any builtin function for this and retrofitted an example I found in a tutorial online somewhere (cannot find it . And so far we have been trying to keep our ORM samples as native as possible, i.e. not add any handy extensions unless there's no other choice (one exception to this is for example that we added an extra library to the Hibernate sample to be able to support `jsonb`, as Hibernate itself does not have native support for it).

src/test/python/sqlalchemy/orm_error_in_read_write_transaction_continue.py
olavloite: Added to this file now: https://github.com/GoogleCloudPlatform/pgadapter/blob/ef00fb99a7a95f0385c0cf5ce788daf12100c2ae/src/test/java/com/google/cloud/spanner/pgadapter/python/sqlalchemy/SqlAlchemyOrmTest.java#L607

aseering (approved):

PR #509, opened 2022-11-24
Hibernate support description in README

pratickchokhani:
Hibernate support description in README


olavloite (approved):
LGTM

nit: Follow the [conventional commits conventions](https://www.conventionalcommits.org/en/v1.0.0/) when creating a PR. That makes sure that the release plugin that we are using is picking up the change in the correct way, and includes it in the release notes.

PR #521, opened 2022-11-29
test: Django conditional expressions

gauravsnj:


src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoConditionalExpressionsTest.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  @Test
+  public void testConditionalAggregation() throws IOException, InterruptedException {
+    String sqlSelect =
+        "SELECT COUNT(\"singers\".\"singerid\") FILTER (WHERE \"singers\".\"firstname\" = 'hello') AS \"hello_count\", COUNT(\"singers\".\"singerid\") FILTER (WHERE \"singers\".\"firstname\" = 'world') AS \"world_count\" FROM \"singers\"";
+
+    List<String> result = new ArrayList<>();
+
+    result.add("int");
```
olavloite, line 110: Here and elsewhere: Could we use a more appropriate data structure for this instead of a list with specific keywords? I understand the need to use something like a simple list when sending the data to Python, but as far as I can tell, this list is only used in the Java test code.

src/test/python/django/conditional_expressions_test.py
```diff
@@ -0,0 +1,92 @@
...
+    sys.exit()
+  host = sys.argv[1]
+  port = sys.argv[2]
+
+  try:
+    create_django_setup(host, port)
+    from data.models import Singer
+  except Exception as e:
+    print(e)
+    sys.exit()
```
olavloite, line 87: Would it make more sense to call `sys.exit(1)` to indicate an exceptional exit?

olavloite (approved):
LGTM with some minor nits/questions

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoConditionalExpressionsTest.java
gauravsnj: Very Good Point. Since, these are just unit tests, I didn't think much about creating some proper class for this case. I am merging this PR for now. But in future, I'll definitely refactor this part to look better

src/test/python/django/conditional_expressions_test.py
gauravsnj: Done

PR #522, opened 2022-11-29
test: added tests for the aggregation in django

gauravsnj:


src/test/python/django/aggregation_tests.py
```diff
@@ -0,0 +1,87 @@
...
+    sys.exit()
+  host = sys.argv[1]
+  port = sys.argv[2]
+
+  try:
+    create_django_setup(host, port)
+    from data.models import Singer
+  except Exception as e:
+    print(e)
+    sys.exit(0)
```
olavloite: nit: I think `sys.exit(1)` makes more sense when there is an error

olavloite (approved):
The changes LGTM, but there are multiple test failures. Would you mind taking a look at those?

src/test/python/django/aggregation_tests.py
gauravsnj: Done

PR #533, opened 2022-12-06
feat: support TRUNCATE

olavloite:
Add support for the `TRUNCATE` statement. This is translated into one or more `DELETE` statements. The statement can be combined with the `spanner.autocommit_dml_mode` setting to make the statement non-atomic and usable for tables that contain a large number of rows.

go/pgadapter-truncate


src/main/java/com/google/cloud/spanner/pgadapter/statements/TruncateStatement.java
```diff
@@ -0,0 +1,155 @@
...
+
+  static ParsedTruncateStatement parse(String sql) {
+    Preconditions.checkNotNull(sql);
+
+    SimpleParser parser = new SimpleParser(sql);
+    ParsedTruncateStatement.Builder builder = new Builder();
+    if (!parser.eatKeyword("truncate")) {
+      throw PGExceptionFactory.newPGException("not a valid TRUNCATE statement: " + sql);
+    }
+    parser.eatKeyword("table");
```
pratickchokhani, line 130: Should we make this a static string to reduce memory?

src/main/java/com/google/cloud/spanner/pgadapter/statements/TruncateStatement.java
```diff
@@ -0,0 +1,155 @@
...
+    builder.only = parser.eatKeyword("only");
```
pratickchokhani, line 131: Same.

src/main/java/com/google/cloud/spanner/pgadapter/statements/TruncateStatement.java
```diff
@@ -0,0 +1,155 @@
...
+    builder.star = parser.eatToken("*");
+    while (parser.eatToken(",")) {
+      TableOrIndexName name = parser.readTableOrIndexName();
+      if (name == null) {
+        throw PGExceptionFactory.newPGException("invalid or missing table name");
+      }
+      builder.tables.add(name);
+    }
+    if (!parser.eatKeyword("continue", "identity")) {
+      builder.restartIdentity = parser.eatKeyword("restart", "identity");
```
pratickchokhani, line 146: Same.

pratickchokhani (approved):

src/main/java/com/google/cloud/spanner/pgadapter/statements/TruncateStatement.java
olavloite: That is actually not necessary in Java. String literals are automatically 'interned' (added to the internal String pool of the JVM), which means that repeating the same string literal throughout an application will not use more memory than creating a single String constant and reuse that. See https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#intern--

src/main/java/com/google/cloud/spanner/pgadapter/statements/TruncateStatement.java
olavloite: See above

src/main/java/com/google/cloud/spanner/pgadapter/statements/TruncateStatement.java
olavloite: See above

PR #536, opened 2022-12-08
test: added tests for Django PG Specific aggregates

gauravsnj:


src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
```diff
@@ -0,0 +1,230 @@
...
+          rowBuilder.addValues(value);
+        } else rowBuilder.addValues(Value.newBuilder().setStringValue(rows.get(j)).build());
+      }
+      resultSetBuilder.addRows(rowBuilder.build());
+    }
+    return resultSetBuilder.build();
+  }
+
+  @BeforeClass
+  public static void mockUnnecessarySql() {
```
olavloite: I think it would be better to add this to the base class instead of here, as this will probably be needed for other Python tests as well (and it cleans up this test, so it only contains code that is relevant for what we are actually testing here)

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
```diff
@@ -0,0 +1,230 @@
...
+
+    mockSpanner.putStatementResult(StatementResult.query(Statement.of(unnecessarySql), rs1));
+    mockSpanner.putStatementResult(StatementResult.query(Statement.of(unnecessarySql2), rs1));
+  }
+
+  @Test
+  public void strAggTest() throws IOException, InterruptedException {
+
+    String sql =
+        "SELECT \"singers\".\"firstname\", STRING_AGG(\"singers\".\"firstname\", '|' ) AS \"str\" FROM \"singers\" GROUP BY \"singers\".\"firstname\" LIMIT 21";
```
olavloite: See this thread for some tips on how to get rid of the `LIMIT 21` clause in these SQL statements: https://stackoverflow.com/questions/60797028/django-annotate-unexpected-limit-21

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
```diff
@@ -0,0 +1,230 @@
...
+
+    List<String> result = new ArrayList<>();
```
olavloite: Same as for the other tests: Let's try to use a more structured data type for this as long as it does not have to pass the border between Java and Python.

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
```diff
@@ -0,0 +1,230 @@
...
+    assertEquals(expectedOutput, actualOutput);
+
+    assertEquals(3, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+    assertEquals(sql, mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(2).getSql());
+  }
+
+  @Test
+  public void arrAggTest() throws IOException, InterruptedException {
+
+    String sql = "SELECT ARRAY_AGG(\"singers\".\"firstname\" ) AS \"arr\" FROM \"singers\"";
```
olavloite: Here and elsewhere: This generates a valid PostgreSQL SQL statement, but is this function supported in Spangres?

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
```diff
@@ -0,0 +1,230 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+@Category(PythonTest.class)
+public class DjangoPgAggregates extends DjangoTestSetup {
```
olavloite: nit: Unit test classes should end their name with `Test` (so in this case `DjangoPgAggregatesTest`)

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
```diff
@@ -0,0 +1,230 @@
...
+    result1.add("some_column");
+
+    ResultSet rs1 = createResultSet(result1, 1);
+
+    mockSpanner.putStatementResult(StatementResult.query(Statement.of(unnecessarySql), rs1));
+    mockSpanner.putStatementResult(StatementResult.query(Statement.of(unnecessarySql2), rs1));
+  }
+
+  @Test
+  public void strAggTest() throws IOException, InterruptedException {
```
olavloite: nit (here and elsewhere): the convention is to start test methods with the string `test`. So in this case `testStrAgg`.

src/test/python/django/pg_aggregates_test.py
```diff
@@ -0,0 +1,71 @@
...
+    print(Singer.objects.values('firstname').annotate(str=StringAgg('firstname',delimiter = '|')))
+  elif option == 'arr_agg':
+    print(Singer.objects.values('singerid').aggregate(arr=ArrayAgg('firstname')))
+
+
+if __name__ == '__main__':
+
+  if len(sys.argv) < 4:
+    print('Invalid command line arguments')
+    sys.exit()
```
olavloite, line 55: nit: `sys.exit(1)`

src/test/python/django/test.py
```diff
@@ -0,0 +1,56 @@
...
+
+  host = sys.argv[1]
+  port = sys.argv[2]
+
+  try:
+    create_django_setup(host, port)
+    from data.models import Singer
+  except Exception as e:
+    print(e)
+    sys.exit()
```
olavloite: nit: `sys.exit(1)`

src/test/python/django/test.py
```diff
@@ -0,0 +1,56 @@
+''' Copyright 2022 Google LLC
```
olavloite: Is this file needed? I don't think it's actually being used in this PR.

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
gauravsnj: Yes, I first tested these on a sample database in the cloud spanner via PG Adapter and then only added the unit tests

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
gauravsnj: Yes, I have that action item on my mind.

src/test/python/django/pg_aggregates_test.py
gauravsnj: done

src/test/python/django/test.py
gauravsnj: Thanks for spotting this. I have removed the file

src/test/python/django/test.py
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoPgAggregates.java
gauravsnj: Thats really helpful. Will keep that in mind for further tests

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoTestSetup.java
```diff
@@ -16,15 +16,139 @@
...
+          rowBuilder.addValues(value);
+        } else rowBuilder.addValues(Value.newBuilder().setStringValue(rows.get(j)).build());
+      }
+      resultSetBuilder.addRows(rowBuilder.build());
+    }
+    return resultSetBuilder.build();
+  }
+
+  @BeforeClass
+  public static void mockUnnecessarySql() {
```
olavloite: nit: Could we rename this to something like `mockDriverInternalSql`? These are actual queries that Django will execute, and without the additional common table expressions for `pg_namespace` and `pg_type`, Django would not work.

olavloite (approved):
Thanks for applying the changes. LGTM.

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoTestSetup.java
gauravsnj: Done. And since these tests were passing on actual database also, it looks like PG Adapter is handling these statements well enough as of now.

PR #537, opened 2022-12-09
sample: Enabled limit query in hibernate sample application

pratickchokhani:
Added read only transaction optimization in hibernate README.md


samples/java/hibernate/README.md
```diff
@@ -79,4 +79,17 @@ public class User {
     @GeneratedValue
     private UUID id;
 }
+```
+
+## Optimization
+
+### Transaction
+Hibernate by default initiates every transaction as read-write even when there
+are only read queries. This causes over-head due to the locks required for read-write
```
olavloite: nit: 
```suggestion
are only read queries. This causes overhead due to the locks required for read-write
```

samples/java/hibernate/README.md
```diff
@@ -79,4 +79,17 @@ public class User {
...
+transaction.
```
olavloite: nit: 

```suggestion
transactions.
```

samples/java/hibernate/README.md
```diff
@@ -79,4 +79,17 @@ public class User {
...
+This can be optimized by initiating read only transaction for read queries.
```
olavloite:
```suggestion
This can be optimized by initiating read only transactions for workloads that only execute read queries.
```

samples/java/hibernate/README.md
```diff
@@ -79,4 +79,17 @@ public class User {
...
+
+```java
+Session session = entityManager.unwrap(Session.class);
```
olavloite: This will make the entire session read-only. Would it be possible to also show an example for how you could do this for a single transaction?

Furthermore: We should add a test that verifies that this actually works as expected, and does start a read-only instead of a read/write transaction.

pratickchokhani: Have removed this section as this was not working in testing. Will take is up as a separate bug.

olavloite (approved):

PR #539, opened 2022-12-09
docs: added sample application model for Django

gauravsnj:
Added the model part of the sample application for the Django


samples/python/django/models.py
```diff
@@ -0,0 +1,76 @@
...
+  release_date = models.DateField()
+  cover_picture = models.BinaryField()
+  singer_id = models.ForeignKey(Singer, on_delete=models.CASCADE())
+  createdAt = models.DateTimeField()
+  updatedAt = models.DateTimeField()
+
+class Track(models.Model):
+  class Meta():
+    db_table = 'tracks'
+    unique_together = ('id', 'track_number')
```
olavloite: We should probably add some additional comments / documentation here. If I understand it correctly, Django does not support composite primary keys, so this is a way of telling Django that these two will be unique together. However; I think that Django will only use this information to generate the schema and add a unique constraint for these two columns, right? If so, we should probably not add it here, as we don't want an additional unique constraint on these two columns to be generated (but maybe put it in. a comment that these two are guaranteed to be unique because they form the underlying primary key).

samples/python/django/models.py
```diff
@@ -0,0 +1,76 @@
...
+class Singer(models.Model):
+  class Meta():
+    db_table = 'singers'
+  id = models.IntegerField(primary_key=True, null=False)
+  first_name = models.CharField()
+  last_name = models.CharField(null=False)
+
+  @property
+  def _get_full_name(self):
+    return '{first_name} {last_name}'.format(first_name=self.first_name, last_name=self.last_name)
```
olavloite: This will not use a `full_name` generated column in the database. Is there any way in Django that we could use a generated column instead of doing the calculation here? (I know that the calculation itself is simple in this case, the intention here is to show developers how they could use a generated column with the ORM)

samples/python/django/models.py
```diff
@@ -0,0 +1,76 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+'''
+
+from django.db import models
+from django.db.models import Q, F
+class Singer(models.Model):
+  class Meta():
+    db_table = 'singers'
+  id = models.IntegerField(primary_key=True, null=False)
```
olavloite: Is it possible to move the properties that are common to all models into some kind of an abstract base model. That would then include the properties `id`, `createdAt` and `updatedAt`.

samples/python/django/models.py
```diff
@@ -0,0 +1,76 @@
...
+  cover_picture = models.BinaryField()
+  singer_id = models.ForeignKey(Singer, on_delete=models.CASCADE())
+  createdAt = models.DateTimeField()
+  updatedAt = models.DateTimeField()
+
+class Track(models.Model):
+  class Meta():
+    db_table = 'tracks'
+    unique_together = ('id', 'track_number')
+  id = models.ForeignKey(Album, on_delete=models.CASCADE())
```
olavloite: Will this cause Django to generate additional `DELETE` statements for the Tracks that are related to an Album? If so, we should remove this and maybe add a comment. What we want to show here is how they should create their model if they have an interleaved table structure with a `ON DELETE CASCADE` clause in the database. That is; the database does the deletion automatically and does not depend on statements that are generated by the ORM.

samples/python/django/models.py
```diff
@@ -0,0 +1,76 @@
...
+  track_number = models.BigIntegerField(null=False)
+  title = models.CharField(null=False)
+  sample_rate = models.DecimalField()
```
olavloite: This should be a `float8`. Is it possible to make a distinction between `decimal` and floating point types in Django?

gauravsnj: Done

samples/python/django/models.py
gauravsnj: Yes, it would've caused Django to generate some extra queries. Hence, I've changed it to DO_NOTHING

samples/python/django/models.py
gauravsnj: Moved `created_at` and `updated_at` to the base mode but since we want `id` to be primary key, it needs to be in each of the models itself.

samples/python/django/models.py
gauravsnj: Yes, it would just create a new variable(not field) with the concatenation of `first_name` and `last_name`. Hence, removed it

samples/python/django/models.py
gauravsnj: Done

samples/python/django_sample/sample_app/model.py
```diff
@@ -0,0 +1,84 @@
...
+  class Meta():
+    abstract = True
+  created_at = models.DateTimeField()
+  updated_at = models.DateTimeField()
+
+class Singer(models.Model):
+  class Meta():
+    db_table = 'singers'
+
+  id = models.CharField(primary_key=True, null=False)
```
olavloite, line 29: Could we move `id` into `BaseModel` as well? AFAICT, it's the same for all models.

samples/python/django_sample/sample_app/model.py
```diff
@@ -0,0 +1,84 @@
...
+  title = models.CharField(null=False)
+  marketing_budget = models.DecimalField()
+  release_date = models.DateField()
+  cover_picture = models.BinaryField()
+  singer = models.ForeignKey(Singer, on_delete=models.DO_NOTHING)
+  created_at = models.DateTimeField()
+  updated_at = models.DateTimeField()
+
+# Here, id is a column that is supposed to be primary key by Django.
```
olavloite: nit: I think it would be better to move this comment down to the actual `id` field. Now it's not clear which element this refers to.

samples/python/django_sample/sample_app/model.py
```diff
@@ -0,0 +1,84 @@
...
+# But id column will just have a unique index in the actual table.
+# In the actual table, (album_id, track_number) will be the primary key.
+# This is done because Django doesn't support composite primary keys.
+# But we need to have a composite primary key
+# due to the interleaved structure of the actual table
```
olavloite:
```suggestion
# This is done because Django doesn't support composite primary keys,
# but we need to have a composite primary key due to the fact that
# the "tracks" table is interleaved in "albums".
```

olavloite (approved):

samples/python/django_sample/sample_app/model.py
gauravsnj: Since `id` is supposed to be the primary key for each of the model, moving it to the `BaseModel` is gonna cause problem as it would then work like the primary key only to the `BaseModel` and not to all the children models and Django will create another primary key Integer field for each of the children models. I tried to putting it to the base class but then insertions were not working as expected

samples/python/django_sample/sample_app/model.py
gauravsnj: Done

samples/python/django_sample/sample_app/model.py
olavloite: Ah, OK. Thanks for the explanation, that makes sense to me.

PR #556, opened 2022-12-20
test: Django Sample Application

gauravsnj:


samples/python/django/sample.py
```diff
@@ -0,0 +1,145 @@
...
+import sys
+import datetime
+import django
+import pytz
+from django.db import connection
+from django.db.transaction import atomic
+from django.db import transaction
+x = 0
+
+def get_singer(singer_id):
```
olavloite: nit: Could we rename these `get_` methods to something like `create_test_singer` etc.? `get_singer` sounds like reading the singer from the database.

samples/python/django/sample.py
```diff
@@ -0,0 +1,145 @@
...
+  global x
+  x += 1
+  return Singer(id=singer_id, first_name='singer', last_name=str(x), full_name= 'singer'+str(x), active=True, created_at=datetime.datetime.now(pytz.UTC), updated_at=datetime.datetime.now(pytz.UTC))
+
+def get_album(album_id, singer=None):
+  global x
+  x += 1
+  return Album(id=album_id, singer=singer, title='album'+str(x), marketing_budget = 200000, release_date = datetime.date.today(), cover_picture= b'hello world',created_at=datetime.datetime.now(pytz.UTC), updated_at=datetime.datetime.now(pytz.UTC))
```
olavloite: nit: split this over multiple lines.

samples/python/django/sample.py
```diff
@@ -0,0 +1,145 @@
...
+
+def get_venue(venue_id):
+  global x
+  x += 1
+  return Venue(id=venue_id, name='venue'+str(x), description='description'+str(x), created_at=datetime.datetime.now(pytz.UTC), updated_at=datetime.datetime.now(pytz.UTC))
+
+def get_concert(concert_id, venue = None, singer = None):
+  global x
+  x += 1
+  return Concert(id=concert_id, venue=venue, singer=singer, name='concert'+str(x), start_time=datetime.datetime.now(pytz.UTC), end_time=datetime.datetime.now(pytz.UTC)+datetime.timedelta(hours=1), created_at=datetime.datetime.now(pytz.UTC), updated_at=datetime.datetime.now(pytz.UTC))
```
olavloite: nit: here also, split this over multiple lines for better readability

samples/python/django/sample.py
```diff
@@ -0,0 +1,145 @@
...
+@atomic(savepoint=False)
+def test_adding_data():
+  singer = get_singer('1')
+  singer.save()
+
+  album = get_album('1', singer)
+  album.save()
+
+  track = get_track('1', '2', album)
+  track.save(force_insert=True)
```
olavloite, line 81: Why do we need force_insert here and not for the others?

samples/python/django/sample.py
```diff
@@ -0,0 +1,145 @@
...
+  return Concert(id=concert_id, venue=venue, singer=singer, name='concert'+str(x), start_time=datetime.datetime.now(pytz.UTC), end_time=datetime.datetime.now(pytz.UTC)+datetime.timedelta(hours=1), created_at=datetime.datetime.now(pytz.UTC), updated_at=datetime.datetime.now(pytz.UTC))
+
+
+def create_tables():
+  file = open('create_data_model.sql', 'r')
+  ddl_statements = file.read()
+  with connection.cursor() as cursor:
+    cursor.execute(ddl_statements)
+
+@atomic(savepoint=False)
```
olavloite, line 72: Is `savepoint=False` required? I would have thought that it was the default.

samples/python/django/sample.py
```diff
@@ -0,0 +1,145 @@
...
+  except:
+    raise Exception("Album1 doesn't belong to singer1")
+
+  singer2 = get_singer('2')
+  singer2.save()
+
+  global x
+  x += 1
+  album2 = singer2.album_set.create(id='2', title='album'+str(x), marketing_budget=250000, cover_picture=b'new world', created_at=datetime.datetime.now(pytz.UTC), updated_at=datetime.datetime.now(pytz.UTC))
+  album2.singer_id == singer2.id
```
olavloite: Is this necessary when we call `singer2.album_set.create(..)` above? I would expect the album to automatically be associated with singer2 in that case.

samples/python/django/sample.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def create_tables():
+  file = open('create_data_model.sql', 'r')
+  ddl_statements = file.read()
+  with connection.cursor() as cursor:
+    cursor.execute(ddl_statements)
+
+@atomic(savepoint=False)
+def test_adding_data():
```
olavloite: As this is a sample application, I think it would be better to rename these methods to something like just `add_data` or `create_singers_and_albums`, and also remove the test code (so no assertions). Instead, the sample functions can print the data that they read/create, and we can have a separate test that just runs the sample and verifies that the expected data is being printed. That will make the sample look more like sample code and less like test code.

samples/python/django/sample.py
gauravsnj: The reason is because in case of using `save()`, Django will first send `UPDATE` statement and then send an `INSERT` statement if no rows are updated. Now, the `UPDATE` query will look some like `UPDATE tracks set tracks.id = 1, tracks.track_number = '2' ....... where tracks.track_id='1'`. This query will be generated because Django thinks that `track_id` is the primary key but in the actual database combination of `(id, track_number)` is the primary key. So, the database will throw an error something similar to key column cannot be updated.

So, there are two ways to tackle this, either use `force_insert = True` with `save` or use `objects.create()`. The second way will look something like follows:

`Track.objects.create(track_id='1', album=album, track_number='2', title='track'+str(x), sample_rate=124.543, created_at=datetime.datetime.now(pytz.UTC), updated_at=datetime.datetime.now(pytz.UTC))`

samples/python/django/sample.py
gauravsnj: By default, `savepoint=True` is there but that doesn't matter in this case as it is a simple atomic block. `savepoints` come into picture in the case of nested atomic blocks and for nested atomic blocks, we will have to manually do `savepoints=False`. 

In this particular case, we can remove `savepoint=False` and it will work fine. But I thought of including it as it will remind people that in general, savepoints are not supported by us

samples/python/django/sample.py
gauravsnj: Thanks for spotting this!! I actually intended to `assert`, if this condition holds true or not (that's why I use `==` and not `=`) and yes it would automatically associate it with the `singer2`

samples/python/django/sample.py
gauravsnj: done

samples/python/django/sample.py
gauravsnj: done

samples/python/django/sample.py
gauravsnj: done

samples/python/django/sample.py
gauravsnj: done

samples/python/django/sample.py
olavloite: Ah, thanks for the detailed explanation! But if I understand this correctly, it also means that we can never update a row of an interleaved table, right?

Would setting `Field.editable=False` on the fields fix that? See https://docs.djangoproject.com/en/4.1/ref/models/fields/#editable

Otherwise (and unless there's some other way of marking these fields as non-updateable), I'm afraid that we have to ditch support for interleaved tables in Django. Not being able to update a row seems like too big of a limitation.

samples/python/django/sample.py
olavloite: OK, makes sense.

olavloite (approved):
LGTM. There's one open question regarding interleaved tables, but we can pick that up in a separate PR if necessary.

PR #557, opened 2022-12-20
test: added integration tests for Django

gauravsnj:


samples/python/django/sample.py
```diff
@@ -145,6 +146,8 @@ def transaction_rollback():
 if __name__ == "__main__":
 
   try:
+    tables_created = 0
```
olavloite: nit: could this be a boolean?

src/test/java/com/google/cloud/spanner/pgadapter/python/django/ITDjangoTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+    testEnv.stopPGAdapterServer();
+    testEnv.cleanUp();
+  }
+
+  @Test
+  public void test() throws IOException, InterruptedException {
+    String actualOutput = executeIntegrationTests();
+    String expectedOutput = "Django Test Completed Successfully\n";
+    assertTrue(actualOutput.contains(expectedOutput));
```
olavloite: nit: It's good to add the `actualOutput` as the first argument to a call to `assertTrue`. That will cause JUnit to print out the contents of `actualOutput` if the assertion fails, which can make it easier to determine the reason that the assertion failed (especially if the assertion fails during a CI run).

src/test/java/com/google/cloud/spanner/pgadapter/python/django/ITDjangoTest.java
```diff
@@ -0,0 +1,88 @@
...
+      writer.write(originalDjangoSetting);
+    }
+    testEnv.stopPGAdapterServer();
+    testEnv.cleanUp();
+  }
+
+  @Test
+  public void test() throws IOException, InterruptedException {
+    String actualOutput = executeIntegrationTests();
+    String expectedOutput = "Django Test Completed Successfully\n";
```
olavloite: nit: Could we change the text at the end of the sample to
```suggestion
    String expectedOutput = "Django Sample Completed Successfully\n";
```

olavloite (approved):

samples/python/django/sample.py
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/ITDjangoTest.java
gauravsnj: Done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/ITDjangoTest.java
gauravsnj: Done

PR #560, opened 2022-12-22
docs: added documentation for django

gauravsnj:


samples/python/django/README.md
```diff
@@ -0,0 +1,82 @@
...
+Cloud Spanner supports the following data types in combination with `Django`.
+
+| PostgreSQL Type                        | Django Model Field                           |
+|----------------------------------------|----------------------------------------------|
+| boolean                                | BoolField                                    |
+| bigint / int8                          | BigIntegerField                              |
+| varchar                                | CharField                                    |
+| text                                   | CharField                                    |
+| float8 / double precision              | FloatField                                   |
+| numeric(x, y)                          | DecimalField(max_digits=x, decimal_places=y) |
```
olavloite: nit: Cloud Spanner does not support type modifiers for `numeric`. That is; only `numeric` and not `numeric(x, y)` is supported.

samples/python/django/README.md
```diff
@@ -0,0 +1,82 @@
...
+See [sample-schema.sql](src/main/resources/sample-schema-sql) for the data model for this example.
+
+### Generated Primary Keys
+`Sequences` are not supported. Hence, auto increment primary key is not supported and should be replaced with primary key definitions that
+are manually assigned. See https://cloud.google.com/spanner/docs/schema-design#primary-key-prevent-hotspots
+for more information on choosing a good primary key. This sample uses UUIDs that are generated by the client for primary
+keys.
+
+```java
+public class User {
```
olavloite: Replace with a Python example

samples/python/django/README.md
```diff
@@ -0,0 +1,82 @@
...
+	// This uses auto generated UUID.
+    @Id
+    @Column(columnDefinition = "varchar", nullable = false)
+    @GeneratedValue
+    private UUID id;
+}
```
olavloite: We should add a 'Performance Considerations' section for Django as well. See https://github.com/GoogleCloudPlatform/pgadapter/blob/e7c4797ce35cd0fccc688a1269731762cd1c9c86/samples/python/sqlalchemy-sample/README.md#performance-considerations for an example.

samples/python/django/README.md
```diff
@@ -0,0 +1,82 @@
...
+| float8 / double precision              | FloatField                                   |
+| numeric(x, y)                          | DecimalField(max_digits=x, decimal_places=y) |
+| timestamptz / timestamp with time zone | DateTimeField                                |
+| bytea                                  | BinaryField                                  |
+| date                                   | DateField                                    |
+
+## Limitations
+The following limitations are currently known:
+
+| Limitation                           | Workaround                                                                                                                                         |
```
olavloite: This table should at least also include:
1. The limitation that interleaved tables are not fully supported, other than with some workarounds that come with its own limitations (including the lmitations around saving / updating rows in an interleaved table).
2. Nested transactions / savepoints.

README.md
```diff
@@ -10,6 +10,7 @@ PGAdapter can be used with the following drivers and clients:
 3. `pgx`: Version 4.15 and higher are supported. See [pgx support](docs/pgx.md) for more details.
 4. `psycopg2`: Version 2.9.3 and higher (but not `psycopg3`) are supported. See [psycopg2](docs/psycopg2.md) for more details.
 5. `node-postgres`: Version 8.8.0 and higher are supported. See [node-postgres support](docs/node-postgres.md) for more details.
+6. `django`: Version 4.1.1 and higher are supported. See [django support](docs/django.md) for more details
```
olavloite: Move this down to the 'Frameworks' section, as django is not a driver, but an ORM.

Also change the text to something like:
```suggestion
6. `django`: Version 4.1.1 and higher have experimental support. See [django support](docs/django.md) for more details. See also [this directory](samples/python/django) for a sample application using `django`.
```

docs/django.md
```diff
@@ -0,0 +1,45 @@
...
+  -x
+```
+
+Then connect to PGAdapter and use `django` by including the following in your `setting.py` file:
+
+```python
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'NAME': 'postgres',
```
olavloite, line 33: Is this the name of the Cloud Spanner database? If so, we should change it to `my-database`. That makes it clearer to the user that they need to change it to their own database name.

docs/django.md
```diff
@@ -0,0 +1,45 @@
...
+        'USER': 'postgres',
+        'PASSWORD': 'postgres',
```
olavloite, line 35: Can we remove these? PGAdapter will ignore them, so if django/psycopg2 both accept the connection without these, then we should remove it from the example.

samples/python/django/README.md
olavloite: ping: this still needs to be changed

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+|----------------------------------------|----------------------------------------------|
+| boolean                                | BoolField                                    |
+| bigint / int8                          | BigIntegerField                              |
+| varchar                                | CharField                                    |
+| text                                   | CharField                                    |
+| float8 / double precision              | FloatField                                   |
+| numeric(x, y)                          | DecimalField(max_digits=x, decimal_places=y) |
+| timestamptz / timestamp with time zone | DateTimeField                                |
+| bytea                                  | BinaryField                                  |
+| date                                   | DateField                                    |
```
olavloite: Can we add `jsonb` to this list and to the sample? The `Venue.description` field is intended as the `jsonb` sample field (it's currently `text` in some older samples)

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Limitations
+The following limitations are currently known:
+
+| Limitation             | Workaround                                                                                                                                                                                                                                                                                          |
+|------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Annotate               | Cloud Spanner does not support selecting columns in an aggregation query that are not involved in the aggregation functions or `GROUP BY` clause. Due to this, many times performing `annotate` function over a querySet might fail.                                                                |
```
olavloite:
```suggestion
| Annotate               | Cloud Spanner does not support selecting columns in an aggregation query that are not included in the aggregation functions or `GROUP BY` clause. Adding an aggregation on a query using `annotate` will therefore often generate an unsupported query.                                                                |
```

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+| Generated primary keys | Cloud Spanner does not support `sequences`. Auto-increment primary key is not supported. Avoid using fields like `AutoField`, `BigAutoField`, etc. . The recommended type of primary key is a client side generated `UUID` stored as a string.                                                      |
+| Savepoint              | Cloud Spanner does not support `savepoints` and hence, functions related to `savepoints` in django are not supported in the PGAdapter.                                                                                                                                                              |
```
olavloite:
```suggestion
| Savepoint              | Cloud Spanner does not support `savepoints`. Creating or rolling back to a `Savepoint` in django is not supported in PGAdapter.                                                                                                                                                              |
```

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+| Nested Atomic Blocks   | Nested atomic blocks make use of `savepoints`, which is currently not supported by Cloud Spanner. Hence, while using nested atomic blocks make sure to set `savepoint=False`, in the argument.                                                                                                      |
```
olavloite:
```suggestion
| Nested Atomic Blocks   | Nested atomic blocks use `savepoints`, which is currently not supported by Cloud Spanner. Ensure to set `savepoint=False` when using nested atomic blocks.                                                                                                      |
```

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+| Interleaved Table      | Django doesn't support composite primary keys and interleaved table in cloud spanner requires a composite primary key and on the top of that, it is also not possible to create a django model without a primary key. A possible workaround around for using interleaved table is documented below. |
```
olavloite:
```suggestion
| Interleaved Table      | Django does not support composite primary keys. Interleaved tables in Cloud Spanner require composite primary keys. A possible workaround for using interleaved tables is documented below. |
```

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+### Annotate
+`annotate` is not fully supported in the PGAdapter.
```
olavloite, line 70: Add more context and a concrete example for something that does not work. And if there's a workaround, then add an example for that as well.

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+### Generated Primary Keys
+`Sequences` are not supported. Hence, auto increment primary key is not supported and should be replaced with primary key definitions that
+are manually assigned. See https://cloud.google.com/spanner/docs/schema-design#primary-key-prevent-hotspots
+for more information on choosing a good primary key. This sample uses UUIDs that are generated by the client for primary
+keys.
+
+### Savepoint
+`savepoints` are supported in the Cloud Spanner. Hence, following functions from the `django.db.transaction` are not supported
```
olavloite:
```suggestion
`savepoints` are not supported in Cloud Spanner. The following functions in Django are therefore also not supported:
```

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+transaction.savepoint()
+transaction.savepoint_rollback()
+transaction.savepoint_commit()
+transaction.clean_savepoints()
+```
+
+### Nested Atomic Blocks
+As mentioned earlier also, `savepoints` are not supported and hence, a code like follows might not work with PGAdapter
+because it will make use of savepoint to make sure the inner nested block is rolled back properly 
+if some error occurs over there.
```
olavloite: This section is too much like 'informal spoken language'. Please rephrase to:

```suggestion
`Savepoints` are not supported in Cloud Spanner. Nested atomic blocks in `django` will by default create a
`Savepoint` when an inner block is entered. You can suppress this behavior by adding `savepoint=False` to
the `@transaction.atomic` annotation.
```

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+from django.db import transaction
+
+@transaction.atomic
+def func():
+  do_stuff
+  with transaction.atomic():
+    do_some_more_stuff
+```
+
+Instead of above, always set `savepoint=False` while using nested atomic blocks
```
olavloite: Remove this and only show the sample that works

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+```python
+from django.db import transaction
+
+@transaction.atomic(savepoint=False)
```
olavloite, line 99: Add a code comment here and below to indicate why we are including `savepoint=False`, e.g.:

`# Include 'savepoint=False' to prevent django from creating a savepoint.`

samples/python/django/README.md
```diff
@@ -0,0 +1,118 @@
...
+    do_some_more_stuff
+```
+
+### Interleaved Tables
+A possible workaround for Django's inability to use interleaved table with Django and PGAdapter includes 
+having a column which has unique index in the actual table while pretending to Django like it's a primary key. 
+
+One important thing to note here is that we can't use usual `save` function for creating new records for such a model 
+because that might generate an `UPDATE` statement which will try to update the unique index. 
+So, we will have to use `force_insert = True`, while creating new records. For more details, see the sample application.
```
olavloite: [Interleaved tables](https://cloud.google.com/spanner/docs/schema-and-data-model#primary-keys) in Cloud Spanner require composite primary keys. Django does not support models with a composite primary key. You
can work around these limitations and use an interleaved table with Cloud Spanner and Django as follows:

1. Manually create the interleaved table in your database by executing a `CREATE TABLE` statement. This table will use a composite primary key.
2. Define the model in Python code so that the parent key column of the table is defined as a `ForeignKey` referencing the parent table, and the child key column is defined as the primary key of the table.
3. Add a unique index to the child primary key column. This will ensure that the child primary key part is always globally unique, and can be used to efficiently look up a single row.

Example:

```sql
CREATE TABLE tracks (
    track_id character varying NOT NULL,
    id character varying NOT NULL,
    ...
    PRIMARY KEY(id, track_number)
) INTERLEAVE IN PARENT albums ON DELETE CASCADE;

CREATE UNIQUE INDEX unique_idx_track_id ON tracks(track_id);
```

```python
class Track(BaseModel):
  class Meta():
    db_table = 'tracks'
  track_id = models.CharField(primary_key=True, null=False)
  album = models.ForeignKey(Album, on_delete=models.DO_NOTHING, db_column='id')
```

Note that using the standard `save` function in Django is not supported for interleaved tables, as this function will also try to update the `album` column. This column is part of the primary key definition of the table and cannot be updated. Add a `force_insert=True` to create new child records:

```python
track.save(force_insert=True)
```

See the sample application for more details and a working example.

PR #603, opened 2023-01-20
docs: document pgbench usage

olavloite:
1. Documents the steps that are needed to execute `pgbench` using PGAdapter. `pgbench` can be used with PGAdapter, but with a number of limitations. The initialisation function in `pgbench` will try to insert large amounts of data in a single transaction. These need to be split into multiple smaller transactions in PGAdapter. Also, `pgbench` tries to create tables without primary keys. This is also not supported in Spangres.
2. Adds automatic recognition of `pgbench` connections.
3. Adds a feature that allows recognised clients to return `Notice` messages during startup, for example to return a hint on how they should be used.
4. Adds a feature that allows recognised clients to include hints with error messages, for example to direct users to a specific documentation page.


src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/ErrorResponse.java
```diff
@@ -36,58 +39,83 @@ public class ErrorResponse extends WireOutput {
   private static final byte HINT_FLAG = 'H';
   private static final byte NULL_TERMINATOR = 0;
 
-  private final byte[] severity;
-  private final byte[] errorMessage;
-  private final byte[] errorState;
-  private final byte[] hints;
```
olavloite, line 42: These fields are replaced by a single `PGException` field.

src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
```diff
@@ -67,6 +75,48 @@ public ImmutableList<LocalStatement> getLocalStatements(ConnectionHandler connec
...
+      volatile long lastHintTimestampMillis = 0L;
+
+      @Override
+      public void reset() {
+        lastHintTimestampMillis = 0L;
+      }
+
+      @Override
+      boolean isClient(List<String> orderedParameterKeys, Map<String, String> parameters) {
+        // PGBENCH makes it easy for us, as it sends its own name in the application_name parameter.
```
thiagotnunes, line 89: Handy!

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/ErrorResponse.java
```diff
@@ -36,58 +39,83 @@ public class ErrorResponse extends WireOutput {
...
-              + pgException.getHints().getBytes(StandardCharsets.UTF_8).length
-              + NULL_TERMINATOR_LENGTH;
+    byte[] hints = getHints(pgException, client);
+    if (hints.length > 0) {
+      length += FIELD_IDENTIFIER_LENGTH + hints.length + NULL_TERMINATOR_LENGTH;
     }
     return length;
   }
 
+  static byte[] getSeverity(PGException pgException) {
```
thiagotnunes: nit: should getters be moved into the PGException class? Or is this a class that we do not control

src/test/java/com/google/cloud/spanner/pgadapter/wireoutput/ErrorResponseTest.java
```diff
@@ -102,4 +157,63 @@ public void testSendPayloadWithHint() throws Exception {
...
+            PGException.newBuilder("test message")
+                .setSeverity(Severity.ERROR)
+                .setSQLState(SQLState.RaiseException)
+                .build());
+
+    response.sendPayload();
+
+    assertEquals(
+        String.format(
+            "SERROR\0CP0001\0Mtest message\0H%s\0\0", ClientAutoDetector.PGBENCH_USAGE_HINT),
```
thiagotnunes, line 182: As I understand, this test case includes a few nulls (represented by the `\0`), right? If not could we test that case as well?

thiagotnunes (approved):
LGTM

src/main/java/com/google/cloud/spanner/pgadapter/wireoutput/ErrorResponse.java
olavloite: We do control that class, but these aren't getters, but methods that convert the severity of the `PGException` to the corresponding PG wire-protocol bytes. I've renamed these methods accordingly. (So these methods should stay here, as we don't want to add PG wire-protocol logic to the `PGException` class)

src/test/java/com/google/cloud/spanner/pgadapter/wireoutput/ErrorResponseTest.java
olavloite: The `\0` are not nulls, but markers in the PG wire-protocol that the specific part of the message ends at that position. So instead of sending for example the length of a string, the wire-protocol uses null bytes to indicate the end of a field.

PR #605, opened 2023-01-20
feat: make table and function replacements client-aware

olavloite:
The current pg_catalog replacement strategy is fixed for all clients. This means that if a specific client wants to use a pg_catalog table that is currently not supported by Cloud Spanner, we need to add it for all clients. This can be tricky if we cannot really support the table or function. This change makes the pg_catalog implementation in PGAdapter client-aware. This also means that the initialization of pg_catalog must use a lazy-initialized version of the WellKnownClient, as the client may only be known once we get the first query. The detection of the client based on the first query happens before the query is executed. This means that we can safely assume that the WellKnownClient has been initialized when the backend connection needs pg_catalog the first time.


src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
```diff
@@ -88,7 +88,7 @@ public enum DdlTransactionMode {
   }
 
   private static final Logger logger = Logger.getLogger(OptionsMetadata.class.getName());
-  private static final String DEFAULT_SERVER_VERSION = "14.1";
+  public static final String DEFAULT_SERVER_VERSION = "14.1";
```
thiagotnunes, line 91: I am curious, where did we get this version from (since we are not bound to any specific version in Spangres)

src/main/java/com/google/cloud/spanner/pgadapter/utils/LazyInit.java
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+
+package com.google.cloud.spanner.pgadapter.utils;
+
+import java.util.function.Supplier;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/** A simple not-thread-safe lazy initializer. */
+@NotThreadSafe
+public abstract class LazyInit<T> implements Supplier<T> {
```
thiagotnunes: Do we use guava? If so, we can use `Suppliers.memoize` instead of creating this class

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
olavloite: Specifics shared offline.

src/main/java/com/google/cloud/spanner/pgadapter/utils/LazyInit.java
olavloite: Ah, yes. Thanks for the tip, I was looking around for a standard solution for this in the JDK libs and couldn't find anything, but I didn't think of looking into Guava. Changed.

thiagotnunes (approved):
LGTM

PR #609, opened 2023-01-22
fix: binary copy header should be included in first data message

olavloite:
The PostgreSQL wire protocol requires that the binary copy header is sent before any data. It does however not specify whether it should be in a separate CopyData message, or whether it should be included in the first actual CopyData message. PGAdapter used to send the header in a separate message. This is accepted by most clients. PostgreSQL sends the header as part of the first data message, and some clients (npgsql) assume that this will always be the case.

This fix therefore changes the behavior of PGAdapter so it will always include the header with the first actual CopyData message, instead of sending it in a separate message. This also means that if the copy operation does not actually copy any data at all, the header will be included in the trailer message. This is also consistent with real PostgreSQL.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
@@ -674,8 +674,10 @@ public class OptionsMetadata {
         OPTION_DEBUG_MODE,
         "debug-mode",
         false,
-        "-- ONLY USE FOR DEBUGGING -- This option only intended for debugging. It will "
-            + "instruct the server to keep track of all messages it receives.");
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
```diff
@@ -674,8 +674,10 @@ private CommandLine buildOptions(String[] args) {
         OPTION_DEBUG_MODE,
         "debug-mode",
```
pratickchokhani: Should we consider changing the name of the flag?

pratickchokhani (approved):

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
olavloite: Yes, good point. I've moved that out of this change and will make a separate PR for that.

PR #614, opened 2023-01-25
Changed default values in line with Spanner

pratickchokhani:


src/main/resources/com/google/cloud/spanner/myadapter/session/system_variables.txt
```diff
@@ -12,10 +12,10 @@ lower_case_table_names	INT64	0
 max_allowed_packet	INT64	67108864
 net_write_timeout	INT64	60
 performance_schema	INT64	1
-sql_mode	STRING	ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION
```
olavloite, line 15: I would expect `NO_ZERO_DATE` to be something that should be default enabled for Cloud Spanner. My understanding is that this disallows (or emits a warning for) the date '0000-00-00' (see https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date), which is not a valid date in Cloud Spanner (see https://cloud.google.com/spanner/docs/reference/standard-sql/data-types#date_type).

pratickchokhani: Right. I read it in reverse for some reason. Will revert this.

olavloite:
Also here: Please use [semantic commit messages](https://www.conventionalcommits.org/en/v1.0.0/#summary) as the PR title. This ensures that our release bot picks up the change correctly.

PR #629, opened 2023-02-01
fix: show statements failed in pgx

olavloite:
SHOW variable_name statements failed in pgx, because pgx tries to describe all statements (including SHOW). This failed as these statements do not have any metadata, as they are not returned by the backend.

This fix also adds automatic detection of the pgx driver based on the first Parse message that we receive. pgx uses a very specific name for the prepared statements that it creates that can be used for (relatively) safe detection.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
@@ -754,19 +754,25 @@ public class ConnectionHandler extends Thread {
   }
 
   private void maybeSetApplicationName() {
-    if (this.wellKnownClient != WellKnownClient.UNSPECIFIED
-        && getExtendedQueryProtocolHandler() != null
-        && Strings.isNullOrEmpty(
-            getExtendedQueryProtocolHandler()
-                .getBackendConnection()
-                .getSessionState()
-                .get(null, "application_name")
-                .getSetting())) {
-      getExtendedQueryProtocolHandler()
-          .getBackendConnection()
-          .getSessionState()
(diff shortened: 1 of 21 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -713,23 +724,52 @@ public void setWellKnownClient(WellKnownClient wellKnownClient) {
...
     // Make sure that we only try to detect the client once.
     this.hasDeterminedClientUsingQuery = true;
   }
 
+  private void maybeSetApplicationName() {
+    if (this.wellKnownClient != WellKnownClient.UNSPECIFIED
+        && getExtendedQueryProtocolHandler() != null
+        && Strings.isNullOrEmpty(
+            getExtendedQueryProtocolHandler()
+                .getBackendConnection()
```
thiagotnunes: Wondering if anything can return null in this chain of calls?

thiagotnunes (approved):

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Good point. Just verified that:
- ExtendedQueryProtocolHandler.backendConnection is final and initialised to a non-null in all constructors.
- BackendConnection.sessionState is final and initialised to a non-null value in all constructors.
- SessionState.get(String, String) is guaranteed to return a non-null value or throw an exception in case of an undefined parameter name (`application_name` is a fixed parameter name that should always be known).

I've added an extra safeguard in the form of a try-catch-ignore to account for the theoretical possibility that we allow users to supply their own pg_settings.txt file in the future, and that anyone removes the `application_name` setting from that file.

PR #632, opened 2023-02-04
test: add tests for psycopg3

olavloite:
Adds tests and documentation for `psycopg3`.


.github/workflows/integration.yaml
```diff
@@ -82,17 +82,13 @@ jobs:
         with:
           go-version: '^1.17.7'
       - run: go version
-      - uses: actions/setup-python@v4
```
olavloite, line 85: This is duplicated in this script. Once is enough.

.github/workflows/units.yaml
```diff
@@ -47,13 +48,6 @@ jobs:
           distribution: zulu
           java-version: ${{matrix.java}}
       - run: java -version
-      - uses: actions/setup-python@v4
```
olavloite, line 50: This is duplicated in this script. Having it once is enough.

pratickchokhani (approved):

PR #633, opened 2023-02-05
fix: CopyResponse did not return correct column format

olavloite:
CopyResponse for BINARY copy statements did not return the correct format code for the individual columns. The COPY protocol will always use either only TEXT or only BINARY for each column, and the overall format is given in a separate field. The format per column is a precaution in the protocol to support a mix of TEXT and BINARY in potential future versions of the protocol. Most clients therefore ignore the values for the separate columns in a COPY statement. psycopg3 however does inspect these values and will fail if they contain an invalid value. PGAdapter would return a byte array containing only 1's for the column formats if the BINARY protocol was used. As the individual column formats are interpreted as int2 values, that meant that the client saw a format code value of 257 instead of 1.


thiagotnunes (approved):
LGTM

PR #637, opened 2023-02-06
docs: added some more operations in Django sample application

gauravsnj:


samples/python/django/sample.py
```diff
@@ -6,86 +6,117 @@
...
+                   hours=1),
                  created_at=datetime.datetime.now(pytz.UTC),
                  updated_at=datetime.datetime.now(pytz.UTC))
 
+
 def create_tables():
   file = open('create_data_model.sql', 'r')
   ddl_statements = file.read()
-  print(ddl_statements)
+  # print(ddl_statements)
```
olavloite: nit: remove commented code

samples/python/django/sample.py
```diff
@@ -6,86 +6,117 @@
...
-  track.save(force_insert=True)
-
-  venue = create_sample_venue('1')
-  venue.save()
+  try:
+    singer = create_sample_singer('1')
+    singer.save()
+
+    singer_object = Singer.objects.get(id=singer.id)
+    if singer_object.first_name != singer.first_name or singer_object.last_name != singer.last_name:
```
olavloite, line 117: Here and elsewhere: As this is a sample, I think it would be better to just print the singer properties to standard out, and then check for that output in the test case. The sample should as much as possible be aimed at showing how a user can use Django with PGAdapter, and contain as little test code as possible.

samples/python/django/sample.py
```diff
@@ -6,86 +6,117 @@
...
+
+    album = create_sample_album('1', singer)
+    album.save()
+
+    album_object = Album.objects.get(id=album.id)
+    if album_object.title != album.title:
+      raise Exception('Saving Album Data Failed')
+
+    track = create_sample_track('1', '2', album)
+    track.save(force_insert=True)
```
olavloite, line 131: Explain in a comment why we include `force_insert=True`.

samples/python/django/sample.py
```diff
@@ -139,27 +173,99 @@ def transaction_rollback():
...
 
-  #checking if singer3 is present in the actual table or not
+  # checking if singer3 is present in the actual table or not
   if len(Singer.objects.filter(id='3')) > 0:
-    raise Exception('Transaction Rollback Unsucessful')
+    raise Exception('Transaction Rollback Unsuccessful')
+
+
+def interleaved_table_update():
+  Track.objects.filter(track_number='2', album_id='1').update(sample_rate=0.25)
```
olavloite, line 217: Explain in a comment why we need to do it this way.

samples/python/django/sample.py
```diff
@@ -139,27 +173,99 @@ def transaction_rollback():
...
+  updated_track = Track.objects.get(track_number='2', album_id='1')
+  if updated_track.sample_rate != 0.25:
+    raise Exception('Update on Interleaved Table Failed')
+
+
+def fetch_data():
```
olavloite: nit: rename this from `fetch_data` to something `create_sample_data`.

PR #653, opened 2023-02-12
feat: support ARRAY typed query parameters

olavloite:
Adds support for explicitly typed ARRAY parameters using both text and binary formats. Previously, ARRAY query parameters were only supported using the text format and without explicit typing. This also fixes a formatting/parsing issue of textual array parameter values that contained double quotes. These were not escaped correctly.


thiagotnunes (approved):
LGTM

PR #657, opened 2023-02-15
test: some more tests for django

gauravsnj:


src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoBasicTest.java
```diff
@@ -230,4 +265,24 @@ public void testInsertAllTypesAllNull() throws IOException, InterruptedException
     assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
     assertEquals(sqlInsert, mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0).getSql());
   }
+
+  @Test
+  public void testSelectAllNull() throws IOException, InterruptedException {
```
olavloite, line 276: We should test that null values are correctly mapped for all possible data types, not only for strings. So we should create an ORM model / class that has all data types and then verify that we can select an instance of that model when:
1. All values are non-null
2. All values are null

See for example this test for `gorm`:
1. `gorm` struct for an entity with all data types: https://github.com/GoogleCloudPlatform/pgadapter/blob/8a19695373ed5371e9baea4a5980b10c50736403/src/test/golang/pgadapter_gorm_tests/gorm.go#L65
2. A test that verifies that selecting an instance of that entity with all null values works as expected: https://github.com/GoogleCloudPlatform/pgadapter/blob/8a19695373ed5371e9baea4a5980b10c50736403/src/test/java/com/google/cloud/spanner/pgadapter/golang/GormMockServerTest.java#L333

src/test/python/django/transaction_test.py
```diff
@@ -80,6 +80,20 @@ def test_nested_atomic():
...
+def test_error_during_transaction():
+  try:
+    transaction.set_autocommit(False)
+
+    singer = Singer(singerid=1, firstname='hello', lastname='world')
+    singer2 = Singer(singerid=2, firstname='hello', lastname='python')
+
+    singer.save()
+
+    raise Exception("Some error has occurred")
```
olavloite: We should not raise an exception in the application itself, but simulate an error in Cloud Spanner. The reason for this is that PostgreSQL has a very strict model when it comes to handling database errors in transactions:
1. A database error in a transaction will invalidate that transaction, and all further statements in that transaction will also fail with the error `current transaction is aborted, commands ignored until end of transaction block`.
2. The only way to get out of that state is to roll back the transaction (or roll back to a savepoint in the transaction, but that is not supported in PGAdapter at the moment).

We should therefore verify that PGAdapter in combination with DJango really shows that behavior.

See this example:
1. `gorm` code: https://github.com/GoogleCloudPlatform/pgadapter/blob/8a19695373ed5371e9baea4a5980b10c50736403/src/test/golang/pgadapter_gorm_tests/gorm.go#L527
2. Corresponding Java test case: https://github.com/GoogleCloudPlatform/pgadapter/blob/8a19695373ed5371e9baea4a5980b10c50736403/src/test/java/com/google/cloud/spanner/pgadapter/golang/GormMockServerTest.java#L780

olavloite (changes requested):

src/test/python/django/transaction_test.py
gauravsnj: Okay got it. Changed the code.

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoBasicTest.java
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoBasicTest.java
```diff
@@ -230,4 +262,92 @@ public void testInsertAllTypesAllNull() throws IOException, InterruptedException
...
+            + "col_numeric : 6.626 , "
+            + "col_timestamptz : 2022-02-16 13:18:02.123456+00:00 , "
+            + "col_date : 2022-03-29 , "
+            + "col_varchar : test , "
+            + "col_jsonb : {'key': 'value'} , \n";
+    List<String> options = new ArrayList<>(Arrays.asList("select_all_types"));
+    String actualOutput = executeBasicTests(pgServer.getLocalPort(), host, options);
+    assertEquals(expectedOutput, actualOutput);
+
+    assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
```
olavloite, line 359: nit: could we also have a verification here that the statement is executed using a single-use read-only transaction? So something like `assertTrue(request.getTransaction().hasSingleUse());`

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoBasicTest.java
```diff
@@ -230,4 +262,92 @@ public void testInsertAllTypesAllNull() throws IOException, InterruptedException
     assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
     assertEquals(sqlInsert, mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0).getSql());
```
olavloite: nit: could we add a verification that the statement is executed in auto-commit mode. That is:
1. The execute request includes a BeginTransaction option with a read/write-option=true.
2. There is a commit request on the server.

The reason for the above is that Django is a very big exception to most ORMs. Most ORMs will not use auto-commit by default, but will use read/write transactions that must be committed explicitly by default. We should verify this behavior in order to ensure that our implementation is correct, and to be able to detect potential regressions in the future if we or the Django maintainers make any changes.

olavloite (approved):
LGTM (once we have been able to fix the test failure)

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoBasicTest.java
gauravsnj: done

src/test/java/com/google/cloud/spanner/pgadapter/python/django/DjangoBasicTest.java
gauravsnj: done

PR #664, opened 2023-02-16
fix: Server version to be fetched from options.

pratickchokhani:


olavloite:
nit: use [semantic commit messages](https://www.conventionalcommits.org/en/v1.0.0/#summary) (e.g. `fix: fetch server version from options`) as the title of a PR. Our release bot uses that when creating releases and determining the next version number.

Additional information about the change should be added to the PR description.

src/main/java/com/google/cloud/spanner/myadapter/command/CommandHandler.java
```diff
@@ -49,7 +49,8 @@ public CommandHandler(
     this.sessionState = sessionState;
     this.backendConnection = backendConnection;
 
-    this.serverGreetingsMessage = new ServerGreetingsMessage(connectionMetadata, sessionState);
+    this.serverGreetingsMessage = new ServerGreetingsMessage(connectionMetadata, sessionState,
```
olavloite, line 52: In our main protocol, we have the server version as part of the session state. See https://github.com/GoogleCloudPlatform/pgadapter/blob/68d0b14017c8d474d0a9cd44272eb366091edee4/src/main/java/com/google/cloud/spanner/pgadapter/session/SessionState.java#L378

The advantage of that is that it allows the user to set the server version when a connection is created, or even using a `SET` statement during a connection. Is that something that we should consider here as well?

PR #666, opened 2023-02-16
test: use virtual environments for python tests

olavloite:
Creates separate virtual environments for all Python tests (except Django, as those are currently WIP). This ensures that different Python tests will not have version conflicts with each other. It also makes it easier to run the tests locally, as the test setup includes creating a virtual environment and adding the requirements to that venv.


.github/workflows/integration.yaml
```diff
@@ -86,9 +86,6 @@ jobs:
         with:
           python-version: '3.9'
       - run: pip install -r ./src/test/python/requirements.txt
```
olavloite, line 88: This one is not yet being removed, as it is used by the Django tests. This will be fixed in a following PR.

src/test/java/com/google/cloud/spanner/pgadapter/python/PythonTestUtil.java
```diff
@@ -0,0 +1,72 @@
...
+        directoryName);
+  }
+
+  public static String run(String[] command, String directoryName) throws Exception {
+    File directory = new File(directoryName);
+    ProcessBuilder builder = new ProcessBuilder();
+    builder.command(command);
+    builder.directory(directory);
+    builder.environment().put("VIRTUAL_ENV", directoryName + "/venv");
+    builder
```
olavloite, line 50: Required for local builds on MacOS. Ignored by all other environments.

PR #690, opened 2023-02-19
feat: support arrays in copy operations

olavloite:
Adds support for ARRAY types to COPY operations.


src/main/java/com/google/cloud/spanner/pgadapter/utils/BinaryCopyParser.java
```diff
@@ -282,6 +284,53 @@ public Value getValue(Type type, int columnIndex) {
         case DATE:
           return Value.date(field.data == null ? null : DateParser.toDate(field.data));
         case ARRAY:
+          switch (type.getArrayElementType().getCode()) {
+            case STRING:
+              return Value.stringArray(
+                  field.data == null
+                      ? null
+                      : cast(ArrayParser.binaryArrayToList(field.data, true)));
```
aseering: Nit:  Seems like a lot of copypasta here.  Might it be worth pre-computing

```
                  field.data == null
                      ? null
                      : ArrayParser.binaryArrayToList(field.data, true)
```

so that most of the return statements are just `return Value.{type}Array(cast(valuesList));` for various `{type}`?  This wouldn't preclude a future array type doing something different, though it would result in a bit of wasted compute-work (depending on how clever the Java compiler and JVM are) in that case.

(This would also presumably require updating `cast()` to handle nulls usefully.)

olavloite: Good point, fixed

aseering (approved):
Oops, thought I approved modulo comments last time.  Sorry!

PR #697, opened 2023-03-01
test: test different Liquibase versions

olavloite:


olavloite:
Error for v4.17.2 is not an actual error, but a codecov upload that failed.

PR #700, opened 2023-03-02
fix: add pg_sequence and pg_sequences views

olavloite:
There was no replacement for the pg_sequence and pg_sequences views in pg_catalog. These are added as empty views in this fix.


dcaudell:
Hell yeah. Nice @olavloite !

dcaudell:
Hell yeah. Nice @olavloite !

PR #719, opened 2023-03-13
chore: add pg_enum to list of default table replacements

olavloite:
Adds pg_enum as one of the pg_catalog tables that is replaced by default for all clients.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
@@ -20,7 +20,6 @@ import com.google.cloud.spanner.pgadapter.ConnectionHandler;
 import com.google.cloud.spanner.pgadapter.error.PGException;
 import com.google.cloud.spanner.pgadapter.error.SQLState;
 import com.google.cloud.spanner.pgadapter.statements.PgCatalog.EmptyPgAttribute;
-import com.google.cloud.spanner.pgadapter.statements.PgCatalog.EmptyPgEnum;
 import com.google.cloud.spanner.pgadapter.statements.PgCatalog.PgCatalogTable;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
```diff
@@ -184,16 +183,9 @@ boolean isClient(ParseMessage parseMessage) {
...
           ImmutableMap.of(
-              "pg_catalog.pg_attribute",
-              "pg_attribute",
-              "pg_attribute",
-              "pg_attribute",
-              "pg_catalog.pg_enum",
-              "pg_enum",
-              "pg_enum",
-              "pg_enum");
+              "pg_catalog.pg_attribute", "pg_attribute", "pg_attribute", "pg_attribute");
```
pratickchokhani, line 186: I didn't follow this?

pratickchokhani (approved):

src/main/java/com/google/cloud/spanner/pgadapter/utils/ClientAutoDetector.java
olavloite: This removes `pg_enum` from the specific table translations for `NPGSQL`. So far, NPGSQL (the .NET driver) is the only driver that queries the `pg_enum` table. We had therefore added this translation only for that driver. `SQLAlchemy` however also seems to query this table in some cases, which means that we need it for more clients in the future. It is not probable that we will have built-in support for `enum` data types soon, which means that having a generic replacement for all clients using an empty view is reasonable, instead of only adding this replacement for the clients that we know use it.

Note that removing it from here is just a code cleanup, as leaving it in place would just mean that it would first added as a generic replacement, and then added once more if NPGSQL would be detected as the connected client. These replacements are put in a `Map`, and putting it multiple times with the same value is a no-op.

PR #720, opened 2023-03-13
chore: allow client-detection to skip the first N messages

olavloite:
Some clients might need to skip the first N Parse messages that we receive before doing the actual client detection. This is for example the case if the client first executes a number of very generic statements, like BEGIN or SHOW ISOLATION LEVEL.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
@@ -756,6 +756,13 @@ public class ConnectionHandler extends Thread {
    */
   public void maybeDetermineWellKnownClient(ParseMessage parseMessage) {
     if (!this.hasDeterminedClientUsingQuery) {
+      // We skip up to 10 Parse messages before forcing the connection to detect a client (or not).
+      // This is a safety measure against clients that might send a very large number of Parse
+      // messages with client-side statements directly after connecting. This could in worst case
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -743,14 +756,21 @@ && getServer().getOptions().shouldAutoDetectClient()) {
    */
   public void maybeDetermineWellKnownClient(ParseMessage parseMessage) {
     if (!this.hasDeterminedClientUsingQuery) {
-      if (this.wellKnownClient == WellKnownClient.UNSPECIFIED
-          && getServer().getOptions().shouldAutoDetectClient()) {
-        setWellKnownClient(ClientAutoDetector.detectClient(parseMessage));
+      if (parseMessage.getStatement().getStatementType() == StatementType.CLIENT_SIDE
+          && skippedAutoDetectParseMessages.size() < 10) {
```
pratickchokhani, line 767: How is number 10 decided?

pratickchokhani (approved):

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: A relatively random choice. This is the maximum number of messages that we will buffer before forcing a client detection. This is just a safety measure against clients that might send a lot of random client-side statements at the beginning. I'll add a comment to the code to explain that.

PR #735, opened 2023-03-20
test: update psycopg3 savepoint documentation

olavloite:
Adds tests and documentation for what happens when using a `SAVEPOINT` with `psycopg3`.


olavloite:
Closing as we now have emulated savepoint support

PR #749, opened 2023-03-24
feat: add support for Ruby ActiveRecord

olavloite:
Adds experimental support for Ruby ActiveRecord 7.x. This PR adds:
* Tests using an in-mem mock Spanner server to verify that the requests that we receive from Ruby ActiveRecord.
* Integration tests running against a real Cloud Spanner instance.
* A sample application using all supported data types.
* Documentation and instructions for how to set up an application to use Ruby ActiveRecord with PGAdapter, including known limitations.


olavloite:
@palladius FYI

src/test/java/com/google/cloud/spanner/pgadapter/ruby/RubyTest.java
```diff
@@ -0,0 +1,18 @@
+// Copyright 2022 Google LLC
```
hengfengli: nits: 2022 -> 2023

hengfengli (approved):
Generally, LGTM. Why do we need a mock server just for ruby/active record? Can't we have one for all cases?

olavloite:
> Generally, LGTM. Why do we need a mock server just for ruby/active record? Can't we have one for all cases?

Thanks for the view. We have a specific mock server for Ruby/ActiveRecord, but it is based on the generic mock server that we have for other tests: https://github.com/GoogleCloudPlatform/pgadapter/blob/ef17dba47753b73b5b7eb6d084364e09eb3a504e/src/test/java/com/google/cloud/spanner/pgadapter/ruby/AbstractRubyMockServerTest.java#L47

The reason that we have a specific one for Ruby is that the Ruby driver will execute some specific queries either when a new connection is created, or when specific methods are executed on the driver (e.g. it will check for the existence of specific types the first time those types are used). We keep these specific queries in separate mock servers per driver/ecosystem. We for example also have a similar setup for .NET, as the .NET driver also executes specific queries when some data types are used.

src/test/java/com/google/cloud/spanner/pgadapter/ruby/RubyTest.java
olavloite: Done

PR #797, opened 2023-04-30
feat: support cursors and foreign data wrapper

olavloite:
Add support for DECLARE, FETCH/MOVE and CLOSE cursors. This also enables the use of PGAdapter as a `FOREIGN SERVER` in a real PostgreSQL database, which again makes it possible to use features like user-defined functions, stored procedures or built-in functions that are not supported by Cloud Spanner for data in Cloud Spanner (albeit with limitations).


src/main/java/com/google/cloud/spanner/pgadapter/statements/AbstractFetchOrMoveStatement.java
```diff
@@ -0,0 +1,269 @@
...
+      } else {
+        direction = Direction.FORWARD;
+      }
+    } else if (parser.eatKeyword("backward")) {
+      if (parser.eatKeyword("all")) {
+        direction = Direction.BACKWARD_ALL;
+      } else {
+        direction = Direction.BACKWARD;
+      }
+    } else {
```
pratickchokhani, line 220: Do need to add a validation for ill formed query here?

src/main/java/com/google/cloud/spanner/pgadapter/statements/CloseStatement.java
```diff
@@ -0,0 +1,127 @@
...
+            originalStatement),
+        NO_PARAMS,
+        ImmutableList.of(),
+        ImmutableList.of());
+    this.closeStatement = parse(originalStatement.getSql());
+  }
+
+  @Override
+  public String getCommandTag() {
+    return "CLOSE CURSOR";
```
pratickchokhani, line 68: Spaces in the command tag seems weird. Should we change it to '_'?

src/main/java/com/google/cloud/spanner/pgadapter/statements/SimpleParser.java
```diff
@@ -836,6 +836,53 @@ boolean skipCommentsAndLiterals() {
...
+      pos = originalPos;
+      return true;
+    }
+    return false;
+  }
+
+  boolean skipNumericLiteral() {
+    skipWhitespaces();
+    int startPos = pos;
+    if (eatToken("+") || eatToken("-")) {
```
pratickchokhani, line 864: Is there a possibility of both being present? Or do we need checks for malformed query?

src/main/java/com/google/cloud/spanner/pgadapter/statements/AbstractFetchOrMoveStatement.java
olavloite: No, that handles itself automatically. I've added the following comment to clarify:

```
              // This ensures that the stream will return the first valid direction keyword that it
              // finds, or null if no valid direction keyword is found. If no valid direction
              // keyword is found, then the position of the parser will also not be moved.
```

So basically: If no valid direction keyword is found, then direction is null (which could be valid, as direction is optional). The further parsing of the statement might fail, as it expects for example a count or one of the `from | in` keywords to follow.

src/main/java/com/google/cloud/spanner/pgadapter/statements/CloseStatement.java
olavloite: This is consistent with what PostgreSQL returns, so we should keep it with spaces. This is the text that for example `psql` prints out to the console after executing the command.

src/main/java/com/google/cloud/spanner/pgadapter/statements/SimpleParser.java
olavloite: Other methods that call this method need to check the result (it returns null if it was not a valid integer). The convention in this class is to return null in case the token was not consistent with the thing that the caller wanted to parse, as it makes it easier to implement optional steps, e.g. `fetch 1 from my_cursor` and `fetch from my_cursor` are equivalent.

`+-` is not allowed at the beginning, and this prevents that from being accepted, as this line will only read one of the tokens (either `+` or `-`). The rest of this method then expects the following token to be a digit (or `inf`).

I've added a couple of tests to verify.

pratickchokhani (approved):

PR #841, opened 2023-05-18
feat: support drop cascade

olavloite:
Adds support for the `CASCADE` option for dropping schemas and tables. The feature is conditional on executing `SET spanner.support_drop_cascade=true` (or similar) first.

* `DROP SCHEMA <schema> CASCADE` will drop all objects in the schema, but not the schema itself.
* `DROP TABLE <table> CASCADE` will first drop all dependent objects of the table and then the table itself. This happens in one DDL batch.
* `DROP TABLE <table>` will allow dropping a table, even if it has a secondary index.


src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
```diff
@@ -66,14 +73,52 @@ static String unquoteIdentifier(String identifier) {
...
+    } else {
+      ImmutableList<Statement> allStatements =
+          getDependentStatements(translate(parsedStatement, statement));
+      if (allStatements.size() == 1) {
+        return connection.execute(allStatements.get(0));
+      } else {
+        boolean startedBatch = false;
+        try {
+          if (!connection.isDdlBatchActive()) {
+            connection.execute(Statement.of("START BATCH DDL"));
```
pratickchokhani: Should the string be made static?

src/main/java/com/google/cloud/spanner/pgadapter/session/SessionState.java
```diff
@@ -399,6 +399,14 @@ public boolean isForceAutocommit() {
     return getBoolSetting("spanner", "force_autocommit", false);
   }
 
+  /**
+   * Returns whether DROP ... CASCADE statements are supported. This also includes implicit
+   * cascading operations, such as dropping indexes of tables that are being dropped.
+   */
+  public boolean isSupportDropCascade() {
+    return getBoolSetting("spanner", "support_drop_cascade", false);
```
pratickchokhani, line 407: Static string?

src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
```diff
@@ -66,14 +73,52 @@ static String unquoteIdentifier(String identifier) {
...
+            startedBatch = true;
+          }
+          StatementResult result = null;
+          for (Statement dropDependency : allStatements) {
+            result = connection.execute(dropDependency);
```
pratickchokhani: Should the result be validated?

src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
olavloite: The entire `Statement` class can be made a `static final` field. Done.

src/main/java/com/google/cloud/spanner/pgadapter/session/SessionState.java
olavloite: I would prefer not to do that here. The string is only used in this place in the actual code (excluding test code), and having the literal string here makes it more readable.

An additional reason to prefer not to make this a static field is that it is then also tempting to reference that static field in tests. That again means that if someone changes the value of the static field, then the tests will remain green, while the change itself is actually a breaking change for users. By having this literal string here, and not accessible from any tests, means that anyone making a change to this string is also likely to break tests.

src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
olavloite: Hmm... We don't need to validate it, because we know that it is a `StatementResult.NO_RESULT` in all cases, as the statements are all `DROP` statements that are generated in this class. The reason that it was assigned here, was to guarantee that there is a result that we can return, but that is also not really necessary. Instead, we can just return the static field `NO_RESULT` that we already have.

pratickchokhani (approved):

PR #863, opened 2023-06-02
chore: cleanup client-side IF [NOT] EXISTS checks

olavloite:
Removes the client-side handling of IF [NOT] EXISTS clauses in DDL statements, and instead assumes by default that the server supports this.


src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
```diff
@@ -1313,26 +1312,15 @@ int executeStatementsInBatch(int fromIndex) {
...
         && returnedUpdateCounts[index] == 1
         && successfullyExecutedCount < statementResults.size()) {
-      if (!(statementResults.get(successfullyExecutedCount) instanceof NotExecuted)) {
-        index++;
-      }
-      successfullyExecutedCount++;
-    }
-    while (successfullyExecutedCount < statementResults.size()
-        && statementResults.get(successfullyExecutedCount) instanceof NotExecuted) {
+      index++;
```
pratickchokhani: It seems that both index and successfullyExecutedCount will have the same value. Do we need both the fields?

src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
```diff
@@ -171,180 +52,33 @@ static String unquoteIdentifier(String identifier) {
...
-      return maybeRemovePrimaryKeyConstraintName(createTableStatement);
-    }
-    boolean unique = parser.eatKeyword("unique");
-    if (parser.eatKeyword("index")) {
-      return translateCreateIndex(parser, statement, unique);
-    }
-    return statement;
-  }
-
-  Statement translateDrop(SimpleParser parser, Statement statement) {
```
pratickchokhani, line 216: Just to confirm, we are removing translateDrop because IF exist translation is no longer required?

src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
olavloite: Good point, they are indeed now duplicates of each other, as we will send all statements to the backend. Removed the separate `index` variable.

src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
olavloite: Correct, and we also don't have any other translations for `DROP` statements.

(For comparison: For `CREATE` statements we don't drop the entire translate method, because we still translate `CREATE TABLE` statements that include a named primary key, when that primary key just uses the same name as what Cloud Spanner would otherwise generate)

pratickchokhani (approved):

PR #878, opened 2023-06-10
feat: create local PostgreSQL copy and support pg_dump and pg_restore

olavloite:
Adds a docker-compose.yml file and some scripts that will automatically set up a local copy of a Cloud Spanner database in a PostgreSQL database. The PostgreSQL database is created in a Docker container. The docker-compose file also starts PGAdapter in a Docker container, and connects the real PostgreSQL database to this PGAdapter instance, so that the Cloud Spanner database is imported as a FOREIGN SCHEMA in the PostgreSQL database.

Also adds scripts for automatically copying data between a real PostgreSQL database and Cloud Spanner. These scripts can then be used to support the following:
1. `pg_dump`: The `run_pg_dump.sh` script copies the schema and all data of a Cloud Spanner database to a real PostgreSQL database, and then runs `pg_dump` on the real PostgreSQL database. The script also exports a `schema.sql` Cloud Spanner DDL script and stores it together with the `pg_dump`. The combination of these can be used to re-create both a native PostgreSQL database, as well as the original Cloud Spanner database.
2. `pg_restore`: The `run_pg_restore.sh` script runs `pg_restore` to restore a real PostgreSQL database. It then executes the `schema.sql` file on a Cloud Spanner database to re-create the original Cloud Spanner database, and copies all data from the PostgreSQL database to the Cloud Spanner database.

The above enables users to:
1. Easily create a `pg_dump` that they can use if they ever want to move their Cloud Spanner to a different PostgreSQL-compatible database service.
2. Easily create offline copies of their Cloud Spanner database that they can use to both restore to different Cloud Spanner instances/databases, for example to create standardized development or test databases.


local-postgresql-copy/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+```shell
+docker compose --env-file .env up -d
+```
+
+3. Start an interactive `psql` shell on the local PostgreSQL database. The command below assumes
+   that the `POSTGRES_CONTAINER_NAME` was left unmodified with its default value `postgres` in the `.env` file.
+
+```shell
+docker exec -it postgres /bin/bash -c 'psql -U "${POSTGRES_USER}" -d "${POSTGRES_DB}"'
```
aseering: Minor nit, could this be `/bin/sh` and save a few cycles?

local-postgresql-copy/copy-data.sh
```diff
@@ -0,0 +1,49 @@
...
+
+echo "COPYING DATA FROM CLOUD SPANNER TO POSTGRESQL"
+
+# First truncate all tables.
+PGPASSWORD="$POSTGRES_PASSWORD" psql -v ON_ERROR_STOP=1 -U "$POSTGRES_USER" -h "$POSTGRES_HOST" -p "$POSTGRES_PORT" -d "$POSTGRES_DB" -qAtX \
+  -c "select table_name from information_schema.tables where table_type='BASE TABLE' and table_schema='public';" \
+| while read table_name ; do
+  echo ""
+  echo "Truncating data for $table_name"
+  PGPASSWORD="$POSTGRES_PASSWORD" psql -a -v ON_ERROR_STOP=1 -U "$POSTGRES_USER" -h "$POSTGRES_HOST" -p "$POSTGRES_PORT" -d "$POSTGRES_DB" -c "truncate $table_name cascade"
```
aseering, line 30: Will (or should) this error if the table is missing?  (I forget when psql returns what error codes.)

local-postgresql-copy/copy-data.sh
```diff
@@ -0,0 +1,49 @@
...
+    -c "select array_to_string(array_agg(column_name), ',') as c from information_schema.columns where table_name='$table_name' and table_schema='public' and not is_generated='ALWAYS';")
+
+  echo ""
+  echo "Copying data for $table_name"
+  # Set the read_only_staleness to use in PGOPTIONS. This will automatically be included by psql
+  # when connecting to PGAdapter.
+  # We also set the PostgreSQL session in 'session_replication_mode='replica' to disable all triggers
+  # during the import. Foreign keys are implemented as triggers in PostgreSQL, so this automatically
+  # also disables foreign key constraint checks during the import.
+  PGOPTIONS="-c spanner.read_only_staleness='read_timestamp $READ_TIMESTAMP'" \
```
aseering, line 67: If the customer has a large database to export, by the time this script gets to the oldest tables, might this exceed the oldest supported staleness?

local-postgresql-copy/copy-data.sh
```diff
@@ -0,0 +1,49 @@
...
+    psql -v ON_ERROR_STOP=1 \
+       -h "$PGADAPTER_HOST" \
+       -p "$PGADAPTER_PORT" \
+       -d "$SPANNER_DATABASE" \
+       -c "copy $table_name ($column_names) to stdout binary" \
+  | PGPASSWORD="$POSTGRES_PASSWORD" psql -v ON_ERROR_STOP=1 -a -U "$POSTGRES_USER" -h "$POSTGRES_HOST" -p "$POSTGRES_PORT" -d "$POSTGRES_DB" \
+      -c "set session_replication_role='replica'" \
+      -c "copy $table_name ($column_names) from stdin binary" \
+      -c "set session_replication_role='origin'"
```
aseering: Would there be any advantage to running these COPYs in parallel?  (`&` here and `wait` after `done`, or more sophisticated if you're feeling fancy.)

Spanner ought to be faster with that approach; we scale pretty well.  PostgreSQL might bottleneck?

local-postgresql-copy/run-pg-dump.sh
```diff
@@ -0,0 +1,59 @@
...
+  --port "$PGADAPTER_PORT" \
+  --dbname "$SPANNER_DATABASE" \
+  -qAtX \
+  -c "show database ddl" > "$schema_file"
+
+# Check if we should copy the data from Cloud Spanner to the PostgreSQL database first.
+if [ "$copy_data" = "true" ];
+then
+  echo "Copying schema from Cloud Spanner to the PostgreSQL database $POSTGRES_DB"
+  source ./copy-schema.sh
```
aseering: Does this script set any variables that matter here?  If so, that's implicit / not immediately obvious; worth adding a comment to explain?  If not, should it be `source`'d here, or just called?

local-postgresql-copy/run-pg-restore.sh
```diff
@@ -0,0 +1,210 @@
...
+  -d "$spanner_restore_database" \
+  -c "start batch ddl" \
+  -f "$schema_file" \
+  -c "run batch"
+
+# Re-import the Cloud Spanner database as foreign schemas in the PostgreSQL database.
+# This will allow us to query the Cloud Spanner information_schema using the PostgreSQL query engine,
+# which again allows us to execute a recursive query on the information_schema.tables table in Cloud Spanner.
+echo "Importing the Cloud Spanner schema as a foreign schema in the PostgreSQL database"
+PGPASSWORD="$POSTGRES_PASSWORD" psql -v ON_ERROR_STOP=1 \
```
aseering, line 74: Minor readability nit -- it looks like psql is only invoked in a few (2?) ways?  Would it be more readable to create well-named variables containing the environment/args for each, rather than repeating all of the args each time?

aseering (approved):
This is a lengthy CL and I'm unfortunately quite busy for the next few days so I've only had time for a somewhat cursory review.  That said, LGTM.

local-postgresql-copy/README.md
olavloite: It could, but it would require that the user has `psql` installed on localhost. This has all been set up with the assumption that the user does not need to have any PostgreSQL tools installed locally, and only requires that Docker is available. So it is really trying to be as much 'anyone can get this running without installing anything' as possible. The only requirement is the aforementioned dependency on Docker.

There is an example for how to connect to the PostgreSQL and Cloud Spanner database directly from localhost a bit further down in the document: https://github.com/GoogleCloudPlatform/pgadapter/blob/ba1499aebb6940ee2514607beec322a29341043f/local-postgresql-copy/README.md#connecting-to-the-containers

local-postgresql-copy/copy-data.sh
olavloite: Yes, a `Relation not found` error will cause `psql` to stop and return exit code 1. That is also the intention here. The table should be there, assuming that the user has either followed the instruction of creating the schema first, or is running one of the standardized scripts that will handle it for them.

local-postgresql-copy/copy-data.sh
olavloite: Yes, that is possible, although the COPY implementation in PGAdapter actually uses partitioned queries for tables that are big enough to be split into multiple partitions. Partitioned queries prevent Cloud Spanner from garbage collecting data that is still being read (see https://cloud.google.com/spanner/docs/timestamp-bounds#maximum_timestamp_staleness).

So that, in combination with parallelism (see below) would make it possible to work around that if it becomes a problem.

local-postgresql-copy/copy-data.sh
olavloite: Good point. PostgreSQL is indeed likely to become the bottleneck. In addition, it makes the simple case of 'trying it out with a development database' harder to follow, as running them in parallel will also cause the `COPY` operations to log output in random order. So what I've done is:
1. Introduced a `max_parallelism` variable at the beginning of the script with a default value of 1. Setting a higher value will run at most that many COPY operations in parallel.
2. Added a `wait_for_copy` function that waits for a given `COPY` operation to finish. This is called whenever max_parallelism is reached.

This keeps the standard case relatively simple (no parallelism), while makes it possible for people to increase parallelism as it fits their use case best. Executing all `COPY` operations in parallel is probably not a good idea, as Cloud Spanner supports up to 5,000 tables in a database.

local-postgresql-copy/README.md
aseering: Why would this change require `psql` to be installed?

Just to clarify, I'm not suggesting to replace the `docker exec` with `/bin/sh`, just the `/bin/bash`.

local-postgresql-copy/run-pg-dump.sh
olavloite: No, it does not (anymore). Replaced with just executing the script.

local-postgresql-copy/README.md
olavloite: Ah, I misunderstood that then. Maybe, let me see how that works out with the environment variables in that case.

olavloite: Done

local-postgresql-copy/run-pg-restore.sh
olavloite: I'm going to defer this to a later pull request in order to be able to finish both this and the Spring Data JPA sample before my vacation :-)

olavloite:
Build error is caused by codecov upload failure (i.e. no actual test or build failure).

PR #995, opened 2023-08-24
test: add dialect to Connection URI and support autoConfigEmulator

olavloite:
Automatically add the `PostgreSQL` dialect to the Connection URL. This ensures that the underlying client library will create a `PostgreSQL` dialect database when the `autoConfigEmulator=true` property has also been set in the connection URL.

This pull request also moves the logic to build the Connection URL to a separately testable method.


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -255,7 +230,40 @@ public void connectToSpanner(String database, @Nullable Credentials credentials)
...
-  private String appendPropertiesToUrl(String url, Properties info) {
+  @VisibleForTesting
+  static String buildConnectionURL(
+      String database, OptionsMetadata options, Properties properties) {
+    String uri =
+        options.hasDefaultConnectionUrl()
+            ? options.getDefaultConnectionUrl()
+            : options.buildConnectionURL(database);
+    uri = appendPropertiesToUrl(uri, properties);
+    uri = uri + ";dialect=postgresql";
```
manu2, line 245: I am a bit confused here. Does client library today needs to know during connection set up that it is connecting to Postgres database? If yes, how is this being handled without this option?
And if not, why is this needed for emulator? Is there a difference in the interface?

manu2 (approved):

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Added a comment to the code to explain this. The dialect property is __only__ used in combination with `autoConfigEmulator=true`. The latter will automatically create the instance and database that is set in the connection string, as that makes it easier to spin up an emulator instance and connect to it for testing. The connection then needs to know what dialect that the database it creates should be.

PR #1020, opened 2023-09-01
docs: update Hibernate sample to Hibernate 6.2

olavloite:
Updates the Hibernate sample to use the most recent Hibernate version. This is more relevant, as new users as more likely to use a recent version of Hibernate than an old 5.x version that is no longer supported.


olavloite:
The lower code coverage is a false positive. It is caused by the fact that I recently restricted the number of integration tests that are running for pull requests, in order to keep the test time for PRs within reasonable limits. All integration tests are still executed on nightly builds.

pratickchokhani (approved):

PR #1055, opened 2023-09-15
docs: use Go wrapper for gorm sample

olavloite:
Changes the `gorm` sample application to use the Go wrapper for PGAdapter. This allows us to start PGAdapter directly from the sample application and makes the ramp-up for running a simple sample a lot lower.

It also makes it possible to show that you can run the same sample application on both Cloud Spanner PostgreSQL and open-source PostgreSQL. This update therefore also introduces an extra command-line argument that instructs the sample application to use an open-source PostgreSQL instance instead of Cloud Spanner.


samples/golang/gorm/create_data_model.sql
```diff
@@ -1,11 +1,17 @@
...
 create table if not exists singers (
     id         varchar not null primary key,
     first_name varchar,
     last_name  varchar not null,
-    full_name  varchar generated always as (coalesce(concat(first_name, ' '::varchar, last_name), last_name)) stored,
+    full_name varchar(300) generated always as (
+        CASE WHEN first_name IS NULL THEN last_name
+             WHEN last_name  IS NULL THEN first_name
+             ELSE first_name || ' ' || last_name
+        END) stored,
```
olavloite, line 14: This needed to be changed, as `concat` is not an `IMMUTABLE` function on open-source PostgreSQL, which again means that it is not allowed for a generated column. The `||` operator is considered `IMMUTABLE`.

pratickchokhani (approved):

PR #1181, opened 2023-11-10
build: Docker image with emulator

olavloite:
Adds a pre-built Docker image that contains both PGAdapter and the Cloud Spanner Emulator. This Docker image can be used to quickly start an instance of PGAdapter with the Emulator, and that is pre-configured to use the emulator inside the same Docker container, and to automatically create any Spanner instance and database on the emulator that is referenced in a connection request.

That is, all of the following will work:

```
docker run -d --name pgadapter-emulator -p 5432:5432 gcr.io/cloud-spanner-pg-adapter/pgadapter-emulator
sleep 2
psql -h localhost -p 5432 -d test-database -c "select 'Hello World from the Emulator';"
psql -h localhost -p 5432 -d test-database -c "select current_database()"
psql -h localhost -p 5432 -d some-other-database -c "select current_database()"
psql -h localhost -p 5432 -d projects/some-project/instances/some-instance/databases/some-database -c "select current_database()"
docker container stop pgadapter-emulator
```

The output will be:

```
           ?column?            
-------------------------------
 Hello World from the Emulator
(1 row)

 current_database 
------------------
 test-database
(1 row)

  current_database   
---------------------
 some-other-database
(1 row)

 current_database 
------------------
 some-database
(1 row)
```


README.md
```diff
@@ -177,6 +177,24 @@ See [samples/java/jdbc](samples/java/jdbc) for a small sample application that a
...
 ## Emulator
+A pre-built Docker image that contains both PGAdapter and the Spanner Emulator can be started with
+these commands:
+
+```shell
+docker pull gcr.io/cloud-spanner-pg-adapter/pgadapter-emulator
+docker run \
+  -d -p 5432:5432 \
+  gcr.io/cloud-spanner-pg-adapter/pgadapter-emulator
+sleep 2
```
hengfengli, line 188: Is this necessary?

hengfengli (approved):
LGTM

README.md
olavloite: Yeah, to be sure that users who copy-paste this exact script don't run into anything, then we need something here. The reason is that:

1. PGAdapter needs a bit of time to start up (some ~hundreds of milliseconds).
2. The PostgreSQL TCP port is only available once PGAdapter has started.
3. However, as Docker is started in detached mode in this case, the `docker run` command will finish directly.
4. That again means that the `psql` command following will fail with a somewhat random error if it tries to connect before PGAdapter has started up properly.

One alternative could have been add some kind of polling loop that checks whether port 5432 is available, but this feels 'safer' that it will work on all hosts.

PR #1182, opened 2023-11-12
feat: add OpenTelemetry tracing

olavloite:
Adds `OpenTelemetry` tracing and configures Google Cloud Trace exports.


docs/open_telemetry.md
```diff
@@ -0,0 +1,176 @@
...
+
+PGAdapter uses a different set of default values for the following OpenTelemetry Autoconfigure
+properties:
+- `otel.traces.exporter`: `none`
+- `otel.metrics.exporter`: `none`
+- `otel.logs.exporter`: `none`
+
+In addition, PGAdapter always adds an exporter for Google Cloud Trace.
+
+## Traces
```
KiranmayiB, line 35: Can you please add examples for metrics also here?

src/main/java/com/google/cloud/spanner/pgadapter/Server.java
```diff
@@ -31,13 +45,87 @@ public class Server {
...
+      ProxyServer server = new ProxyServer(optionsMetadata, openTelemetry);
       server.startServer();
     } catch (Exception e) {
       printError(e, System.err, System.out);
     }
   }
 
+  static OpenTelemetry setupOpenTelemetry(OptionsMetadata optionsMetadata) throws IOException {
+    if (!optionsMetadata.isEnableOpenTelemetry()) {
+      return OpenTelemetry.noop();
```
KiranmayiB, line 58: Why is this noop?

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
```diff
@@ -218,6 +222,20 @@ public Builder setRequireAuthentication() {
       return this;
     }
 
+    /** Enables OpenTelemetry tracing for PGAdapter. */
+    public Builder setEnableOpenTelemetry() {
```
KiranmayiB, line 226: We are planning to make the Otel metrics in the client library as opt-in by default. So i think we should make this also as opt-out.

KiranmayiB (commented):
Can you please link me to the guide you used for Otel integration?
It looks very different from https://opentelemetry.io/docs/instrumentation/java/manual/

olavloite:
> https://opentelemetry.io/docs/instrumentation/java/manual/

This setup uses the OpenTelemetry automatic configuration module: https://opentelemetry.io/docs/instrumentation/java/manual/#automatic-configuration

The reason for this is that PGAdapter is mostly used as a standalone application, and not as a library. That means that configuration must happen on the command line when starting the application, instead of programmatic configuration that would be logical in a library (e.g. the Java client). The reasoning here is:
1. We need command line configuration, as PGAdapter is mostly used as a standalone application, and not as an in-process dependency (which by the way is also possible, but less frequently used).
2. Instead of re-inventing the wheel and adding a lot of command line arguments to PGAdapter, we use the library offered by OpenTelemetry to do command line configuration (in this case through Java System properties and/or Environment variables).
3. We override the defaults of OpenTelemetry if the user has not specified any exporters. The default in the automatic configuration module of OpenTelemetry is to use the `oltp` exporter (see https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#exporters). That requires an OpenTelemetry endpoint, which we by default do not have. Instead, we set the default exporter that is used by the auto-configuration module to `none`, but allow the user to specify one using a system property or environment variable.
4. In addition, we by default add a Google Trace exporter if OpenTelemetry has been enabled.

src/main/java/com/google/cloud/spanner/pgadapter/Server.java
olavloite: Returning a `noop` OpenTelemetry instance instead of null when OpenTelemetry has not been enabled makes the rest of the code simpler, as we don't need null checks everywhere.

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
olavloite: What exactly do you mean by `opt-in by default`. Does that mean that you will enable it by default? If so, what exactly do you mean with enabled by default?
1. Does it mean 'we enable OpenTelemetry by default, but we don't by default add any exporters'? If so, I would argue that this is still the same as the feature being an opt-in, as you still need to add an exporter before you actually collect any metrics.
2. Or does it mean that we enable both OpenTelemetry and add an exporter (e.g. Google Cloud Tracer) by default? That would mean that we by default enable a feature that could incur additional costs for the customer. Is that intentional?

docs/open_telemetry.md
olavloite: We don't add any metrics in this pull request, only traces.

KiranmayiB: Alright sounds good then. Thanks!

src/main/java/com/google/cloud/spanner/pgadapter/Server.java
KiranmayiB: Thanks!

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
KiranmayiB: I meant the latter. But I realize now that itsn not possible for traces as we cannot specify a namespace.  So lets leave this as-is. Thank you!

KiranmayiB (approved):
Otel integration LGTM. Thanks a lot!

PR #1266, opened 2023-12-14
feat: translate integrity constraint error codes

olavloite:
Translate integrity constraint errors and return the correct PostgreSQL error code.

Cloud Spanner returns its own error codes for PostgreSQL-dialect databases. These are by default returned by PGAdapter to the client as a generic PostgreSQL `P0001` (`raise_exception`) error. This means that PostgreSQL clients that have built-in logic that depend on specific PostgreSQL error codes will not work correctly with PGAdapter.

PGAdapter does translate a small set of specific Cloud Spanner errors to PostgreSQL error codes. Cloud Spanner `Aborted` errors are for example translated to PostgreSQL's `40001` (`serialization_failure`) error codes. This is the error code that PostgreSQL returns when isolation level `Serializable` is used, and the transaction failed due to a lock conflict.

This pull request adds translations for a couple of additional error codes that are known to be used by some Node.js frameworks to determine the type of error that is returned to the user. The translations that are added in this pull request are:
- Primary key violations: Translated to PostgreSQL `23505`
- Unique key violations: Translated to PostgreSQL ` 23505`
- Foreign key violations: Translated to PostgreSQL `23503`

The translations are done based on the error code and the error message that is returned by Cloud Spanner. Cloud Spanner does not return enough error details to do this translation without also matching on the error message.

We also need separate matches for Cloud Spanner and the Emulator, as these return different error messages.

The full list of error codes that is used by PostgreSQL can be found here: https://www.postgresql.org/docs/current/errcodes-appendix.html


src/main/java/com/google/cloud/spanner/pgadapter/error/SQLState.java
```diff
@@ -135,6 +135,15 @@ public enum SQLState {
   SqlJsonScalarRequired("2203F"),
   SqlJsonItemCannotBeCastToTargetType("2203G"),
 
+  // Class 23 - Integrity Constraint Violation
```
olavloite, line 138: See https://www.postgresql.org/docs/current/errcodes-appendix.html

src/main/java/com/google/cloud/spanner/pgadapter/error/PGExceptionFactory.java
```diff
@@ -27,10 +27,26 @@
 public class PGExceptionFactory {
   private static final Pattern RELATION_NOT_FOUND_PATTERN =
       Pattern.compile("relation .+ does not exist");
+  private static final Pattern COLUMN_NOT_FOUND_PATTERN =
+      Pattern.compile("column .+ of relation .+ does not exist");
   private static final Pattern CANNOT_DROP_TABLE_WITH_INDICES_PATTERN =
-      Pattern.compile("Cannot drop table test with indices");
```
olavloite, line 31: fixed copy-paste error from previous error code translations

src/test/java/com/google/cloud/spanner/pgadapter/ITErrorsTest.java
```diff
@@ -0,0 +1,172 @@
...
+      assertEquals(SQLState.UniqueViolation.toString(), exception.getSQLState());
+    }
+  }
+
+  @Test
+  public void testUniqueIndexViolation() throws SQLException {
+    try (Connection connection = DriverManager.getConnection(getConnectionUrl())) {
+      assertEquals(
+          1,
+          connection
```
ankiaga, line 108: We are copying this assertion across tests. Can we create a separate happy test for this?

src/test/java/com/google/cloud/spanner/pgadapter/ITErrorsTest.java
```diff
@@ -0,0 +1,172 @@
...
+              .createStatement()
+              .executeUpdate("insert into my_table (id, value) values (1, 'One')"));
+      PSQLException exception =
+          assertThrows(
+              PSQLException.class,
+              () ->
+                  connection
+                      .createStatement()
+                      .executeUpdate("insert into my_table (id, value) values (1, 'Other one')"));
```
ankiaga: Can we create a private method and pass 2 things from each test - sql query and the SQLstate enum as thats the only difference in all tests or better write a parametrized test?

olavloite: Added a separate private method to verify the expected SQL state.

We use Java 8 with JUnit4, so support for parameterized tests is very limited (it supports parameterized test classes, but not parameterized tests).

src/test/java/com/google/cloud/spanner/pgadapter/ITErrorsTest.java
olavloite: See below

ankiaga (approved):

PR #1302, opened 2023-12-25
fix: use array lower bound 1 by default

### LINKED ISSUE #1259: Arrays encoded using the binary protocol are always sent with lower bound 0
Arrays that are encoded using the binary format are always sent with lower bound zero:

https://github.com/GoogleCloudPlatform/pgadapter/blob/b55f92439036d319b35972ff668a954a8fc4a335/src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java#L323

The default in PostgreSQL is to use 1 as the lower bound of an array.

olavloite:
PGAdapter would return arrays in binary format with lower bound 0 (zero) by default. This is not consistent with open-source PostgreSQL, which uses 1 as the default lower bound, and also not consistent with the text format, which does not have a field for returning the lower bound. The lower bound is then assumed to be 1. This changes the default lower bound for binary encoded arrays to 1.

Fixes #1259


ankiaga (approved):

jackpertschuk:
thanks!

PR #1465, opened 2024-02-26
chore: support OpenTelemetry metrics

hengfengli:
Add two latency metrics: 

* `spanner/pgadapter/roundtrip_latencies`: measure the roundtrip latency in PGAdapter that can include multiple messages between the client and PGAdapter. 
* `spanner/pgadapter/client_lib_latencies`: measure the time to make a request and get a response from the client library.

### DIFF
--- a/docs/open_telemetry.md
+++ b/docs/open_telemetry.md
@@ -1,7 +1,8 @@
-# Open Telemetry in PGAdapter
+# OpenTelemetry in PGAdapter
 
-PGAdapter supports Open Telemetry tracing. You can enable this by adding the `-enable_otel` command
-line argument when starting PGAdapter.
+PGAdapter supports OpenTelemetry tracing and metrics. You can enable the tracing or the metrics
+by adding the `-enable_otel` or `-enable_otel_metrics` command line argument when starting
+PGAdapter. Tracing and metrics can be enabled at the same time.
 
 Optionally, you can also set a trace sample ratio to limit the number of traces that will be
 collected and set with the `-otel_trace_ratio=<ratio>` command line argument. If you omit this
@@ -16,7 +17,8 @@ docker run \
   gcr.io/cloud-spanner-pg-adapter/pgadapter \
   -p my-project -i my-instance \
   -c /credentials.json -x \
-  -enable_otel -otel_trace_ratio=0.1
+  -enable_otel -otel_trace_ratio=0.1 \
+  -enable_otel_metrics
 ```
 
 ## Exporter
@@ -143,6 +145,15 @@ statements consists of the following spans:
    execute this batch request on Cloud Spanner.
 
 ![PGAdapter Cloud Trace - Batch DML example](img/dml_batch_trace_sample.png?raw=true "PGAdapter Cloud Trace - Batch DML example")
+
+## Metrics
+
+The available metrics in PGAdapter are:
+
+* `spanner/pgadapter/roundtrip_latencies`: Latency between PGAdapter receiving a statement from
+   the client and PGAdapter returning the last row of the response to the client.
+* `spanner/pgadapter/client_lib_latencies`: Latency when the Spanner's client library receives
+   a call and returns a response.
 
 ## Frequently Asked Questions
 
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
     <excludedIntegrationTests/>
 
     <junixsocket.version>2.8.3</junixsocket.version>
-    <opentelemetry.exporter.version>0.27.0</opentelemetry.exporter.version>
+    <opentelemetry.exporter.version>0.23.0</opentelemetry.exporter.version>
   </properties>
 
   <modelVersion>4.0.0</modelVersion>
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,7 @@
     <excludedIntegrationTests/>
 
     <junixsocket.version>2.8.3</junixsocket.version>
+    <opentelemetry.exporter.version>0.27.0</opentelemetry.exporter.version>
   </properties>
 
   <modelVersion>4.0.0</modelVersion>
@@ -141,7 +142,12 @@
     <dependency>
       <groupId>com.google.cloud.opentelemetry</groupId>
       <artifactId>exporter-trace</artifactId>
-      <version>0.23.0</version>
+      <version>${opentelemetry.exporter.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.opentelemetry</groupId>
+      <artifactId>exporter-metrics</artifactId>
+      <version>${opentelemetry.exporter.version}</version>
     </dependency>
     <dependency>
       <groupId>org.jetbrains.kotlin</groupId>
--- a/src/main/java/com/google/cloud/spanner/pgadapter/Server.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/Server.java
@@ -25,6 +25,7 @@ import com.google.devtools.cloudtrace.v2.AttributeValue;
 import com.google.devtools.cloudtrace.v2.TruncatableString;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdkBuilder;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
 import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader;
 import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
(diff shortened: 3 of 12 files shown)


pom.xml
```diff
@@ -37,6 +37,7 @@
     <excludedIntegrationTests/>
 
     <junixsocket.version>2.9.0</junixsocket.version>
+    <opentelemetry.exporter.version>0.27.0</opentelemetry.exporter.version>
```
olavloite: There's unfortunately a bug in the documentation (or implementation?) of this library. It claims to support Java 8, but it actually doesn't. That is why the tests are currently failing for all Java 8 builds. That is the reason that PGAdapter is using an older version of the library.

src/main/java/com/google/cloud/spanner/pgadapter/statements/ExtendedQueryProtocolHandler.java
```diff
@@ -98,6 +110,16 @@ public BackendConnection getBackendConnection() {
     return backendConnection;
   }
 
+  public static Attributes getMetricAttributes(DatabaseId databaseId) {
+    AttributesBuilder attributesBuilder = Attributes.builder();
+    if (databaseId != null) {
```
olavloite: `databaseId` should never be null (at least not for a real connection, maybe in some tests it is). Could we therefore change this into a more direct implementation, where:
1. `metricAttributes` are always created when `ExtendedQueryProtocolHandler` is created.
2. This method is not static, but an instance method, and always just returns the actual `metricAttributes` of this `ExtendedQueryProtocolHandler`. It should also not take `DatabaseId` (or anything else) as an input argument, as the `DatabaseId` is fixed for an `ExtendedQueryProtocolHandler`. The structure here is that:
1. Each connection from a client to PGAdapter has one `ConnectionHandler`.
2. There is a one-to-one relationship between `ConnectionHandler` and `ExtendedQueryProtocolHandler`.
3. There is a one-to-one relationship between `ExtendedQueryProtocolHandler` and `BackendConnection`.
4. All of the above classes also have exactly one `DatabaseId` that never changes during their lifetime.

Now it seems like `metricAttributes` sometimes needs to be created on-the-fly, but I don't think that is ever true.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
```diff
@@ -0,0 +1,75 @@
...
+  @InternalApi
+  public void recordClientLibLatency(long value, Attributes attributes) {
+    if (spannerClientLibLatencies != null) {
+      spannerClientLibLatencies.record(value, attributes);
+    }
+  }
+
+  @InternalApi
+  public void recordPGAdapterLatency(long value, Attributes attributes) {
+    if (pgadapterLatencies != null) {
```
olavloite: Is this ever null?

And if it isn't (which based on the above code I would guess that it never is), could we have some other flag that indicates that a quick return is ok. For example some kind of flag that says 'metrics are disabled'.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
```diff
@@ -0,0 +1,75 @@
...
+
+@InternalApi
+public class Metrics {
+  static final String INSTRUMENTATION_SCOPE = "cloud.google.com/java";
+  static final String SPANNER_CLIENT_LIB_LATENCY = "spanner/pgadapter/client_lib_latencies";
+  static final String SPANNER_CLIENT_LIB_LATENCY_DESCRIPTION =
+      "Latency when the client library receives a call and returns a response";
+  static final String PGADAPTER_LATENCY = "spanner/pgadapter/roundtrip_latencies";
+  static final String PGADAPTER_LATENCY_DESCRIPTION =
+      "Latency when the JDBC makes a call and gets a response, measured in PGAdapter";
```
olavloite: Note that the client that connects to PGAdapter can be any PostgreSQL client, not just JDBC. So we should not talk about JDBC as being 'the client'. My suggestion would be:

```suggestion
      "Latency between PGAdapter receiving a statement from the client and PGAdapter returning the last row of the response to the client";
```

pom.xml
hengfengli: Okay. Let me to use an older version then.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
hengfengli: I got this from the [java client lib](https://github.com/googleapis/java-spanner/blob/4f7d28bdbabce5faf6d37b8a3224235bb1012f78/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerRpcMetrics.java#L61-L73). 

I will remove these null checks and add the support of a flag to enable/disable metrics.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
hengfengli: Updated.

src/main/java/com/google/cloud/spanner/pgadapter/statements/ExtendedQueryProtocolHandler.java
hengfengli: Yes, I was originally creating `metricAttributes` in `ExtendedQueryProtocolHandler` and pass it to the `BackendConnection`. But I thought we already pass `databaseId` into it in [here](https://github.com/GoogleCloudPlatform/pgadapter/blob/8f7dcb99bbce4334e5c4ffd56f7d3aaeaf78acc3/src/main/java/com/google/cloud/spanner/pgadapter/statements/ExtendedQueryProtocolHandler.java#L71). I need to construct the attributes in `Execute` of `BackendConnection`.

hengfengli: I also think `databaseId` should not be null but I add this check due to an error. If I add `metricAttributes` as an instance method, can the `Execute` of `BackendConnection` access it? Let me explore it more.

hengfengli:
@olavloite , currently, `client_lib_latencies` only covers `execute_on_spanner` but `execute_batch_on_spanner` is not included. Do we want to include `execute_batch_on_spanner` in `client_lib_latencies` as well?

src/main/java/com/google/cloud/spanner/pgadapter/statements/ExtendedQueryProtocolHandler.java
hengfengli: Please take a look my latest changes. Basically, I need the attributes in `Execute` of `BackendConnection`.

docs/open_telemetry.md
```diff
@@ -1,7 +1,8 @@
-# Open Telemetry in PGAdapter
+# OpenTelemetry in PGAdapter
 
-PGAdapter supports Open Telemetry tracing. You can enable this by adding the `-enable_otel` command
-line argument when starting PGAdapter.
+PGAdapter supports OpenTelemetry tracing and metrics. You can enable the tracing or the metrics
```
olavloite, line 3: nit:

```suggestion
PGAdapter supports OpenTelemetry tracing and metrics. You can enable tracing or metrics
```

docs/open_telemetry.md
```diff
@@ -144,6 +146,15 @@ statements consists of the following spans:
...
 ![PGAdapter Cloud Trace - Batch DML example](img/dml_batch_trace_sample.png?raw=true "PGAdapter Cloud Trace - Batch DML example")
 
+## Metrics
+
+The available metrics in PGAdapter are:
+
+* `spanner/pgadapter/roundtrip_latencies`: Latency between PGAdapter receiving a statement from
+   the client and PGAdapter returning the last row of the response to the client.
+* `spanner/pgadapter/client_lib_latencies`: Latency when the Spanner's client library receives
+   a call and returns a response.
```
olavloite, line 156:
```suggestion
* `spanner/pgadapter/client_lib_latencies`: Latency between the Spanner client library receiving
   a request and returning a response.
```

src/main/java/com/google/cloud/spanner/pgadapter/Server.java
```diff
@@ -73,45 +79,62 @@ static OpenTelemetry setupOpenTelemetry(OptionsMetadata optionsMetadata) {
...
-      } else {
-        sampler =
-            Sampler.parentBased(
-                Sampler.traceIdRatioBased(optionsMetadata.getOpenTelemetryTraceRatio()));
+      if (optionsMetadata.isEnableOpenTelemetryMetrics()) {
+        MetricExporter cloudMonitoringExporter =
+            GoogleCloudMetricExporter.createWithConfiguration(
+                MetricConfiguration.builder()
+                    // Configure the cloud project id.
+                    .setProjectId(projectId)
```
olavloite, line 128: This needs a null check, as `projectId` can be null, and `.setProject` will throw a NullPointerException if the project is null.

`projectId` is null if PGAdapter has been started without a specific project on the command line.

olavloite:
@hengfengli I took the liberty of doing a small refactor of the code in order to simplify things a bit. I think it is better if we only keep the tracer and metrics in one place (BackendConnection). In addition, we do not need to create a new `Metrics` instance for each connection, as it does not have any connection-specific state. Instead, one instance per `ProxyServer` is enough.

olavloite:
> @olavloite , currently, `client_lib_latencies` only covers `execute_on_spanner` but `execute_batch_on_spanner` is not included. Do we want to include `execute_batch_on_spanner` in `client_lib_latencies` as well?

Yes, that should be included as well. `execute_batch_on_spanner` is the same as `ExecuteBatchDml` (and `UpdateDatabaseDdlRequest` with multiple DDL statements, but that is less relevant here).

olavloite (approved):

PR #1481, opened 2024-03-04
feat: support float4 data type

olavloite:
Adds support for the `float4` / `real` data type.

Integration tests use `float8` on environments that do not yet support `float4`.


src/main/java/com/google/cloud/spanner/pgadapter/parsers/ArrayParser.java
```diff
@@ -212,8 +214,6 @@ private static Object toValidSpannerElement(@Nonnull Object value, int elementOi
         return ((Short) value).longValue();
       case Oid.INT4:
         return ((Integer) value).longValue();
-      case Oid.FLOAT4:
```
olavloite, line 215: This is being removed now, as these are data types that needed special handling because:
1. We allow PostgreSQL clients to send us `int2[]`, `int4[]` and `float4[]` arrays.
2. These are converted to `int8[]` and `float8[]` before being sent to Cloud Spanner.
3. This conversion is no longer needed for `float4` when Cloud Spanner supports `float4` natively.

src/main/java/com/google/cloud/spanner/pgadapter/statements/PgCatalog.java
```diff
@@ -384,7 +384,6 @@ public class PgType implements PgCatalogTable {
             + "  || (case typname\n"
             + "        when 'int2' then 'false'\n"
             + "        when 'int4' then 'false'\n"
-            + "        when 'float4' then 'false'\n"
```
olavloite, line 387: This is the 'typisdefined' column in `pg_type`. Up until now we listed `float4` as a 'type that we know, but that is not implemented'.

src/main/java/com/google/cloud/spanner/pgadapter/utils/MutationWriter.java
```diff
@@ -257,10 +257,9 @@ public StatementResult call() throws Exception {
         }
         if (record.numColumns() != this.tableColumns.keySet().size()) {
           throw PGExceptionFactory.newPGException(
-              "Invalid COPY data: Row length mismatched. Expected "
-                  + this.tableColumns.keySet().size()
-                  + " columns, but only found "
-                  + record.numColumns(),
+              String.format(
```
olavloite, line 260: This change is not directly related to the addition of `float4`, but it popped up during the development that the error message was a bit weird. The error is thrown when the number of columns in the input of the COPY operation is different from the expected number of columns. The error message 'assumed' that the number of input columns was always smaller than the expected number of values (see the '... but only found' part).

arawind (approved):
LGTM, thanks!

PR #1561, opened 2024-03-19
feat: support client lib's OpenTelemetry metrics

hengfengli:
This includes the following changes: 

* When enabling OpenTelemetry metrics in PGAdapter, we also enable the client lib's OT metrics. 
* The builder for OptionsMetadata should support the configuration to enable OT metrics.


olavloite (approved):

olavloite:
(I changed the PR title to `feat:`, as I think this should be considered a feature, and deserves to be included in the release notes.)

PR #1629, opened 2024-04-18
fix: add a random service name for OpenTelemetry

olavloite:
Adds a random number after the standard pgadapter service name used for OpenTelemetry, and adds this to the Metrics exporter. This prevents the `One or more TimeSeries could not be written: Points must be written in order.` error.


hengfengli (approved):
LGTM

PR #1699, opened 2024-05-01
chore: support TPCC benchmark

hengfengli:
This adds the support to run TPCC benchmarking with the following runners: 

- PGAdapter + PG JDBC
- Spanner JDBC

In addition, we enabled the client library's metrics and added a OpenTelemetry metric in the benchmark runner to record roundtrip latencies.


benchmarks/tpcc/README.md
```diff
@@ -0,0 +1,87 @@
...
+mvn spring-boot:run -Dspring-boot.run.arguments="
+  --tpcc.benchmark-duration=PT600s
+  --tpcc.warehouses=10
+  --tpcc.benchmark-threads=1
+  --tpcc.load-data=true
+  --tpcc.truncate-before-load=false
+  --tpcc.run-benchmark=false
+  --tpcc.use-read-only-transactions=false
+  --tpcc.lock-scanned-ranges=false
+  --spanner.project=appdev-soda-spanner-staging
```
olavloite: nit: maybe change these to placeholder arguments

olavloite: (Also for the values below)

benchmarks/tpcc/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>42.6.0</version>
+    </dependency>
+    <!-- Add the Cloud Spanner JDBC driver -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-spanner-jdbc</artifactId>
+      <version>2.17.0</version>
```
olavloite: nit: it should not be necessary to add any version number here, as we are already importing the libraries-bom

benchmarks/tpcc/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    </dependency>
+    <!-- Add PGAdapter as a dependency, so we can start it in-process -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-spanner-pgadapter</artifactId>
+      <version>0.33.1-SNAPSHOT</version>
```
olavloite: I don't think we want to use a snapshot version here

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
```diff
@@ -0,0 +1,762 @@
...
+            stockLevel(statement);
+            statistics.incStockLevel();
+          } else {
+            LOG.info("No transaction");
+          }
+          if (Thread.interrupted()) {
+            break;
+          }
+        } catch (Throwable exception) {
+          if ((exception instanceof PSQLException psqlException)
```
olavloite, line 104: This will not work when running on the Spanner JDBC driver. Is that intentional?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
```diff
@@ -0,0 +1,762 @@
...
+
+    Object[] row;
+    execute(statement, "begin transaction");
+    row =
+        queryRow(
+            statement,
+            (tpccConfiguration.isLockScannedRanges() ? "/*@ lock_scanned_ranges=exclusive */" : "")
+                + "SELECT c_discount, c_last, c_credit, w_tax "
+                + "FROM customer c, warehouse w "
+                + "WHERE w.w_id = %d AND c.w_id = w.w_id AND c.d_id = %d AND c.c_id = %d",
```
olavloite: All these queries appear to use String.format(..) to add literals to the statements. The correct way to do that in JDBC would be to use a `java.sql.PreparedStatement` and add it as query parameters. This will otherwise cause Spanner to have to re-parse the query every time, as the literals change for each execution.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  @Override
+  public void run(String... args) throws Exception {
+    ProxyServer server = pgAdapterConfiguration.isInProcess() ? startPGAdapter() : null;
+    String pgadapterConnectionUrl =
+        server == null
+            ? pgAdapterConfiguration.getConnectionUrl()
+            : String.format(
+                "jdbc:postgresql://localhost:%d/tpcc?preferQueryMode=simple",
```
olavloite: I don't think we should use `preferQueryMode=simple`. That will just slow things down, especially if the benchmark is modified to also use `java.sql.PreparedStatement`.

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
```diff
@@ -230,6 +231,12 @@ public Builder setCredentials(Credentials credentials) {
       return this;
     }
 
+    /** Disables internal retries of aborted read/write transactions. */
+    public Builder setDisableInternalRetries() {
```
olavloite: We should remove this from this pull request. If it is something that we want for this benchmark, or in general, then it should be in a separate PR, so that this PR only contains changes related to the actual benchmark.

src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
```diff
@@ -261,15 +261,17 @@ private final class Execute extends BufferedStatement<StatementResult> {
...
+        DatabaseId databaseId) {
+      this(command, parsedStatement, statement, statementBinder, databaseId, false);
     }
 
     Execute(
         String command,
         ParsedStatement parsedStatement,
         Statement statement,
         Function<Statement, Statement> statementBinder,
+        DatabaseId databaseId,
```
olavloite: This change should also either be removed, or moved to a separate PR (but it does not appear to do anything, so removing is probably the best)

olavloite (commented):
I have not looked into the actual TPC-C implementation, but I assume that it is still the same as the original. The main question that I have is 'Can we use `java.sql.PreparedStatement` (that works for both JDBC drivers) and remove the use of the simple query mode for the PostgreSQL JDBC driver.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
hengfengli: Removed.

src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
hengfengli: Removed.

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
hengfengli: Removed.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
hengfengli: I got this code from https://github.com/GoogleCloudPlatform/pgadapter/blob/1e8b1bf6d5c66c4264c4c780d9bad613e65b371a/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkRunner.java#L54-L98. I did not change it. Do we have a corresponding exception for the Spanner JDBC driver?

benchmarks/tpcc/pom.xml
hengfengli: Fixed.

benchmarks/tpcc/pom.xml
hengfengli: Removed the version number.

benchmarks/tpcc/README.md
hengfengli: Updated.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
olavloite: This exception is the specific exception that the Spanner JDBC driver will throw if a transaction is aborted: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/e3be3915ecbd73d2c918ee144296152adbd55ef6/spanner/jdbc/src/main/java/com/example/spanner/jdbc/TransactionWithRetryLoopExample.java#L81

More in general, all SQLExceptions that are thrown by the Spanner JDBC driver implement `com.google.cloud.spanner.jdbc.JdbcSqlException`

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
hengfengli: Updated to use `java.sql.PreparedStatement`.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
hengfengli: Updated to also handle with `JdbcAbortedException`.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
```diff
@@ -0,0 +1,781 @@
...
+              customerId = resultSet.getLong(1);
+            }
+          }
+        }
+      }
+    }
+    row =
+        paramQueryRow(
+            connection,
+            (tpccConfiguration.isLockScannedRanges() ? "/*@ lock_scanned_ranges=exclusive */" : "")
```
olavloite: nit: we don't seem to add the `lock_scanned_ranges` hint to all queries. E.g. the query on line 368 does not add the hint. Is that intentional?
Would it otherwise make sense to move the logic for conditionally adding the hint to a query/DML statement into the `paramQueryRow` and similar methods, so it is easier to maintain and make sure all queries use the hint when the option has been enabled?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
```diff
@@ -0,0 +1,781 @@
...
+            "DELETE "
+                + "FROM new_orders "
+                + "WHERE o_id = ? AND c_id = ? AND d_id = ? AND w_id = ?",
+            new Object[] {newOrderId, customerId, districtId, warehouseId});
+        row =
+            paramQueryRow(
+                connection,
+                "SELECT c_id FROM orders WHERE o_id = ? AND d_id = ? AND w_id = ?",
+                new Object[] {newOrderId, districtId, warehouseId});
+        // long customerId = (long) row[0];
```
olavloite: Is this intentional?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
```diff
@@ -0,0 +1,241 @@
...
+        String.format(
+            "jdbc:cloudspanner:/projects/%s/instances/%s/databases/%s?numChannels=%d;minSessions=%d;maxSessions=%d"
+                + (pgAdapterConfiguration.getCredentials() == null
+                    ? ""
+                    : ";credentials=" + pgAdapterConfiguration.getCredentials()),
+            spannerConfiguration.getProject(),
+            spannerConfiguration.getInstance(),
+            spannerConfiguration.getDatabase(),
+            pgAdapterConfiguration.getNumChannels(),
+            Math.max(100, tpccConfiguration.getBenchmarkThreads()),
```
olavloite: Would it not make more sense to keep this consistent with the PGAdapter configuration? So either use the `pgAdapterConfiguration.getMinSessions()` value here, or change the way that PGAdapter is being configured, so PGAdapter uses the same logic as the Spanner JDBC driver. Otherwise, someone might get surprising results when they are comparing the two with a high or low value for either of these settings.

benchmarks/tpcc/src/main/resources/application.properties
```diff
@@ -0,0 +1,85 @@
...
+# The number of items that can be bought. This number does not scale with the number of warehouses.
+# 100,000 is the number of items that is specified by the tpc-c standard.
+tpcc.item-count=100000
+
+# --- Possible optimizations for TPC-C --- #
+tpcc.use-read-only-transactions=false
+tpcc.lock-scanned-ranges=false
+
+# Change these to match your Cloud Spanner PostgreSQL-dialect database.
+spanner.project=appdev-soda-spanner-staging
```
olavloite: nit: update these to placeholder values

benchmarks/tpcc/src/main/resources/application.properties
```diff
@@ -0,0 +1,85 @@
...
+
+# Set this to true to instruct the benchmark runner to start a PGAdapter instance in-process with
+# the benchmark application.
+pgadapter.in-process=true
+pgadapter.num-channels=32
+
+# Set this if you want the in-process PGAdapter instance to use a specific service account
+# credentials file.
+# Leave unset if the application should use the APPLICATION_DEFAULT_CREDENTIALS.
+pgadapter.credentials=/Users/loite/Downloads/appdev-soda-spanner-staging.json
```
olavloite: nit: update to placeholder value

olavloite (approved):

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
hengfengli: Removed. It's from the previous code.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
hengfengli: What is the purpose of `lock_scanned_ranges ` hint? I did not know we need to add it to all queries/DMLs. 

I moved the logic to `paramQueryRow`, `executeParamStatement`, and also a few places using `executeParamQuery`.

benchmarks/tpcc/src/main/resources/application.properties
hengfengli: Updated.

benchmarks/tpcc/src/main/resources/application.properties
hengfengli: Updated.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
hengfengli: Updated to use `pgAdapterConfiguration.getMinSessions()` and `pgAdapterConfiguration.getMaxSessions()` here.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/AbstractBenchmarkRunner.java
olavloite: The hint is a potential optimization in a use-case like this. Whether it actually improves throughput or not, is currently not clear to me, but in theory it might. Adding this hint makes Spanner behave more like a 'traditional' database in case of lock conflicts; instead of optimistically only taking a read-lock, it will instruct Spanner to take an exclusive write lock.

And now that I think of it: I was the one who only added this in some places.... I added it where the standard implementation would use `select ... for update`. Which is where we would potentially want to add this. Sorry for the confusion. Let's just leave it as is for now, and optimize this in a follow-up PR (if necessary).

hengfengli: Yeah, probably, it's not quite right to use `exclusive lock` to every query and DML. I will create a PR to revert this change.

PR #1923, opened 2024-06-04
deps: update opentelemetry.exporter.version to 0.29

hengfengli:
Problem: with the old version, it can get weird an error reporting that the type of a field label is BOOL instead of STRING when creating MetricDescriptor. Similar to this [issue](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/159) which should be fixed in the latest version. 

After upgrading to the 0.29 version, the metrics in Cloud Monitoring will be put into `Generic Node` category instead of `Generic Task` category.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
     <excludedIntegrationTests/>
 
     <junixsocket.version>2.9.1</junixsocket.version>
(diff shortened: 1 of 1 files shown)


hengfengli:
As discussed offline, we need update the min Java version of PGAdapter in order to make this upgrade. Close this PR for now.

PR #1961, opened 2024-06-16
chore: add client lib PG runner to tpcc benchmark

hengfengli:
Add a runner that uses client library with PostgreSQL dialect.

### DIFF
--- a/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
+++ b/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
@@ -118,10 +118,7 @@ public class BenchmarkApplication implements CommandLineRunner {
       }
 
       if (tpccConfiguration.isRunBenchmark()
-          && (tpccConfiguration.getBenchmarkRunner().equals(TpccConfiguration.PGADAPTER_JDBC_RUNNER)
-              || tpccConfiguration
-                  .getBenchmarkRunner()
-                  .equals(TpccConfiguration.SPANNER_JDBC_RUNNER))) {
+          && (TpccConfiguration.RUNNERS.contains(tpccConfiguration.getBenchmarkRunner()))) {
         LOG.info("Starting benchmark");
         // Enable the OpenTelemetry metrics in the client library.
         OpenTelemetry openTelemetry = enableOpenTelemetryMetrics();
@@ -137,7 +134,12 @@ public class BenchmarkApplication implements CommandLineRunner {
             statistics.setRunnerName("PGAdapter benchmark");
             executor.submit(
                 new JdbcBenchmarkRunner(
-                    statistics, pgadapterConnectionUrl, tpccConfiguration, metrics));
+                    statistics,
+                    pgadapterConnectionUrl,
+                    tpccConfiguration,
+                    pgAdapterConfiguration,
+                    spannerConfiguration,
+                    metrics));
           } else if (tpccConfiguration
               .getBenchmarkRunner()
               .equals(TpccConfiguration.SPANNER_JDBC_RUNNER)) {
@@ -145,7 +147,24 @@ public class BenchmarkApplication implements CommandLineRunner {
             statistics.setRunnerName("Spanner JDBC benchmark");
             executor.submit(
                 new JdbcBenchmarkRunner(
-                    statistics, spannerConnectionUrl, tpccConfiguration, metrics));
+                    statistics,
+                    spannerConnectionUrl,
+                    tpccConfiguration,
+                    pgAdapterConfiguration,
+                    spannerConfiguration,
+                    metrics));
+          } else if (tpccConfiguration
+              .getBenchmarkRunner()
+              .equals(TpccConfiguration.CLIENT_LIB_PG_RUNNER)) {
+            // Run client library PG benchmark
+            statistics.setRunnerName("Client library PG benchmark");
+            executor.submit(
+                new JavaClientBenchmarkRunner(
+                    statistics,
+                    tpccConfiguration,
+                    pgAdapterConfiguration,
+                    spannerConfiguration,
+                    metrics));
           }
         }
 
--- a/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
+++ b/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
@@ -179,7 +179,8 @@ public class BenchmarkApplication implements CommandLineRunner {
           throw new TimeoutException("Timed out while waiting for benchmark runners to shut down");
         }
       } else {
-        LOG.error("Unknown benchmark runner option: " + tpccConfiguration.getBenchmarkRunner());
+        throw new RuntimeException(
+            "Unknown benchmark runner option: " + tpccConfiguration.getBenchmarkRunner());
       }
     } catch (IOException exception) {
       throw new RuntimeException(exception);
--- a/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
+++ b/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
@@ -117,8 +117,10 @@ public class BenchmarkApplication implements CommandLineRunner {
         System.out.printf("Finished loading %d rows\n", loadDataFuture.get());
       }
 
-      if (tpccConfiguration.isRunBenchmark()
-          && (TpccConfiguration.RUNNERS.contains(tpccConfiguration.getBenchmarkRunner()))) {
+      if (!tpccConfiguration.isRunBenchmark()) {
+        return;
+      }
+      if (TpccConfiguration.RUNNERS.contains(tpccConfiguration.getBenchmarkRunner())) {
         LOG.info("Starting benchmark");
         // Enable the OpenTelemetry metrics in the client library.
         OpenTelemetry openTelemetry = enableOpenTelemetryMetrics();
--- a/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
+++ b/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
@@ -117,8 +117,8 @@ class JavaClientBenchmarkRunner extends AbstractBenchmarkRunner {
     bindParams(builder, params);
     Statement stmt = builder.build();
 
-    ResultSet resultSet = transactionThreadLocal.get().executeQuery(stmt);
     Stopwatch stopwatch = Stopwatch.createStarted();
+    ResultSet resultSet = transactionThreadLocal.get().executeQuery(stmt);
     // The above executeQuery() does not actually execute the query until we call
     // next() for the first time.
     boolean is_next_successful = resultSet.next();
@@ -166,6 +166,10 @@ class JavaClientBenchmarkRunner extends AbstractBenchmarkRunner {
     int paramIndex = 1;
     while (index < sql.length()) {
       char c = sql.charAt(index);
+      // We have an assumption in any SQL strings: 1) comments do not contain question
+      // marks; 2) string literals do not contain question marks. This is because we
+      // believe all benchmark queries are hard-coded and none of them can cause the
+      // above cases. If any of above cases happen, it should fail explicitly.
       if (c == paramChar) {
         named.append(namedParamPrefix).append(paramIndex);
         paramIndex++;
@@ -216,8 +220,8 @@ class JavaClientBenchmarkRunner extends AbstractBenchmarkRunner {
     bindParams(builder, params);
     Statement stmt = builder.build();
 
-    ResultSet resultSet = transactionThreadLocal.get().executeQuery(stmt);
     Stopwatch stopwatch = Stopwatch.createStarted();
+    ResultSet resultSet = transactionThreadLocal.get().executeQuery(stmt);
     // The above executeQuery() does not actually execute the query until we call
     // next() for the first time.
     boolean is_next_successful = resultSet.next();
--- a/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
+++ b/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
@@ -1,0 +1,257 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.google.cloud.pgadapter.tpcc;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.pgadapter.tpcc.config.PGAdapterConfiguration;
+import com.google.cloud.pgadapter.tpcc.config.SpannerConfiguration;
+import com.google.cloud.pgadapter.tpcc.config.TpccConfiguration;
+import com.google.cloud.spanner.AbortedException;
+import com.google.cloud.spanner.DatabaseClient;
+import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.SessionPoolOptions;
+import com.google.cloud.spanner.Spanner;
+import com.google.cloud.spanner.SpannerOptions;
+import com.google.cloud.spanner.Statement;
+import com.google.cloud.spanner.TransactionContext;
+import com.google.cloud.spanner.TransactionManager;
+import com.google.cloud.spanner.Value;
+import com.google.common.base.Stopwatch;
+import com.google.common.base.Strings;
+import com.google.common.util.concurrent.Uninterruptibles;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.sql.SQLException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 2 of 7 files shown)


benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
```diff
@@ -118,10 +118,7 @@ public void run(String... args) throws Exception {
       }
 
       if (tpccConfiguration.isRunBenchmark()
-          && (tpccConfiguration.getBenchmarkRunner().equals(TpccConfiguration.PGADAPTER_JDBC_RUNNER)
-              || tpccConfiguration
-                  .getBenchmarkRunner()
-                  .equals(TpccConfiguration.SPANNER_JDBC_RUNNER))) {
+          && (TpccConfiguration.RUNNERS.contains(tpccConfiguration.getBenchmarkRunner()))) {
```
olavloite: nit: should we show a clear error message if an invalid benchmark runner name has been given? Now we just skip running any benchmarks.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+import java.sql.SQLException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class JavaClientBenchmarkRunner extends AbstractBenchmarkRunner {
+  private static final Logger LOG = LoggerFactory.getLogger(JavaClientBenchmarkRunner.class);
+  private final ThreadLocal<Spanner> spannerThreadLocal = new ThreadLocal<>();
```
olavloite: `Spanner` instances are intended a long-lived shared objects. So we should not create a separate `Spanner` instance for each thread.
The same also applies to `DatabaseClient`, but that happens automatically if you use a single `Spanner` instance. Multiple calls to `Spanner.getDatabaseClient(DatabaseId)` will return the same `DatabaseClient` instance.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+      Statistics statistics,
+      TpccConfiguration tpccConfiguration,
+      PGAdapterConfiguration pgAdapterConfiguration,
+      SpannerConfiguration spannerConfiguration,
+      Metrics metrics) {
+    super(statistics, tpccConfiguration, pgAdapterConfiguration, spannerConfiguration, metrics);
+  }
+
+  void setup() throws SQLException, IOException {
+    Spanner spanner = createSpanner();
```
olavloite: We should not do this. Instead, the entire benchmark should use a single, shared `Spanner` instance.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+    spannerThreadLocal.set(spanner);
+    databaseClientThreadLocal.set(
+        spanner.getDatabaseClient(
+            DatabaseId.of(
+                spannerConfiguration.getProject(),
+                spannerConfiguration.getInstance(),
+                spannerConfiguration.getDatabase())));
+    transactionManagerThreadLocal.set(databaseClientThreadLocal.get().transactionManager());
+    Uninterruptibles.sleepUninterruptibly(Duration.ofMillis(2000L));
```
olavloite: Why is this here?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  void teardown() throws SQLException {
+    transactionManagerThreadLocal.get().close();
+    spannerThreadLocal.get().close();
+  }
+
+  void executeStatement(String sql) throws SQLException {
+    if (sql.equals("begin transaction")) {
+      transactionThreadLocal.set(transactionManagerThreadLocal.get().begin());
```
olavloite, line 76: Does this really work? As in: `TransactionManager` instances are not intended to execute multiple transactions. Executing `TransactionManager#begin()` multiple times using the same `TransactionManager` should normally not work.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+  void setup() throws SQLException, IOException {
+    Spanner spanner = createSpanner();
+    spannerThreadLocal.set(spanner);
+    databaseClientThreadLocal.set(
+        spanner.getDatabaseClient(
+            DatabaseId.of(
+                spannerConfiguration.getProject(),
+                spannerConfiguration.getInstance(),
+                spannerConfiguration.getDatabase())));
+    transactionManagerThreadLocal.set(databaseClientThreadLocal.get().transactionManager());
```
olavloite: See also below: `TransactionManager` is intended to execute one transaction and then be disposed, not to execute multiple, consecutive transactions.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  Object[] paramQueryRow(String sql, Object[] params) throws SQLException {
+    ParametersInfo parametersInfo = convertPositionalParametersToNamedParameters(sql, '?', "$");
+    Statement.Builder builder = Statement.newBuilder(parametersInfo.sqlWithNamedParameters);
+    bindParams(builder, params);
+    Statement stmt = builder.build();
+
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    ResultSet resultSet = transactionThreadLocal.get().executeQuery(stmt);
```
olavloite, line 121: This does not actually execute the query. You will see a query execution time in the range of a few microseconds, or maybe even less, but that is because `executeQuery` does not actually execute the query. Instead, the query will only be executed when you call `ResultSet#next()`

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+    Duration executionDuration = stopwatch.elapsed();
+    metrics.recordLatency(executionDuration.toMillis());
+
+    if (!resultSet.next()) {
```
olavloite: This is where the query is actually executed.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+
+  public ParametersInfo convertPositionalParametersToNamedParameters(
+      String sql, char paramChar, String queryParameterPrefix) {
+    final String namedParamPrefix = queryParameterPrefix;
+    StringBuilder named = new StringBuilder(sql.length() + countOccurrencesOf(paramChar, sql));
+    int index = 0;
+    int paramIndex = 1;
+    while (index < sql.length()) {
+      char c = sql.charAt(index);
+      if (c == paramChar) {
```
olavloite, line 173: This does not work if the SQL string contains for example comments or a string literal with a positional parameter.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+      if (string.charAt(i) == c) {
+        res++;
+      }
+    }
+    return res;
+  }
+
+  // If getAsObject() in Struct can be a public method, then we should use it
+  // instead of using the following approach.
+  Object getObject(Value value) {
```
olavloite, line 198: Instead of relying on exception handling, can't you just call `Value#getType()` and use that in a switch statement to get the correct type?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+  List<Object[]> executeParamQuery(String sql, Object[] params) throws SQLException {
+    ParametersInfo parametersInfo = convertPositionalParametersToNamedParameters(sql, '?', "$");
+    List<Object[]> results = new ArrayList<>();
+
+    Statement.Builder builder = Statement.newBuilder(parametersInfo.sqlWithNamedParameters);
+    bindParams(builder, params);
+    Statement stmt = builder.build();
+
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    ResultSet resultSet = transactionThreadLocal.get().executeQuery(stmt);
```
olavloite, line 224: Here also: This does not actually execute the query, so the value that you are measuring is not relevant and will be in the 'few microseconds' range.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,257 @@
...
+    Duration executionDuration = stopwatch.elapsed();
+    metrics.recordLatency(executionDuration.toMillis());
+
+    while (resultSet.next()) {
```
olavloite: The first call to `resultSet.next()` executes the query

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/BenchmarkApplication.java
hengfengli: If you look at `else`, we actually have the following: 

```
LOG.error("Unknown benchmark runner option: " + tpccConfiguration.getBenchmarkRunner());
```

Updated to throw an exception.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: I looked at your old code: https://github.com/GoogleCloudPlatform/pgadapter/pull/1267/files#diff-0fdd0faac71de04c6b133638083332b7e9210d05921f1acb089a13e311f87647R61

Should I delete it? I thought you must have a reason to add this.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Actually, it works. I thought we can call `TransactionManager` to begin a new transaction and then call `transactionManagerThreadLocal.get().commit();` or `transactionManagerThreadLocal.get().rollback();` to complete the transaction. After that, we can call `begin` again to get a new transaction. 

From what you say, do we always need to create a new `TransactionManager` to begin a new transaction?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Updated `spanner` to be a shared object and removed `DatabaseClient`.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Updated.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Removed.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Yeah, I forgot this. Updated this.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: I looked at the original parser but that is way complicated for this benchmark. I think this simplified version works okay as all queries are hard-coded and we know the format of generated strings.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Fixed.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Updated.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Fixed.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Good point. I didn't know `Value#getType()`. Cannot use switch as it requires `constant expression required`. Updated.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Fixed.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Fixed.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: I tried to set `lock-scanned-ranges` to true which introduces some comments. This works fine. For `a string literal with a positional parameter`, I don't see why we would have such a case in the benchmark. This code just simply replace `?` to positional parameters with `$<num>` (later, also `@p<num>`).

olavloite: If you are certain that the benchmark does not have any SQL strings (and will not in the future) that might contain:
1. Comments that contain question marks.
2. String literals that contain question marks.

Then this should be fine.

But we should at least add a comment here to call that out clearly and explain that it is a deliberate choice, as a new developer who investigates this code in 2 years from now without any further context, might wonder how this ever worked.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
olavloite: (Note for the future: You can call `value.getType().getTypeCode()` and use that in a switch statement)

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  Object[] paramQueryRow(String sql, Object[] params) throws SQLException {
+    ParametersInfo parametersInfo = convertPositionalParametersToNamedParameters(sql, '?', "$");
+    Statement.Builder builder = Statement.newBuilder(parametersInfo.sqlWithNamedParameters);
+    bindParams(builder, params);
+    Statement stmt = builder.build();
+
+    ResultSet resultSet = transactionThreadLocal.get().executeQuery(stmt);
+    Stopwatch stopwatch = Stopwatch.createStarted();
```
olavloite, line 120: nit: If you want to have a fully apples-to-apples comparison, then this should be moved to before the `executeQuery(..)` call above. `executeQuery(..)` does not actually execute the query, but it can include checking out the session from the pool. This is also always included in the `executeQuery(..)` methods in the JDBC driver and PGAdapter.

olavloite (approved):

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Good point. Added a comment.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JavaClientBenchmarkRunner.java
hengfengli: Good catch. Fixed.

PR #2014, opened 2024-06-28
feat: return exit code 0 on successful shutdown

olavloite:
The default behavior of Java applications is to return exit code 128 + signal if the application was shut down due to receiving a signal. This means that SIGINT returns 130 and SIGTERM returns 143.

This change modifies this behavior so exit code 0 is returned if the shutdown is successful. The old behavior that returns the JVM default can be restored by starting PGAdapter with the -exit_mode startup argument.

This also prepares the server for adding Smart Shutdown mode, which will allow the server to start refusing new connections, while giving existing connections time to finish their work. See https://www.postgresql.org/docs/current/server-shutdown.html for more information on Smart Shutdown.

### DIFF
--- a/.ci/run-with-credentials.sh
+++ b/.ci/run-with-credentials.sh
@@ -63,7 +63,7 @@ e2e-psql)
   echo "Starting PGAdapter"
   ls -lh target
   UBER_JAR="pgadapter.jar"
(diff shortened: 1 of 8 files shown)


hengfengli (approved):
LGTM. Just curious: what is the main use case of this change? If both returns 0 as an exit code, we cannot distinguish which signal terminates the process, right?

olavloite:
Do not merge: This is probably going to be replaced by #2079

olavloite:
Closing in favor of #2079

Issue #2040, opened 2024-07-03
Exit PGAdapter Sidecar when main container exits


hiradp:
Thank you in advance for your time!

I have a kubernetes job that's using pgadapter as a side car. When the main container finishes, the pod get stuck in `NotReady` because pg adpter is running.

Is there native way to handle this properly?

I've seen this stackoverflow post https://stackoverflow.com/questions/41679364/kubernetes-stop-cloudsql-proxy-sidecar-container-in-multi-container-pod-job but this looks a bit of a hack.

olavloite:
@hiradp Thanks for reporting this. I'm afraid there isn't any really better methods available at the moment. Sending PGAdapter a `SIGINT` or `SIGTERM` is currently the only way to shut it down.

We are considering an alternative, though. Normal PostgreSQL can be shut down by sending it a command like `COPY (SELECT 1) TO PROGRAM 'pg_ctl stop -m smart --no-wait';`. For PGAdapter, we're considering making this even simpler by just adding a custom SQL statement `SHUTDOWN [SMART|FAST|IMMEDIATE]`. This statement would only be enabled if PGAdapter is started with a command line argument that allows it to be executed, which would typically be used in a situation where it is running as a side-car container. Would that work in your case?

See https://www.postgresql.org/docs/current/server-shutdown.html for more information on the shutdown modes of PostgreSQL.

hiradp:
Yep that would work in our case. In fact, I wanted to suggest using the same pattern as session management to conduct this. 

For anybody else that come across this. We managed to get it working with the following:

- Main container:

```yaml
              command:
                - /bin/bash
                - -c
              args:
                - |
                  trap "touch /cache/kill" EXIT
                  ./command
```

- PGAdapter Sidecar:

```yaml
              command:
                - /bin/bash
                - -c
              args:
                - |
                  /home/pgadapter/startup.sh -p project -i instance -d db &
                  PID=$!
                  while [ ! -e "/cache/kill" ]; do
                      sleep 1;
                  done
                  kill $PID

```

hiradp:
Also cloud-sql-proxy exposes a `/quitquitquit` endpoint but that's probably super involved if you don't have that infra setup.

Issue #1189, opened 2023-11-17
feature request: new 'max-sigterm-delay' flag 


magiccrafter:
The `--max-sigterm-delay` flag is crucial for microservices that require a graceful shutdown period to ensure ongoing transactions are completed successfully. This flag allows the microservice to remain operational for a specified duration after receiving a termination signal (SIGTERM), granting sufficient time for its database connector to remain active and handle pending transactions.

flag | description 
--- | --- 
 --max-sigterm-delay | Maximum number of seconds to wait for connections to close after receiving a TERM signal. 

Disclaimer: The name of the flag and the description are copied from the https://github.com/GoogleCloudPlatform/cloud-sql-proxy

Implementing this feature will smoothen the transition from Postgres to Spanner.

olavloite:
Version 0.38.0 that was shipped recently improved shutdown handling:
1. SIGTERM: This initiates the SMART shutdown mode. In this mode, all new connection requests will be refused, but existing connections are allowed to continue. PGAdapter shuts down when all connections have been closed.
2. SIGINT: This initiates the FAST shutdown mode. This closes all existing connections and shuts down PGAdapter.
3. SIGQUIT: (This is relatively theoretical, as JVMs do not allow applications to register a handler for SIGQUIT.) SIGQUIT initiates the IMMEDIATE shutdown mode. For PGAdapter, this is largely the same as the FAST shutdown mode, with the only exception that in IMMEDIATE shutdown mode, PGAdapter does not wait for the internal Spanner session pool to be cleaned up.

The above is designed to align as much as possible with https://www.postgresql.org/docs/current/server-shutdown.html. You can 'upgrade' from SMART to FAST shutdown mode by first sending PGAdapter a SIGTERM, then wait X seconds, and then send it a SIGINT.

PR #2079, opened 2024-07-16
feat: register signal handlers for graceful shutdown

olavloite:
Register signal handlers for SIGTERM, SIGINT, and SIGQUIT to allow graceful shutdown, and add support for the SMART shutdown mode from PostgreSQL. The SMART shutdown mode allows the server to drain connections and shut down after all connections have closed.

Registering signal handlers is a best-effort and requires that the `sun.misc.Signal` and `sun.misc.SignalHandler` classes are available on the JVM where PGAdapter is running.

Updates #2040
Updates #1189


src/main/java/com/google/cloud/spanner/pgadapter/Server.java
```diff
@@ -40,15 +41,30 @@
...
 import java.util.Locale;
 import java.util.Objects;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
+import org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement;
+
+// import sun.misc.Signal;
```
hengfengli: Remove?

src/test/java/com/google/cloud/spanner/pgadapter/ShutdownModeMockServerTest.java
```diff
@@ -0,0 +1,229 @@
...
+      shutdownHandler.shutdown(ShutdownMode.SMART);
+      // Wait until the server port is no longer open.
+      Stopwatch stopwatch = Stopwatch.createStarted();
+      while (isPortListening(proxyServer.getLocalPort())
+          && stopwatch.elapsed(TimeUnit.MILLISECONDS) < 2000) {
+        Thread.yield();
+      }
+      assertEquals(State.STOPPING, proxyServer.state());
+      // Verify that we can still use the existing connection.
+      try (ResultSet resultSet = connection.createStatement().executeQuery(sql)) {
```
hengfengli, line 93: If this connection is not closed, pgadapter will not shut down under SMART mode, right?

src/test/java/com/google/cloud/spanner/pgadapter/ShutdownModeMockServerTest.java
```diff
@@ -0,0 +1,229 @@
...
+      }
+      // Verify that the server is still stopping, but not yet terminated.
+      assertEquals(State.STOPPING, proxyServer.state());
+    }
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    while (proxyServer.state() == State.STOPPING
+        && stopwatch.elapsed(TimeUnit.MILLISECONDS) < 2000) {
+      Thread.yield();
+    }
+    // Verify that closing the last connection terminates the server.
```
hengfengli, line 120: So the last connection is closed at line 114, right?

src/test/java/com/google/cloud/spanner/pgadapter/ShutdownModeMockServerTest.java
```diff
@@ -0,0 +1,229 @@
...
+      shutdownHandler.shutdown(ShutdownMode.FAST);
+      // Verify that the server stops.
+      Stopwatch stopwatch = Stopwatch.createStarted();
+      while (proxyServer.state() != State.TERMINATED
+          && stopwatch.elapsed(TimeUnit.MILLISECONDS) < 2000) {
+        Thread.yield();
+      }
+      assertEquals(State.TERMINATED, proxyServer.state());
+
+      // Verify that existing connection has been invalidated.
```
hengfengli, line 149: So the existing connection gets the signal to terminate. Currently, FAST is almost same to IMMEDIATE, right?

hengfengli (approved):
LGTM. Left a few comments.

src/main/java/com/google/cloud/spanner/pgadapter/Server.java
olavloite: Oops. Removed.

src/test/java/com/google/cloud/spanner/pgadapter/ShutdownModeMockServerTest.java
olavloite: That is correct. However, initiating a FAST or IMMEDIATE shutdown while a SMART shutdown is pending, will override the SMART shutdown and close PGAdapter.

src/test/java/com/google/cloud/spanner/pgadapter/ShutdownModeMockServerTest.java
olavloite: Yes, correct. The end of the `try` block automatically closes the connection (also if the block is exited through for example an exception).

src/test/java/com/google/cloud/spanner/pgadapter/ShutdownModeMockServerTest.java
olavloite: Correct. The only difference at this moment is that FAST will wait for up to 1 second for the Spanner client to close (so wait for the session pool to clean up all sessions that were in use). IMMEDIATE will just do a fire-and-forget close of the Spanner client.

PR #2173, opened 2024-08-15
feature: add scripts to backup/restore a database

skuruppu:
Saving will copy the schema and data into a `backup` directory.

Restore will load the schema and data from the `backup` directory respecting the table hierarchy.


emulator-save-restore/README.md
```diff
@@ -0,0 +1,44 @@
+# Save and restore an emulator database
+
+This folder contains scripts to save a database created in the emulator and
+restore it when the emulator needs to be restarted.
+
+## Try it out
+
+1. Follow the
+   [instructions](https://github.com/GoogleCloudPlatform/pgadapter/blob/postgresql-dialect/docs/emulator.md#google-cloud-spanner-pgadapter---emulator)
```
olavloite: nit: Use `-` instead of an actual branch name to point to the default branch. This makes the link independent of any changes to the main branch in the future.

```suggestion
   [instructions](https://github.com/GoogleCloudPlatform/pgadapter/blob/-/docs/emulator.md#google-cloud-spanner-pgadapter---emulator)
```

emulator-save-restore/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+# Project, instance and database for Cloud Spanner.
+GOOGLE_CLOUD_PROJECT=my-project
+SPANNER_INSTANCE=my-instance
+SPANNER_DATABASE=my-database
+```
+
+```
+PGADAPTER_HOST=localhost
+PGADAPTER_PORT=9031
```
olavloite: Would it make sense to use the default `5432` in this example? That makes it easier for a user to understand that it is 'just the normal PGAdapter port' that it means.

emulator-save-restore/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+3. Save the schema and data in the emulator locally.
+
+```shell
+./save-emulator-database.sh
+
+```
+
+By default, the schema is saved in `./backup/schema.sql` and the data within
+individual files per table in the `.backup/data/` folder.
```
olavloite:
```suggestion
individual files per table in the `./backup/data/` folder.
```

emulator-save-restore/README.md
```diff
@@ -0,0 +1,44 @@
+# Save and restore an emulator database
```
olavloite: Is there any specific limitation that makes this emulator-only? Should we maybe rename the folder from `emulator-save-restore` to `save-restore`, and for now keep the documentation here as is, but with an additional disclaimer saying something like 'This script has currently only been tested for the Emulator. In the future, it might also support normal Cloud Spanner databases.'

emulator-save-restore/save-emulator-database.sh
```diff
@@ -0,0 +1,77 @@
...
+  -c "select tablename from pg_catalog.pg_tables where schemaname='public';" \
+| while read table_name ; do
+  column_names=$(psql -h "$PGADAPTER_HOST" -p "$PGADAPTER_PORT" -d "$SPANNER_DATABASE" -qAtX \
+    -c "select array_to_string(array_agg(column_name), ',') as c from information_schema.columns where table_name='$table_name' and table_schema='public' and not is_generated='ALWAYS';")
+
+  sql_file="$data_dir/$table_name.sql"
+
+  echo ""
+  echo "Copying data for $table_name"
+  echo "COPY $table_name ($column_names) FROM stdin;" > $sql_file
```
olavloite:
```suggestion
  echo "COPY $table_name ($column_names) TO stdout;" > $sql_file
```

emulator-save-restore/restore-emulator-database.sh
```diff
@@ -0,0 +1,170 @@
...
+# 2. Then all tables at hierarchical level 1, etc.
+
+# Get all tables without a parent.
+table_names=$(psql -v ON_ERROR_STOP=1 \
+    --host "$PGADAPTER_HOST" \
+    --port "$PGADAPTER_PORT" \
+    -d "$spanner_restore_database" \
+    -qAtX \
+    -c "select table_name from information_schema.tables
+            where table_schema not in ('information_schema', 'spanner_sys', 'pg_catalog')
```
olavloite: (Note: this might also be a bug in the script that was used as a base for this, so we might need to fix it there as well)

The import and export scripts do not seem to be completely consistent with regards to potential named schemas. The import script partly supports tables in other schemas than `public`, due to the where clause on this line. However, the table schema name is not used as part of the expected file name, and also not when looking for the corresponding parent table name.
Also, the export script explicitly only exports tables in `public`.

I think that for now it would be better to change this part of the script to also explicitly only use the `public` schema, and list it as a limitation in the documentation that it does not support named schemas yet.

olavloite (commented):
Looks largely good to me. Should we consider promoting this as save/restore in general, and not only specifically for the emulator?

I also added an integration test for it. That test runs on both real Spanner and the emulator, so it should tell us whether the scripts work on both.

emulator-save-restore/README.md
skuruppu: No, there's nothing specific in the scripts that should stop it from working against a prod database. Other than  not using any parallelism. I moved the scripts to indicate that it can be a general local backup/restore option. Also updated the docs.

emulator-save-restore/restore-emulator-database.sh
skuruppu: That's a really good point. I was hopeful that I was supporting named schemas but you're right that I wasn't resolving naming conflicts when writing to local files. I don't think it's a big change to support named schemas. But let's do that as a separate step.

emulator-save-restore/save-emulator-database.sh
skuruppu: But it can't be "TO" right? Because I'm printing a command that will be used for loading data in the restore script. This is so the per-table sql file we generate looks like:
```
COPY singers (singer_info,last_name,first_name,singer_id) FROM stdin;
\N      Smith   Catalina        2
\N      Martin  Lea     4
...
```

skuruppu:
Thanks for adding the test @olavloite. I've now put it in a more general place with appropriate naming. But with a caveat that it hasn't actually been tested for prod use.

I did actually test a smallish database based off [this one](https://github.com/catherinedevlin/opensourceshakespeare/blob/master/shakespeare.sql) on prod and the backup/restore both worked.

However, when I tried to restore the prod db to the emulator, I got the following error.
```
ERROR:  Operation with name "projects/your-project/instances/test-instance/databases/example-db/operations/rd49bfaac_dedd_4e4b_bab7_7880086131e8" failed with status = GrpcStatusCode{transportCode=FAILED_PRECONDITION} and message = Foreign key `paragraph_chapter_fkey` constraint violation on table `shakespeare_paragraph`. Cannot find referenced key `{String("12night"), Int64(5), Int64(2)}` in table `shakespeare_chapter`.
```

I'm guessing this has something to do with me trying to restore something in the emulator when I didn't generate the backup from the emulator. But given that a full backup and restore cycle worked on prod, I'm not so worried.

PTAL

emulator-save-restore/save-emulator-database.sh
olavloite: Oh, sorry, you're right. I thought that this was output going to the actual console and not to the file that we generate. Meaning: I thought this was information for the user to indicate 'The script is now executing `copy ... to stdout`, which would make more sense in the export scripts.

emulator-save-restore/restore-emulator-database.sh
olavloite: SG

olavloite:
> I did actually test a smallish database based off [this one](https://github.com/catherinedevlin/opensourceshakespeare/blob/master/shakespeare.sql) on prod and the backup/restore both worked.
> 
> However, when I tried to restore the prod db to the emulator, I got the following error.
> 
> ```
> ERROR:  Operation with name "projects/your-project/instances/test-instance/databases/example-db/operations/rd49bfaac_dedd_4e4b_bab7_7880086131e8" failed with status = GrpcStatusCode{transportCode=FAILED_PRECONDITION} and message = Foreign key `paragraph_chapter_fkey` constraint violation on table `shakespeare_paragraph`. Cannot find referenced key `{String("12night"), Int64(5), Int64(2)}` in table `shakespeare_chapter`.
> ```
> 
> I'm guessing this has something to do with me trying to restore something in the emulator when I didn't generate the backup from the emulator. But given that a full backup and restore cycle worked on prod, I'm not so worried.
> 

Hmmmm..... I do worry a bit about that. I think that it might have to do with the data in the sample database that you used, and that it might contain certain characters that don't play nice with the CSV file format that is used. I noticed that quite a lot of the rows in the [sample database](https://raw.githubusercontent.com/catherinedevlin/opensourceshakespeare/master/shakespeare.sql) contain a line-feed. I wonder whether we currently handle that correctly.

The reason that you are then only seeing an error when exporting from prod and importing to emulator, but not when importing/exporting on the same, could be due to the fact that the emulator and production by default return query results in opposite order when there is no ORDER BY clause:
1. Prod returns in primary key order ASC.
2. Emulator returns in primary key order DESC.

olavloite:
Ah, I think I've figured out where the problem is coming from. The script to re-create the foreign keys is incorrect, as it does not explicitly state the order of the key columns in the query. That query then (probably) returns the columns in different orders on prod and the emulator.

olavloite:
OK, so a bit of additional testing has shown that:
1. Line feeds and quotes in the data seem to work without problems.
2. Multi-column foreign key constraints that referenced a unique constraint in a different order than the order of the unique constraint itself (e.g. primary key is defined as `(id1, id2)` but the foreign key is defined as `(id2, id1) references (id2, id1)`) were not re-created correctly.

The latter should be fixed in [4ad5ebf](https://github.com/GoogleCloudPlatform/pgadapter/pull/2173/commits/4ad5ebfea42e901c5fdad218be48615dfdcee459)

olavloite (approved):

PR #2211, opened 2024-08-22
fix: add translations for unsupported statements for Liquibase

### LINKED ISSUE #2187: Liquibase fails with UPDATE ... FROM statements are not supported
Recent versions of Liquibase use an `UPDATE ... FROM` statement to tag databases. These statements fail on Spanner, and PGAdapter needs to rewrite these to be able to support recent versions of Liquibase.

olavloite:
This change fixes two issues with Liquibase:
1. The 'tag' command did not work, as it used an UPDATE .. FROM statement.
2. Liquibase 4.28.0 and later did not work due to a NOT LIKE clause.

Both the above types of statements are now automatically replaced by PGAdapter by compatible statements.

Fixes #2187

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/statements/JdbcMetadataStatementHelper.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/statements/JdbcMetadataStatementHelper.java
@@ -34,9 +34,10 @@ class JdbcMetadataStatementHelper {
    * @return true if the query could be a JDBC metadata query, and false if it definitely not.
    */
   static boolean isPotentialJdbcMetadataStatement(String sql) {
-    // All JDBC metadata queries that need any replacements reference the pg_catalog schema or the
-    // pg_settings table.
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/statements/JdbcMetadataStatementHelper.java
```diff
@@ -49,8 +50,13 @@ static String replaceJdbcMetadataStatement(String sql) {
     if (sql.startsWith(PgJdbcCatalog.PG_JDBC_GET_SCHEMAS_PREFIX)) {
       return replaceGetSchemasQuery(sql);
     }
-    if (sql.equals(PgJdbcCatalog.PG_JDBC_GET_EDB_REDWOOD_DATE_QUERY)) {
```
olavloite, line 52: This statement was also for Liquibase, but is no longer needed, as we have a generic solution for `pg_setting`. `pg_setting` is automatically added to any query that references it as a common table expression.

skuruppu (approved):
Thanks for finding a work around for these Liquibase changes.

PR #2544, opened 2024-11-20
[Draft] TPC-C Hibernate runner

rayudu3745:
Transaction metrics , Hibernaterunner(v1) implementation

### DIFF
--- a/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/HibernateBenchmarkRunner.java
+++ b/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/HibernateBenchmarkRunner.java
@@ -34,6 +34,7 @@ import com.google.cloud.pgadapter.tpcc.entities.Warehouse;
 import com.google.cloud.spanner.Dialect;
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
+import jakarta.persistence.criteria.CriteriaUpdate;
 import jakarta.persistence.criteria.Root;
 import java.io.IOException;
 import java.math.BigDecimal;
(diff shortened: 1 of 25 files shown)


benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/HibernateBenchmarkRunner.java
```diff
@@ -0,0 +1,533 @@
+package com.google.cloud.pgadapter.tpcc;
```
olavloite, line 14: nit: add a copyright header to all new files

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/HibernateBenchmarkRunner.java
```diff
@@ -0,0 +1,533 @@
...
+import java.util.Random;
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+import org.hibernate.query.Query;
+
+public class HibernateBenchmarkRunner extends AbstractBenchmarkRunner {
+
+  private final SessionFactory sessionFactory;
+  private final ThreadLocal<Session> sessionThreadLocal = new ThreadLocal<>();
```
olavloite: This makes the session live for the entire lifetime of the benchmark. That is too long. A session can best be compared with a (business) transaction. So typically one session is used to execute one database transaction, or at most a few, and should have a lifespan of at most a few seconds.

So it is probably better to refactor this so that each new transaction (e.g. each call to `newOrder()`, `payment()` etc.) creates a new session, and closes the session at the end of the method.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/entities/Customer.java
```diff
@@ -0,0 +1,246 @@
...
+import java.sql.Timestamp;
+
+@Entity
+@Table(name = "customer")
+public class Customer {
+
+  @EmbeddedId
+  private CustomerId id;
+
+  @ManyToOne(fetch = FetchType.LAZY)
```
olavloite: I would lean towards making this FetchType.EAGER. The decision between LAZY and EAGER here depends on how often you get a customer and then call getDistrict(). If the latter happens (almost) in all cases that you get a customer, then you should mark this as EAGER.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/entities/District.java
```diff
@@ -0,0 +1,137 @@
...
+import jakarta.persistence.Table;
+import java.math.BigDecimal;
+
+@Entity
+@Table(name = "district")
+public class District {
+
+  @EmbeddedId private DistrictId id;
+
+  @ManyToOne(fetch = FetchType.LAZY)
```
olavloite: Same here also as for Customer: This should probably be EAGER, if we in most cases also call get warehouse() after having fetched a District.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/HibernateBenchmarkRunner.java
```diff
@@ -0,0 +1,533 @@
...
+        supplyWarehouses[line] = warehouseId;
+      }
+      quantities[line] = random.nextInt(1, 10);
+    }
+
+    Transaction tx = session.beginTransaction();
+    try {
+      Customer customer = session.get(Customer.class, new CustomerId(customerId, districtId, warehouseId));
+      District district = customer.getDistrict();
+      Warehouse warehouse = district.getWarehouse();
```
olavloite, line 116: These two will cause an extra SELECT statement if these are LAZY.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/HibernateBenchmarkRunner.java
```diff
@@ -0,0 +1,533 @@
...
+      // Create new Order and NewOrder entities
+      Order order = new Order();
+      OrderId orderId = new OrderId(districtNextOrderId, customerId, districtId, warehouseId);
+      order.setId(orderId);
+      order.setEntryD(new Timestamp(System.currentTimeMillis()));
+      order.setOlCnt(orderLineCount);
+      order.setAllLocal(allLocal);
+      order.setCustomer(customer);
+
+
```
olavloite: nit: only one empty line

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/Metrics.java
```diff
@@ -23,10 +23,33 @@
 public class Metrics {
   static final String INSTRUMENTATION_SCOPE = "cloud.google.com/java";
   static final String ROUNDTRIP_LATENCY = "spanner/benchmarker/roundtrip_latencies";
+
   static final String ROUNDTRIP_LATENCY_DESCRIPTION =
       "Latency when the benchmark runner sends out a query and gets a response";
 
+  static final String TPCC_NEWORDER_LATENCY = "spanner/benchmarker/tpcc_neworder_latencies";
```
hengfengli, line 30: Have we considered to add only one metric, e.g., `spanner/benchmarker/tpcc_transaction_latencies` and use a field in the Attributes to indicate the transaction type?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/SessionHelper.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public Session createAutoBatchDmlSession(String transactionTag) {
+    Session session = sessionFactory.openSession();
+    session.doWork(
+        conn -> {
+          conn.createStatement().execute("RESET ALL");
+          conn.setReadOnly(false);
+          conn.createStatement().execute("set auto_batch_dml=true");
+          conn.createStatement().execute("set auto_batch_dml_update_count_verification=false");
+          //conn.createStatement().execute("set transaction_tag='" + transactionTag + "'");
```
hengfengli, line 39: Remove?

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/SessionHelper.java
```diff
@@ -0,0 +1,53 @@
+package com.google.cloud.pgadapter.tpcc;
+
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+
+public class SessionHelper {
```
hengfengli, line 6: Should we be more specific to Hibernate session, e.g., HibernateSessionHelper?

benchmarks/tpcc/src/main/resources/application.properties
```diff
@@ -54,13 +55,13 @@ spanner.database=my-database
...
 # the benchmark application.
-pgadapter.in-process=true
+pgadapter.in-process=false
 pgadapter.num-channels=32
 
 # Set this if you want the in-process PGAdapter instance to use a specific service account
 # credentials file.
 # Leave unset if the application should use the APPLICATION_DEFAULT_CREDENTIALS.
-pgadapter.credentials=/path/to/credentials.json
+pgadapter.credentials=/Users/rayudualp/Downloads/span-cloud-testing-5084c5e7353d.json
```
hengfengli, line 66: We probably should not update this file except adding the hibernate settings below.

hengfengli (commented):
LGTM. Please also add instructions how to run the Hibernate runner in https://github.com/GoogleCloudPlatform/pgadapter/tree/postgresql-dialect/benchmarks/tpcc.

818Nawaf (approved):

PR #2781, opened 2025-01-18
test: fix flaky random results test

### LINKED ISSUE #2688: AbortedMockServerTest.testRandomResults is flaky for dates in year 100
```
Error:  Failures: 
Error:    AbortedMockServerTest.testRandomResults:929->assertEqual:1021 expected:<0100-02-19> but was:<0100-02-20>
```

olavloite:
The random results test would sometimes fail for dates in the year 100. This was caused by the getDate(..) method using the local timezone to calculate the date based on the number of days since the PG epoch.

Instead of using the getDate(..) method, the date should be retrieved as a LocalDate, as that does not use any timezone information or rely on any of the weird java.util.Date / java.sql.Date types.

Fixes #2688


src/main/java/com/google/cloud/spanner/pgadapter/parsers/DateParser.java
```diff
@@ -23,25 +23,13 @@
...
-  // Timezone information is optional. Timezone may also be specified using only hour value.
-  // NOTE: Following algorithm might perform slowly due to exception handling; sadly, this seems
-  //       to be the accepted default method for date validation.
-  private static final SimpleDateFormat[] VALID_DATE_FORMATS = {
-    new SimpleDateFormat("yyyy-MM-dd HH:mm"),
-    new SimpleDateFormat("yyyy-MM-dd +HH:mm"),
-    new SimpleDateFormat("yyyy-MM-dd -HH:mm"),
-    new SimpleDateFormat("yyyy-MM-dd +HH"),
-    new SimpleDateFormat("yyyy-MM-dd -HH")
-  };
```
olavloite, line 44: This was unused, so removed it.

src/test/java/com/google/cloud/spanner/pgadapter/AbortedMockServerTest.java
```diff
@@ -237,8 +244,6 @@ public void testQueryHint() throws SQLException {
 
   @Test
   public void testQueryWithParameters() throws SQLException {
-    JdbcMockServerTest.setupJsonbResults(mockSpanner);
-
```
olavloite, line 241: Moved this to the test class setup method, so these results are available for all tests in this class.

src/test/java/com/google/cloud/spanner/pgadapter/AbortedMockServerTest.java
```diff
@@ -914,19 +919,9 @@ public void testRandomResults() throws SQLException {
...
-                //       try to read type information from the backend when it hits an 'unknown'
-                //       type (jsonb is not one of the types that the JDBC driver will load
-                //       automatically).
-                final int jsonbColumnIndex = 6;
-                final int jsonbArrayColumnIndex = 16;
-                if (col == jsonbColumnIndex || col == jsonbArrayColumnIndex) {
-                  resultSet.getString(col + 1);
-                } else {
-                  resultSet.getObject(col + 1);
-                }
```
olavloite, line 927: Removed this special handling, as the TODO has been resolved.

src/test/java/com/google/cloud/spanner/pgadapter/AbortedMockServerTest.java
```diff
@@ -941,6 +936,74 @@ public void testRandomResults() throws SQLException {
...
+                                .addFields(
+                                    Field.newBuilder()
+                                        .setName("dt")
+                                        .setType(Type.newBuilder().setCode(TypeCode.DATE).build())
+                                        .build())
+                                .build())
+                        .build())
+                .addRows(
+                    ListValue.newBuilder()
+                        .addValues(Value.newBuilder().setStringValue("0100-02-19").build())
```
olavloite, line 960: This is the specific date that caused errors

src/test/java/com/google/cloud/spanner/pgadapter/AbortedMockServerTest.java
```diff
@@ -1018,7 +1082,10 @@ private void assertEqual(com.google.cloud.spanner.ResultSet spannerResult, Resul
             // are weird, as are potential intercalaris values.
             assertNotNull(pgResult.getDate(col + 1));
           } else {
-            assertEquals(expected, pgResult.getDate(col + 1).toLocalDate());
```
olavloite, line 1021: Instead of calling `getDate(col + 1).toLocalDate()` (which internally uses the local timezone for the calculation), we should get the result directly as a `LocalDate` by calling `getObject(col + 1, LocalDate.class)`.

PR #2956, opened 2025-02-27
feat: remove default like escape clauses

hnarra7:
Removes default `LIKE ... ESCAPE '\'` clauses from statements. These clauses are often generated by frameworks, like for example Hibernate, but `ESCAPE '\'` is the default and therefore redundant. This change adds a check for such clauses and removes them from the statement.

The feature is enabled by default. It can be disabled with the connection variable `spanner.remove_default_escape`.


olavloite:
@hnarra7 I made some changes to the proposal, as the initial proposal had a few challenges:
1. We prefer not to have regex-based parsing and replacements in the standard query flow. That is; We use that for specific metadata queries, but only once we have detected that the query uses a system table (e.g. one of the `pg_catalog` tables). I therefore replaced the regex with a token-based parser solution.
2. We should only remove default `ESCAPE '\'` clauses. Removing all ESCAPE clauses would make the query behave differently than specification, as the escape clause that had been given would not be applied. The main problem that this intends to solve is the problem that some frameworks (like Hibernate) automatically include an ESCAPE clause, even when it is the default.
3. I made the feature an opt-out, so users can choose to skip this translation (e.g. for testing or performance purposes).
4. I also added a cache for the replace-function to reduce the performance impact of executing the same statement many times.

hnarra7:
@olavloite  Thanks for your proposed solution.

I tested your implementation and added a test case to check if ESCAPE '' is removed from the query. While your code correctly removes ESCAPE '\', it does not handle ESCAPE '', which is required for Spanner since it doesn’t support escape clauses.

To clarify, I’m sharing a reference to [PostgreSQLSqlAstTranslator.java](https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/dialect/PostgreSQLSqlAstTranslator.java) where you can see how ESCAPE '' is appended. Given how ESCAPE '\' is handled in EscapeClauseParser.java, would it be possible to apply a similar approach for ESCAPE '' ? Or would this impact other parts of the code?

Let me know your thoughts!

olavloite:
@hnarra7 The problem with removing `escape ''` clauses is that it also changes the intended behavior of the query. 

Consider the following three examples when executed on standard PostgreSQL:

```
loite=# select * from (select 'test\' as c) t where t.c like 'test\%';
 c 
---
(0 rows)

loite=# select * from (select 'test\' as c) t where t.c like 'test\%' escape '\';
 c 
---
(0 rows)

loite=# select * from (select 'test\' as c) t where t.c like 'test\%' escape '';
   c   
-------
 test\
(1 row)
```

The first two return zero rows, because backslash is treated as an escape character. This is the default in PostgreSQL, so adding it to the LIKE operator is a no-op.

However, if you add `ESCAPE ''` to the query, then it means that nothing should be considered an escape character, and the string `'test\'` all of a sudden matches the search string.

What does your Hibernate code look like that generates the LIKE expression? It should normally be possible to set the escape character that should be used, and in this case it should then be set to `\`, as that is what Spanner actually uses.

Another option would be that we add a second connection variable to PGAdapter that allows PGAdapter to remove all escape clauses, including non-default ones. This connection variable is disabled by default, as it makes PGAdapter behave differently than standard PostgreSQL.

olavloite:
@hnarra7 I've modified the connection variable `spanner.remove_escape_clause` to accept an enum value (NONE, DEFAULT, ALL). The default value is `DEFAULT` and will only remove `'\'` escape clauses. Set this connection variable to `ALL` to remove all ESCAPE clauses.

See this test for an example: https://github.com/GoogleCloudPlatform/pgadapter/blob/7d5eeb373126d62824dabe4177a1809d00a4ee9e/src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java#L5269

olavloite:
@hnarra7 Note that you need to sign the CLA for me to merge this pull request. If you'd rather not sign the CLA, then I can copy these changes to a separate PR and merge them that way, but I would like to give you the credit for bringing up this issue.

hnarra7:
> @hnarra7 I've modified the connection variable `spanner.remove_escape_clause` to accept an enum value (NONE, DEFAULT, ALL). The default value is `DEFAULT` and will only remove `'\'` escape clauses. Set this connection variable to `ALL` to remove all ESCAPE clauses.
> 
> See this test for an example:
> 
> https://github.com/GoogleCloudPlatform/pgadapter/blob/7d5eeb373126d62824dabe4177a1809d00a4ee9e/src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java#L5269

@olavloite  Thanks for implementing this solution and for the detailed explanation! I appreciate the effort in making spanner.remove_escape_clause configurable to handle different escape clause behaviors.

I see that this setting is now defined in pgsettings.txt with NONE, DEFAULT, and ALL as possible values. Could you confirm the recommended approach for configuring this dynamically? Should we update pgsettings.txt, or is there an alternative way to override this setting at runtime from our application?

Thanks again for your time and support!

hnarra7:
> @hnarra7 Note that you need to sign the CLA for me to merge this pull request. If you'd rather not sign the CLA, then I can copy these changes to a separate PR and merge them that way, but I would like to give you the credit for bringing up this issue.

@olavloite Thanks for letting me know about the CLA requirement. My company has already added me to the Google CLA group, but for some reason, the check is still failing. Could you help verify if there’s anything else I need to do on my end to resolve this?

I appreciate your guidance!

olavloite:
> > @hnarra7 Note that you need to sign the CLA for me to merge this pull request. If you'd rather not sign the CLA, then I can copy these changes to a separate PR and merge them that way, but I would like to give you the credit for bringing up this issue.
> 
> @olavloite Thanks for letting me know about the CLA requirement. My company has already added me to the Google CLA group, but for some reason, the check is still failing. Could you help verify if there’s anything else I need to do on my end to resolve this?
> 
> I appreciate your guidance!

Would you mind checking the possible causes listed here: https://github.com/GoogleCloudPlatform/pgadapter/pull/2956/checks?check_run_id=38335940156

olavloite:
> > @hnarra7 I've modified the connection variable `spanner.remove_escape_clause` to accept an enum value (NONE, DEFAULT, ALL). The default value is `DEFAULT` and will only remove `'\'` escape clauses. Set this connection variable to `ALL` to remove all ESCAPE clauses.
> > See this test for an example:
> > https://github.com/GoogleCloudPlatform/pgadapter/blob/7d5eeb373126d62824dabe4177a1809d00a4ee9e/src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java#L5269
> 
> @olavloite Thanks for implementing this solution and for the detailed explanation! I appreciate the effort in making spanner.remove_escape_clause configurable to handle different escape clause behaviors.
> 
> I see that this setting is now defined in pgsettings.txt with NONE, DEFAULT, and ALL as possible values. Could you confirm the recommended approach for configuring this dynamically? Should we update pgsettings.txt, or is there an alternative way to override this setting at runtime from our application?
> 
> Thanks again for your time and support!

You can set it dynamically in your application by executing `set spanner.remove_escape_clause=ALL`, for example like this:

```java
    try (Connection connection = DriverManager.getConnection(createUrl())) {
      connection.createStatement().execute("set spanner.remove_escape_clause=all");
      try (ResultSet resultSet =
          connection
              .createStatement()
              .executeQuery(
                  "SELECT * FROM users WHERE name LIKE 'test%' ESCAPE '%' AND city LIKE 'NY%' ESCAPE 'Y'")) {
      }
    }
```

Alternatively, if you want to set it to `ALL` for all connections by default, then add the following to your JDBC connection URL:

```
jdbc:postgresql://localhost:5432/my-database?options=-c%20spanner.remove_escape_clause=ALL
```

hnarra7:
> > > @hnarra7 I've modified the connection variable `spanner.remove_escape_clause` to accept an enum value (NONE, DEFAULT, ALL). The default value is `DEFAULT` and will only remove `'\'` escape clauses. Set this connection variable to `ALL` to remove all ESCAPE clauses.
> > > See this test for an example:
> > > https://github.com/GoogleCloudPlatform/pgadapter/blob/7d5eeb373126d62824dabe4177a1809d00a4ee9e/src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java#L5269
> > 
> > 
> > @olavloite Thanks for implementing this solution and for the detailed explanation! I appreciate the effort in making spanner.remove_escape_clause configurable to handle different escape clause behaviors.
> > I see that this setting is now defined in pgsettings.txt with NONE, DEFAULT, and ALL as possible values. Could you confirm the recommended approach for configuring this dynamically? Should we update pgsettings.txt, or is there an alternative way to override this setting at runtime from our application?
> > Thanks again for your time and support!
> 
> You can set it dynamically in your application by executing `set spanner.remove_escape_clause=ALL`, for example like this:
> 
> ```java
>     try (Connection connection = DriverManager.getConnection(createUrl())) {
>       connection.createStatement().execute("set spanner.remove_escape_clause=all");
>       try (ResultSet resultSet =
>           connection
>               .createStatement()
>               .executeQuery(
>                   "SELECT * FROM users WHERE name LIKE 'test%' ESCAPE '%' AND city LIKE 'NY%' ESCAPE 'Y'")) {
>       }
>     }
> ```
> 
> Alternatively, if you want to set it to `ALL` for all connections by default, then add the following to your JDBC connection URL:
> 
> ```
> jdbc:postgresql://localhost:5432/my-database?options=-c%20spanner.remove_escape_clause=ALL
> ```

@olavloite  Thanks for your detailed explanation and for implementing this solution! I really appreciate the overall effort you’ve put into making spanner.remove_escape_clause configurable and improving the handling of escape clauses in PGAdapter.

I tested the changes, and everything is working as expected. The escape ' ' clause is being removed successfully.

Thanks again for your time, support, and all the work you’ve done on this!

hnarra7:
> > > @hnarra7 Note that you need to sign the CLA for me to merge this pull request. If you'd rather not sign the CLA, then I can copy these changes to a separate PR and merge them that way, but I would like to give you the credit for bringing up this issue.
> > 
> > 
> > @olavloite Thanks for letting me know about the CLA requirement. My company has already added me to the Google CLA group, but for some reason, the check is still failing. Could you help verify if there’s anything else I need to do on my end to resolve this?
> > I appreciate your guidance!
> 
> Would you mind checking the possible causes listed here: https://github.com/GoogleCloudPlatform/pgadapter/pull/2956/checks?check_run_id=38335940156

@olavloite  I appreciate your patience! I am currently working on resolving the CLA access issue and actively looking into the necessary steps to get it sorted. I’ll keep you updated as soon as the CLA check is successful.

Thanks again for your support!

olavloite (approved):

olavloite:
@hnarra7 Let's just override the check for now so we can merge this. Thanks for working with me on this!

PR #3033, opened 2025-03-13
fix: drop table if exists .. cascade did not work

olavloite:
Dropping a table with an IF EXISTS clause did not work when support_drop_cascade had been enabled.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
@@ -202,8 +202,8 @@ class DdlExecutor {
     if (!parser.eatKeyword("drop")) {
       return defaultResult;
     }
-    parser.eatKeyword("if", "exists");
     if (parser.eatKeyword("table")) {
+      parser.eatKeyword("if", "exists");
       TableOrIndexName tableName = parser.readTableOrIndexName();
       if (tableName == null) {
         return defaultResult;
@@ -228,6 +228,7 @@ class DdlExecutor {
       builder.add(cascade ? Statement.of(sqlWithoutCascade) : statement);
       return builder.build();
     } else if (parser.eatKeyword("schema")) {
+      parser.eatKeyword("if", "exists");
       TableOrIndexName schemaName = parser.readTableOrIndexName();
       if (schemaName == null || schemaName.schema != null) {
         return defaultResult;
--- a/src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
+++ b/src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
@@ -101,7 +101,28 @@ public class ITDdlTest implements IntegrationTest {
         connection.createStatement().execute("set spanner.support_drop_cascade to on");
         // Dropping the table should now work.
         assertEquals(0, connection.createStatement().executeUpdate(dropStatement));
+        // Verify that the table was really dropped.
+        PSQLException psqlException =
+            assertThrows(
+                PSQLException.class,
+                () ->
+                    connection
+                        .createStatement()
+                        .execute("update " + tableName + " set value='' where true"));
+        assertEquals(SQLState.UndefinedTable.toString(), psqlException.getSQLState());
       }
+      // Dropping a table that does not exist with an 'if exists' clause also succeeds.
+      connection.createStatement().execute("set spanner.support_drop_cascade to on");
+      assertEquals(
+          0,
+          connection
+              .createStatement()
+              .executeUpdate("drop table if exists this_table_does_not_exist"));
+      assertEquals(
+          0,
+          connection
+              .createStatement()
+              .executeUpdate("drop table if exists this_table_does_not_exist cascade"));
     }
   }
 
--- a/src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
+++ b/src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
@@ -72,25 +72,36 @@ public class ITDdlTest implements IntegrationTest {
     try (Connection connection = DriverManager.getConnection(getConnectionUrl())) {
       // First create a table with a secondary index.
       try (Statement statement = connection.createStatement()) {
-        statement.addBatch("create table test (id bigint primary key, value varchar)");
-        statement.addBatch("create index idx_text on test (value)");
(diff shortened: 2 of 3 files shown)


src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
```diff
@@ -72,25 +72,36 @@ public void testDropTableWithIndex() throws SQLException {
...
-      // Now enable drop_cascade.
-      connection.createStatement().execute("set spanner.support_drop_cascade to on");
-      // Dropping the table should now work.
-      assertEquals(0, connection.createStatement().executeUpdate("drop table test"));
+      for (int index = 1; index <= 2; index++) {
+        connection.createStatement().execute("set spanner.support_drop_cascade to off");
+
+        String tableName = "test" + index;
+        boolean useIfExists = index == 2;
+        String dropStatement = "drop table " + (useIfExists ? "if exists " : "") + tableName;
```
hengfengli, line 87: This will always fail because there is no "cascade" keyword.

src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
```diff
@@ -72,25 +72,36 @@ public void testDropTableWithIndex() throws SQLException {
...
+        assertEquals(SQLState.FeatureNotSupported.toString(), exception.getSQLState());
+        assertNotNull(exception.getServerErrorMessage());
+        assertEquals(
+            "Execute 'set spanner.support_drop_cascade=true' to enable dropping tables with indices",
+            exception.getServerErrorMessage().getHint());
+
+        // Now enable drop_cascade.
+        connection.createStatement().execute("set spanner.support_drop_cascade to on");
+        // Dropping the table should now work.
+        assertEquals(0, connection.createStatement().executeUpdate(dropStatement));
```
hengfengli, line 103: In my testing, I found "drop table if exists <table_name> cascade" doesn't drop the table although it returns DROP successfully. We should add it as a test case. Without "cascade", it should always fail to drop a table with an index. With "cascade", it should be successful. 

Can we check if the table is really deleted?

src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
```diff
@@ -202,8 +202,8 @@ ImmutableList<Statement> getDependentStatements(Statement statement) {
     if (!parser.eatKeyword("drop")) {
       return defaultResult;
     }
-    parser.eatKeyword("if", "exists");
     if (parser.eatKeyword("table")) {
+      parser.eatKeyword("if", "exists");
```
hengfengli, line 206: What if the table doesn't exist? Will it just return an error? Can we add a test case for that?

src/test/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutorTest.java
```diff
@@ -174,19 +174,37 @@ ImmutableList<Statement> getDropSchemaViewsStatements(TableOrIndexName schemaNam
...
             Statement.of("drop all views in schema foo"),
             Statement.of("drop all tables in schema foo")),
         ddlExecutor.getDependentStatements(Statement.of("drop schema foo cascade")));
+    assertEquals(
+        ImmutableList.of(
+            Statement.of("drop all indexes in schema foo"),
+            Statement.of("drop all foreign keys in schema foo"),
+            Statement.of("drop all views in schema foo"),
+            Statement.of("drop all tables in schema foo")),
+        ddlExecutor.getDependentStatements(Statement.of("drop schema if exists foo cascade")));
```
hengfengli, line 207: If the table/schema doesn't exist, we will still return an error. This doesn't match the actual behaviour, right?

src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
olavloite: The `spanner.support_drop_cascade` enables the standard cascade behavior of PG. That means:
1. Dropping a table with an index works automatically (also without `cascade`, as this is the default in PG).
2. Dropping a table with other dependencies work if you include the `cascade` keyword.

So that means that the following does work (and if it didn't, then this test would have failed):

```
set spanner.support_drop_cascade to on;
create table foo (id bigint primary key, value varchar);
create index idx_foo on foo (value);
drop table foo;
```

src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
olavloite: See also above: Dropping a table with an index is supported in PostgreSQL, even without including the `cascade` keyword.

There are two possibilities for the scenario that a `drop table` statement seems to return successfully, but does not actually drop the table:

1. DDL batching:

```
start batch ddl;
-- The following statement will always return successfully.
-- Any error will be deferred to the RUN BATCH statement.
drop table foo;
drop table bar;

-- Any errors will only pop up once this is executed.
RUN BATCH;
```

2. Hibernate schema management: The Hibernate schema management has a built-in 'feature' that errors during schema dropping are just ignored. So if the `drop` statement failed during Hibernate schema management and did not show up anywhere, then this is likely the cause.

I'll add an additional validation to the test to verify that the table has really been dropped.

src/main/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutor.java
olavloite: Added

src/test/java/com/google/cloud/spanner/pgadapter/statements/DdlExecutorTest.java
olavloite: No, the original statement is still the one that is being executed. So even though the parser eats the `if exists` clause, that is only to determine whether the statement contains a `cascade` option, and to find the table name that we need to use to search for dependent objects. The parser is not literally used to execute the original drop statement, which remains `drop schema|table if exists ....`.

See also the additional tests that were added based on the comments above.

src/test/java/com/google/cloud/spanner/pgadapter/ITDdlTest.java
hengfengli: I see. An index is not like a foreign key constraint. Thanks for the explanation.

hengfengli (approved):
LGTM. Thanks for fixing this.

PR #3103, opened 2025-03-27
perf: use a stream mark to read null-terminated string

olavloite:
Prepared statements and portals can have names in the PG wire-protocol. These are sent as null-terminated strings. Read these strings by setting a mark in the stream and then scan for the null-terminator, instead of always creating a 128-byte buffer and read into that buffer. Using a stream mark will instead re-use the existing buffer in the existing input stream, and so reduce the amount of memory that is being reserved and released for each statement that is received.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
@@ -23,7 +23,6 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
-import java.util.Arrays;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
```diff
@@ -149,30 +148,47 @@ public String read(int length) throws IOException {
...
+
+  /**
+   * Reads a null-terminated string from a {@link DataInputStream}.
    *
    * @return the string.
    * @throws IOException if an error occurs while reading from the stream, or if no null-terminator
    *     is found before the end of the stream.
    */
   public String readString() throws IOException {
-    byte[] buffer = new byte[128];
```
olavloite, line 162: Getting rid of this is the main improvement here, as we would reserver a 128-byte buffer every time we wanted to read a string. Very often, these strings would be empty strings, for example in the case of unnamed statements and unnamed portals.

hengfengli: Thanks for the explanation.

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
```diff
@@ -149,30 +148,47 @@ public String read(int length) throws IOException {
...
+      if (index == MARK_READ_LIMIT) {
+        throw new IOException("No null terminator found");
       }
     }
-    return new String(buffer, 0, index, StandardCharsets.UTF_8);
+    // Reset the stream to the mark and read the name (if any).
+    this.inputStream.reset();
+    if (index == 0) {
+      // No name, but we still need to skip the null-terminator.
+      //noinspection StatementWithEmptyBody
```
hengfengli, line 181: Remove?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
```diff
@@ -149,30 +148,47 @@ public String read(int length) throws IOException {
...
+      while (this.inputStream.skip(1) < 1) {}
+      return "";
+    }
+
+    byte[] result = new byte[index];
+    this.inputStream.readFully(result);
+    // Skip the null-terminator.
+    //noinspection StatementWithEmptyBody
```
hengfengli, line 189: Same here. What is the purpose of this comment?

hengfengli (approved):

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/WireMessage.java
olavloite: This comment supresses a warning in IntelliJ. IntelliJ normally expect a `while` loop to have an empty body. In this case, we intentionally have an empty loop body, because:
1. `inputStream.skip(1)` is not guaranteed to actually skip anything. It returns 0 if no bytes were skipped.
2. We want to make sure that we actually skip 1 byte, so we execute the skip in a `while` loop.
3. There's nothing that we want to do in the loop itself, we just want to repeat the `skip` until it returns 1.

PR #3154, opened 2025-04-09
feat: update the metric buckets to a fine-grained strategy

hengfengli:
The current bucketing strategy is not precise enough. We need a more fine-grained bucket strategy.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
@@ -34,21 +34,18 @@ public class Metrics {
   private final DoubleHistogram pgadapterLatencies;
 
   public static List<Double> getMetricLatencyMillisBuckets() {
-    int kMaxNumFiniteBuckets = 50;
-    double BASE = 1.25;
-    double SCALE_FACTOR = 0.25;
-    double MAX_VALUE = 600;
-    double bucket_value = SCALE_FACTOR;
+    final int MAX_NUM_FINITE_BUCKETS = 50;
+    final double BASE = 1.25;
+    final double SCALE_FACTOR = 0.25;
+    final double MAX_VALUE = 600;
+    double bucketValue = SCALE_FACTOR;
 
-    List<Double> rpc_millis_bucket_boundaries = new ArrayList<Double>(kMaxNumFiniteBuckets);
-    for (int i = 0; i < kMaxNumFiniteBuckets; i++) {
-      if (bucket_value > MAX_VALUE) {
-        break;
-      }
-      rpc_millis_bucket_boundaries.add(bucket_value);
-      bucket_value = SCALE_FACTOR * BASE;
+    List<Double> rpcMillisBucketBoundaries = new ArrayList<Double>(MAX_NUM_FINITE_BUCKETS);
+    for (int i = 0; i < MAX_NUM_FINITE_BUCKETS && bucketValue <= MAX_VALUE; i++) {
+      rpcMillisBucketBoundaries.add(bucketValue);
+      bucketValue = SCALE_FACTOR * BASE;
     }
-    return rpc_millis_bucket_boundaries;
+    return rpcMillisBucketBoundaries;
   }
 
   public Metrics(OpenTelemetry openTelemetry) {
--- a/src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
@@ -21,6 +21,7 @@ import io.opentelemetry.api.metrics.Meter;
 import java.util.ArrayList;
 import java.util.List;
 
+@InternalApi
 public class Metrics {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
```diff
@@ -31,31 +30,42 @@ public class Metrics {
...
+    double BASE = 1.25;
+    double SCALE_FACTOR = 0.25;
+    double MAX_VALUE = 600;
+    double bucket_value = SCALE_FACTOR;
+
+    List<Double> rpc_millis_bucket_boundaries = new ArrayList<Double>(kMaxNumFiniteBuckets);
+    for (int i = 0; i < kMaxNumFiniteBuckets; i++) {
+      if (bucket_value > MAX_VALUE) {
+        break;
+      }
```
olavloite: nit: you could move this check into the for loop like this:

```suggestion
    for (int i = 0; i < kMaxNumFiniteBuckets && bucket_value <= MAX_VALUE; i++) {
      if (bucket_value > MAX_VALUE) {
        break;
      }
```

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
```diff
@@ -31,31 +30,42 @@ public class Metrics {
   static final String PGADAPTER_LATENCY_DESCRIPTION =
       "Latency between PGAdapter receiving a statement from the client and PGAdapter returning the last row of the response to the client";
 
-  private final LongHistogram spannerClientLibLatencies;
-  private final LongHistogram pgadapterLatencies;
+  private final DoubleHistogram spannerClientLibLatencies;
+  private final DoubleHistogram pgadapterLatencies;
+
+  public static List<Double> getMetricLatencyMillisBuckets() {
```
olavloite, line 37: nit: does this need to be public?

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
```diff
@@ -31,31 +30,42 @@ public class Metrics {
...
+    int kMaxNumFiniteBuckets = 50;
+    double BASE = 1.25;
+    double SCALE_FACTOR = 0.25;
+    double MAX_VALUE = 600;
```
olavloite: nit: these could (should) all be final

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
```diff
@@ -31,31 +30,42 @@ public class Metrics {
...
+    double bucket_value = SCALE_FACTOR;
```
olavloite: nit: variable names in Java should use camelCase

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
```diff
@@ -16,12 +16,11 @@
...
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.common.Attributes;
-import io.opentelemetry.api.metrics.LongHistogram;
+import io.opentelemetry.api.metrics.DoubleHistogram;
 import io.opentelemetry.api.metrics.Meter;
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.List;
 
-@InternalApi
```
olavloite, line 24: Is it intentional to remove this?

hengfengli: I forgot to mention this. This is intended because I need to call this in the tpcc benchmark code.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
hengfengli: Yes, we need this in the tpcc benchmark code - pgadapter/benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/Metrics.java.

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
olavloite: Yeah, but I think that we should still mark it as `@InternalApi`. The annotation is only informational and means that we reserve the right to make (breaking) changes without doing a major version update. So keeping it here won't prevent you from using it in the tpcc benchmark app, but it will tell other potential users that if they do use it, they have to accept that it might change without notice.

olavloite: (I took the liberty of adding the annotation again, we can always reopen the discussion if you feel that we really need to remove this)

olavloite (approved):

src/main/java/com/google/cloud/spanner/pgadapter/utils/Metrics.java
hengfengli: No problem at all. As long as the tpcc benchmark can use it, I am okay with it.

PR #3159, opened 2025-04-11
chore: move control message creation to separate class

olavloite:
Move the logic for creating a control message to a separate MessageReader class. This makes it easier to optimize this specific part of the code for specific use cases, e.g. introducing a busy-wait for messages in low-QPS scenarios.

This change only moves the code into a separate class, but does not make any changes.


hengfengli (approved):
LGTM.

PR #3236, opened 2025-04-30
fix: send UUID values as untyped strings to Spanner

olavloite:
Send UUID query parameter values as untyped string values to Spanner, so these can be used with both varchar and UUID columns. The latter is not yet supported on Spanner, so the tests currently only verify that they can be used with varchar.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/parsers/UuidParser.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/parsers/UuidParser.java
@@ -136,7 +136,8 @@ public class UuidParser extends Parser<UUID> {
   @Override
   public void bind(ImmutableMap.Builder<String, Value> parametersBuilder, String name) {
     // Send UUIDs to Spanner as untyped string values, so these can be used with both varchar and
-    // UUID columns.
+    // UUID columns. This ensures backwards compatibility, as PGAdapter would send UUID values as
+    // strings to Spanner before UUID type support was added to Spanner.
     parametersBuilder.put(
         name,
         this.item == null
--- a/src/main/java/com/google/cloud/spanner/pgadapter/parsers/UuidParser.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/parsers/UuidParser.java
@@ -22,6 +22,7 @@ import com.google.cloud.spanner.pgadapter.error.PGException;
 import com.google.cloud.spanner.pgadapter.error.SQLState;
 import com.google.cloud.spanner.pgadapter.error.Severity;
 import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.NullValue;
 import java.nio.charset.StandardCharsets;
 import java.util.UUID;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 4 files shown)


rahul2393 (approved):

src/main/java/com/google/cloud/spanner/pgadapter/parsers/UuidParser.java
```diff
@@ -131,6 +135,15 @@ public static byte[] convertToPG(ResultSet resultSet, int position, DataFormat f
 
   @Override
   public void bind(ImmutableMap.Builder<String, Value> parametersBuilder, String name) {
-    parametersBuilder.put(name, Value.uuid(this.item));
+    // Send UUIDs to Spanner as untyped string values, so these can be used with both varchar and
```
gagangupt16, line 138: I think we should mention the reason as backward compatibility for this behaviour in the comment. 
It will be instantaneously clear to a new person.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
```diff
@@ -4966,6 +4965,11 @@ public void testUUIDParameter() throws SQLException {
     }
 
     assertEquals(1, mockSpanner.countRequestsOfType(ExecuteSqlRequest.class));
+    ExecuteSqlRequest request = mockSpanner.getRequestsOfType(ExecuteSqlRequest.class).get(0);
+    assertEquals(1, request.getParams().getFieldsCount());
+    // UUIDs should be sent to Spanner as untyped values, so Spanner can infer the correct data
+    // type. This allows clients to use UUID query parameters with both STRING and UUID columns.
+    assertEquals(0, request.getParamTypesCount());
```
gagangupt16, line 4972: Should we have a test where we are binding UUID but querying string (varchar) column?

src/main/java/com/google/cloud/spanner/pgadapter/parsers/UuidParser.java
olavloite: Thanks for the tip, added it to the comment.

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcTest.java
```diff
@@ -1224,6 +1226,33 @@ public void testSelectNamespaces() throws SQLException {
     }
   }
 
+  @Test
+  public void testUUID() throws SQLException {
```
olavloite, line 1230: This verifies that query parameters that are given as UUIDs can be used with varchar columns.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcMockServerTest.java
olavloite: Yes, that is what this test does: https://github.com/GoogleCloudPlatform/pgadapter/pull/3236/files#r2069324656

It is better to add such a test to an end-to-end test, as this also verifies that Spanner accepts the untyped value. The mock server would always accept it.

olavloite:
The Golang sample test failure is unrelated to this change, tracking a fix for that in #3237

PR #3238, opened 2025-05-01
fix: encoded string too long could be returned

### LINKED ISSUE #3235: "encoded string too long" error on longer values in jsonb columns
Since the release of v0.47 this morning we started seeing the error `encoded string ({"name":...none"}}}) too long: xxx bytes` occurring when trying to read `jsonb` columns with longer values as part of `SELECT` queries.

It looks like a related change happened in this PR: https://github.com/GoogleCloudPlatform/pgadapter/pull/3096 since the error comes from good old `java.io.DataOutputStream.writeUTF()`, but that could well be a baseless accusation.

Rolling back to v0.46.2 mitigated the issue.

We are running pgadapter as a GKE sidecar, via the `pg` driver (`8.12.0`) and the unfortunately named Drizzle ORM (`0.31.1`).

olavloite:
Too long string or jsonb values could lead to the error "encoded string too long" to be returned.

This was introduced by the optimization in #3096, and this reverts part of that optimization.

Fixes #3235


olavloite:
`go-sample-test` failure is unrelated, and being fixed in #3240

PR #3275, opened 2025-05-08
feat: allow Spanner startup parameters in 'options=...'

olavloite:
PostgreSQL allows clients to set connection variables in the connection string by adding an `options=-c name=value` element to the connection string. PGAdapter also accepted these, but only for variables that could be set with a `SET` statement. Startup properties in the Spanner Connection API were not supported, which meant that properties like `options=-c enableEndToEndTracing=true` would have no effect.

This change adds support for setting startup parameters with the `options` argument in the PostgreSQL connection string.


hengfengli (approved):
LGTM.

PR #3310, opened 2025-05-15
fix: enable OpenTelemetry traces in Spanner

olavloite:
When OpenTelemetry has been enabled, it also needs to enable tracing using OpenTelemetry in Spanner for traces from the Spanner client library to be included.


hengfengli (approved):
LGTM

PR #3313, opened 2025-05-19
chore: added support to change the isolation level

shobhitsg:


benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JdbcBenchmarkRunner.java
```diff
@@ -29,6 +29,7 @@
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import org.apache.commons.lang3.StringUtils;
```
olavloite: nit: use `com.google.common.base.Strings` instead.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JdbcBenchmarkRunner.java
```diff
@@ -159,4 +165,13 @@ List<Object[]> executeParamQuery(String sql, Object[] params) throws SQLExceptio
     }
     return results;
   }
+
+  private int getTransactionIsolationLevel(String isolationLevel) {
+    return switch (isolationLevel) {
+      case SpannerConfiguration.SERIALIZABLE_ISOLATION_LEVEL -> Connection.TRANSACTION_SERIALIZABLE;
+      case SpannerConfiguration.REPEATABLE_READ_ISOLATION_LEVEL ->
+          Connection.TRANSACTION_REPEATABLE_READ;
+      default -> Connection.TRANSACTION_NONE;
```
olavloite: nit: It would be safer to return `Connection.TRANSACTION_SERIALIZABLE` here. This instructs the driver to not use transactions at all.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/config/SpannerConfiguration.java
```diff
@@ -19,9 +19,12 @@
 @Configuration
 @ConfigurationProperties(prefix = "spanner")
 public class SpannerConfiguration {
+  public static final String REPEATABLE_READ_ISOLATION_LEVEL = "REPEATABLE_READ";
+  public static final String SERIALIZABLE_ISOLATION_LEVEL = "SERIALIZABLE";
```
olavloite: nit: You don't need these string constants. These are equal to the `IsolationLevel` enum value names, so you can just use those instead.

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/config/SpannerConfiguration.java
```diff
@@ -19,9 +19,12 @@
...
   private String project;
   private String instance;
   private String database;
+  private String isolationLevel = SERIALIZABLE_ISOLATION_LEVEL;
```
olavloite:
```suggestion
  private String isolationLevel = IsolationLevel.SERIALIZABLE.name();
```

olavloite (approved):

benchmarks/tpcc/src/main/java/com/google/cloud/pgadapter/tpcc/JdbcBenchmarkRunner.java
shobhitsg: anything for `blank` strings? can't find anything in guava.

olavloite: I'm not sure what you mean (and whether this comment is still up to date). There is a `isNullOrEmpty` method in this package, and I see that this app is now using that.

shobhitsg: My intention with `isBlank` was to ignore the whitespace characters too.

PR #3370, opened 2025-05-28
chore: use IPv4 in samples

olavloite:
Force connections to test containers to use IPv4 to prevent random connection problems.

### DIFF
--- a/samples/dotnet/npgsql-sample/Program.cs
+++ b/samples/dotnet/npgsql-sample/Program.cs
@@ -26,7 +26,7 @@ internal static class Sample
         // Start PGAdapter and the Spanner emulator in a Docker container.
         var pgadapter = await StartPGAdapterAndEmulator();
         
-        var connectionString = $"Host=localhost;Port={pgadapter.GetMappedPublicPort(5432)};Database=my-database;SSL Mode=Disable";
+        var connectionString = $"Host=127.0.0.1;Port={pgadapter.GetMappedPublicPort(5432)};Database=my-database;SSL Mode=Disable";
         using var connection = new NpgsqlConnection(connectionString);
         connection.Open();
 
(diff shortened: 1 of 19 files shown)


samples/dotnet/npgsql-sample/Program.cs
```diff
@@ -26,7 +26,7 @@ private static async Task Main(string[] args)
         // Start PGAdapter and the Spanner emulator in a Docker container.
         var pgadapter = await StartPGAdapterAndEmulator();
         
-        var connectionString = $"Host=localhost;Port={pgadapter.GetMappedPublicPort(5432)};Database=my-database;SSL Mode=Disable";
+        var connectionString = $"Host=127.0.0.1;Port={pgadapter.GetMappedPublicPort(5432)};Database=my-database;SSL Mode=Disable";
```
aakashanandg, line 29: Does this mean that during DNS resolution localhost is getting mapped to IPV6 and since the PGAdapter is configured on IPV4 this connection might be breaking? But i am curious on why this would happen?

aakashanandg (approved):

samples/dotnet/npgsql-sample/Program.cs
olavloite: What can happen is:
1. Docker is asked to expose port 5432 on a random port on the host.
2. It is possible that Docker then assigns two different port numbers for IPv4 and IPv6.
3. The `GetMappedPublicPort(5432)` method returns the port number for IPv4.
4. `localhost` can be translated to both `127.0.0.1` (IPv4) or `::1` (IPv6). Which one is up to the host.
5. So if both of the following happens, then we get an error when running the sample:
5.1. Docker assigns two different ports for IPv4 and IPv6
5.2. `localhost` is translated to `::1`.
5.3. The sample application fails, because it tries to use the port number that was assigned to IPv4, but `localhost` is translated to IPv6 `::1`.

PR #3384, opened 2025-05-29
fix: add metadata query replacement for JDBC 42.7.6

olavloite:
Adds a query replacement for a slightly modified metadata query in version 42.7.6 of the PostgreSQL JDBC driver.


src/main/java/com/google/cloud/spanner/pgadapter/statements/JdbcMetadataStatementHelper.java
```diff
@@ -81,7 +81,8 @@ static String replaceJdbcMetadataStatement(String sql) {
         return replaceGetTablesQuery(sql, PgJdbcCatalog.PG_JDBC_GET_TABLES_REPLACEMENT);
       }
     }
-    if (sql.startsWith(PgJdbcCatalog.PG_JDBC_GET_COLUMNS_PREFIX_V42_7_5)
+    if (sql.startsWith(PgJdbcCatalog.PG_JDBC_GET_COLUMNS_PREFIX_V42_7_6)
```
aakashanandg, line 84: Can we add tests to verify that these queries are correctly transformed into spangres-compatible queries?

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcMetadataTest.java
```diff
@@ -53,12 +53,12 @@
 public class ITJdbcMetadataTest implements IntegrationTest {
   private static final String[] VERSIONS =
       new String[] {
-        "42.7.5", "42.7.4", "42.7.3", "42.7.2", "42.7.1", "42.7.0", "42.6.0", "42.5.4", "42.5.3",
-        "42.5.2", "42.5.1", "42.5.0", "42.4.2", "42.4.1", "42.4.0", "42.3.6", "42.3.5", "42.3.4",
-        "42.3.3", "42.3.2", "42.3.1", "42.3.0", "42.2.25", "42.2.24", "42.2.23", "42.2.22",
-        "42.2.21", "42.2.20", "42.2.19", "42.2.18", "42.2.17", "42.2.16", "42.2.15", "42.2.14",
-        "42.2.13", "42.2.12", "42.2.11", "42.2.10", "42.2.9", "42.2.8", "42.2.7", "42.2.6",
-        "42.2.5", "42.2.4", "42.2.3", "42.2.2", "42.2.1", "42.2.0", "42.1.4"
+        "42.7.6", "42.7.5", "42.7.4", "42.7.3", "42.7.2", "42.7.1", "42.7.0", "42.6.0", "42.5.4",
```
olavloite, line 56: This test invokes all JDBC metadata methods, which use the queries that are translated in JdbcMetadataStatementHelper.

GitHub makes a bit of a mess of this change. The only real diff here is that `42.7.6` is added to the list of versions that is tested.

src/main/java/com/google/cloud/spanner/pgadapter/statements/JdbcMetadataStatementHelper.java
olavloite: The ITJdbcMetadataTest does that. It invokes all JDBC metadata methods, which again use these queries. Without these replacements (or if the replacements were invalid), then those tests would fail.

You can see this

src/test/java/com/google/cloud/spanner/pgadapter/ITJdbcMetadataTest.java
aakashanandg: Thanks for the clarification.

aakashanandg (approved):

PR #3429, opened 2025-06-10
fix: query cancelled was returned as a generic Interrupted error for DML

olavloite:
Cancelled DML statements were not translated to `Query cancelled` errors with the `57014` SQLState. Instead, they were returned as generic errors with the message `ERROR: Interrupted`.

This change makes sure that:
1. The correct error code and message is returned for both DML and queries, both in transactions and in auto-commit.
2. The interrupted flag of the connection thread is cleared.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
@@ -182,6 +182,28 @@ public class BackendConnection {
     return pgException;
   }
 
+  static boolean isQueryCancelled(SpannerException spannerException) {
+    if (Thread.interrupted()) {
+      return true;
+    }
+    if (spannerException.getErrorCode() == ErrorCode.CANCELLED) {
+      return true;
+    }
+    final int maxDepth = 100;
+    Throwable cause = spannerException.getCause();
+    Throwable prevCause = null;
+    int depth = 0;
+    while (cause != null && cause != prevCause && depth < maxDepth) {
+      if (cause instanceof InterruptedException) {
+        return true;
+      }
+      prevCause = cause;
+      cause = cause.getCause();
+      depth++;
+    }
+    return false;
+  }
+
   boolean shouldReplaceStatement(Statement statement) {
     if (!localStatements.get().isEmpty() && localStatements.get().containsKey(statement.getSql())) {
       LocalStatement localStatement = localStatements.get().get(statement.getSql());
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
```diff
@@ -182,6 +182,28 @@ static <T> PGException setAndReturn(SettableFuture<T> future, Throwable throwabl
...
   }
 
+  static boolean isQueryCancelled(SpannerException spannerException) {
+    if (Thread.interrupted()) {
+      return true;
+    }
+    if (spannerException.getErrorCode() == ErrorCode.CANCELLED) {
+      return true;
+    }
+    final int maxDepth = 100;
```
aakashanandg, line 192: Instead of hardcoding this to a default value can't we just iterate till a `null` is encountered? Or are sure that the maximum chained exception length can't exceed 100 in this case?

olavloite: The reason for this magic '100' number here is that in theory we could end up in an eternal loop, if the chain of error causes happen to contain a loop. This is covered in the test case that starts here: https://github.com/GoogleCloudPlatform/pgadapter/blob/f1a4977d05fe896cb2ed1543895e678d126a85bf/src/test/java/com/google/cloud/spanner/pgadapter/statements/BackendConnectionTest.java#L1033

There is no guarantee for how deep the error chain could be, but normally it is between 1 and 5 levels, so 100 seems like a relatively safe limit to just stop searching.

The worst that could happen if the error chain contains more than 100 nested causes, is that the error is returned as a generic error, instead of a specific `Query cancelled` error.

src/test/java/com/google/cloud/spanner/pgadapter/statements/BackendConnectionTest.java
```diff
@@ -993,4 +994,50 @@ public void testExecuteAndVacuum_startsTransaction() {
...
+
+    // A circular set of causes should not lead to an eternal loop.
+    Exception e1 = new Exception("test");
+    Exception e2 = new Exception("test", e1);
+    Exception e3 = new Exception("test", e2);
+    Exception e4 = new Exception("test", e3);
+    e1.initCause(e4);
+    SpannerException spannerException =
+        SpannerExceptionFactory.newSpannerException(ErrorCode.UNKNOWN, "test", e4);
+    assertFalse(isQueryCancelled(spannerException));
```
aakashanandg, line 1041: We should also test for chained exceptions where the root cause is an `InterruptedException` to confirm the error is handled as a query cancellation.

olavloite: The method itself only accepts a `SpannerException` as input argument, not an `InterruptedException`. This is because the Spanner client library always throws `SpannerException`s and nothing else.

The first line of this test verifies the behavior where the first exception in the chain below the `SpannerException` is an `InterruptedException`.

aakashanandg: Thanks for the clarification.

aakashanandg (approved):

PR #3741, opened 2025-09-16
[chore] add pg_regress workflow

hengfengli:
Set up the pg_regress test for emulator and prod (disabled for now). It is scheduled to run daily and the workflow will fail if the number of passed cases is less than previous runs.


.github/workflows/pg-regress-emulator.yml
```diff
@@ -0,0 +1,91 @@
...
+        uses: actions/checkout@v5
+        with:
+          path: pgadapter
+      - name: Checkout OSS PG repo
+        uses: actions/checkout@v5
+        with:
+          repository: postgres/postgres
+          path: postgres
+          ref: REL_16_0
+      - id: 'auth'
```
olavloite, line 29: This shouldn't be necessary if we are only running the tests on the Emulator, right?

olavloite (approved):
LGTM, but I think that we can remove the authentication step when running on the emulator.

.github/workflows/pg-regress-emulator.yml
hengfengli: I need this because we need to call `gcloud storage` command to download the previous json result file for comparison and upload the result file if passes.

PR #3791, opened 2025-09-24
chore: add support to always upload result files for pg_regress

hengfengli:
It is easier to debug if we always upload the result files even for failures. And we only update the latest file when necessary.

### DIFF
--- a/.github/workflows/pg-regress-emulator.yml
+++ b/.github/workflows/pg-regress-emulator.yml
@@ -74,6 +74,12 @@ jobs:
           else
             echo "has_latest_file=false" >> $GITHUB_OUTPUT
           fi
+      - name: Upload results
+        working-directory: ./postgres/src/test/regress/
+        run: |
+          ts=$(date +%s)
+          gcloud storage cp results.json ${{env.GCS_BUCKET_PATH}}/results_$ts.json
+          gcloud storage cp regression.diffs ${{env.GCS_BUCKET_PATH}}/regression_$ts.diffs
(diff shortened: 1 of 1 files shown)


.github/workflows/pg-regress-emulator.yml
```diff
@@ -84,10 +90,4 @@ jobs:
             echo "Fail: $curr_passed is less than $prev_passed. Please check!"
             exit 1
           fi
-      - name: Upload results
-        working-directory: ./postgres/src/test/regress/
-        run: |
-          ts=$(date +%s)
-          gcloud storage cp results.json ${{env.GCS_BUCKET_PATH}}/results_$ts.json
           gcloud storage cp results.json ${{env.GCS_BUCKET_PATH}}/results_latest.json
```
olavloite, line 93: I assume that the reason that this step stays here is that we basically want 'latest' to mean 'best result so far'? Meaning: We only update 'latest' if the results were at least as good as the previous 'latest', which again means that 'latest' could in theory contain results from many months back if we have a long-term regression. Is that intentional?

olavloite (approved):
(Approving, assuming that the way that 'latest' is handled is intentional)

.github/workflows/pg-regress-emulator.yml
hengfengli: Yes, your understanding is correct. We would like to detect regressions so that we need a place to store the best result before. Maybe `latest` is a bit confusing but I'll leave it for now.

PR #3894, opened 2025-10-31
build: tag old Docker images when releasing new ones

olavloite:
Add a 'no-new-use-public-image-' tag to all existing Docker images that do not already have such a tag when new images are being released.


olavloite:
(The Ruby samples build failure is unrelated to this change)

rayudu3745 (approved):

PR #3902, opened 2025-11-05
feat: support for wire protocol 3.2

### LINKED ISSUE #3720: Support wire-protocol version 3.2
Support the PostgreSQL wire-protocol version 3.2 that is being introduced in PostgreSQL 18: https://www.postgresql.org/docs/18/protocol-message-formats.html

rayudu3745:
wire protocol 3.2 added support for wider cancellation keys.
Previously cancel keys were 4 bytes now they can have a maximum of 256 bytes but currently the PGadapter sends 32 byte wide cancel keys

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
@@ -82,6 +82,7 @@ import java.security.SecureRandom;
 import java.text.MessageFormat;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -128,6 +129,9 @@ public class ConnectionHandler implements Runnable {
   private Thread thread;
   private final int connectionId;
   private final int secret;
+  private final byte[] secretBytes;
+  public static final int DEFAULT_KEY_LENGTH_BYTES = 32;
+
   // Separate the following from the threat ID generator, since PG connection IDs are maximum
   //  32 bytes, and shouldn't be incremented on failed startups.
   private static final AtomicInteger incrementingConnectionId = new AtomicInteger(0);
@@ -142,6 +146,7 @@ public class ConnectionHandler implements Runnable {
   private DatabaseId databaseId;
   private WellKnownClient wellKnownClient = WellKnownClient.UNSPECIFIED;
   private boolean hasDeterminedClientUsingQuery;
+  private int protocolVersion;
 
   /**
    * List of PARSE messages that we received before auto-detecting the client. This list can be used
@@ -162,7 +167,12 @@ public class ConnectionHandler implements Runnable {
   ConnectionHandler(ProxyServer server, Socket socket, Connection spannerConnection) {
     this.server = server;
     this.socket = socket;
-    this.secret = new SecureRandom().nextInt();
+
+    SecureRandom secureRandom = new SecureRandom();
+    this.secret = secureRandom.nextInt();
+    this.secretBytes = new byte[DEFAULT_KEY_LENGTH_BYTES];
+    secureRandom.nextBytes(this.secretBytes);
+
     this.connectionId = incrementingConnectionId.incrementAndGet();
     this.spannerConnection = spannerConnection;
   }
@@ -718,6 +728,8 @@ public class ConnectionHandler implements Runnable {
    * specific connection identified by connectionId. Since cancellation is a flimsy contract at
    * best, it is not imperative that the cancellation run, but it should be attempted nonetheless.
    *
+   * <p>This method is for protocol 3.0 which uses integer secrets.
+   *
    * @param connectionId The connection whose statement must be cancelled.
    * @param secret The secret value linked to the connection that is being cancelled. If it does not
    *     match, we cannot cancel.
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -127,7 +129,11 @@ public class ConnectionHandler implements Runnable {
   private volatile ConnectionStatus status = ConnectionStatus.UNAUTHENTICATED;
   private Thread thread;
   private final int connectionId;
-  private final int secret;
+  private int protocolVersion;
+  private byte[] secret;
+  public static final int DEFAULT_KEY_LENGTH_BYTES = 32;
+  public static final int PROTOCOL_3_0_KEY_LENGTH_BYTES = 4;
```
olavloite, line 135: nit: could we move these to the StartupMessage class file?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -740,7 +745,23 @@ public boolean cancelActiveStatement(int connectionId, int secret) {
...
+            SQLState.ProtocolViolation);
+      }
+    }
+
+    if (StartupMessage.PROTOCOL_VERSION_3_2_IDENTIFIER == connectionToCancel.getProtocolVersion()) {
+      if (secret.length != DEFAULT_KEY_LENGTH_BYTES) {
+        throw PGExceptionFactory.newPGException(
+            "protocol version 3.2 accepts 32 bytes for the secret ", SQLState.ProtocolViolation);
+      }
+    }
```
olavloite: Can we change this into either an `if-else if-else` block, or a `switch` block, and also throw an error if the protocol version is neither of the two expected values?

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CancelMessage.java
```diff
@@ -26,16 +27,17 @@
...
 
   private final int connectionId;
-  private final int secret;
+  private final byte[] secret;
 
-  public CancelMessage(ConnectionHandler connection) throws Exception {
-    super(connection, MESSAGE_LENGTH);
+  public CancelMessage(ConnectionHandler connection, int length) throws Exception {
+    super(connection, length);
+    int secretLen = length - 12;
```
olavloite, line 37: Is the secret length in the new protocol variable? Or is it fixed at 256 bytes, and is this only here because of the difference between protocol 3.0 and 3.2?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -127,7 +129,11 @@ public class ConnectionHandler implements Runnable {
   private volatile ConnectionStatus status = ConnectionStatus.UNAUTHENTICATED;
   private Thread thread;
   private final int connectionId;
-  private final int secret;
+  private int protocolVersion;
+  private byte[] secret;
+  public static final int DEFAULT_KEY_LENGTH_BYTES = 32;
```
olavloite, line 134: This seems to indicate that 32 bytes is the __default__. Meaning that in 3.2 the client could also use for example 16 bytes. Is that correct?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -740,7 +745,23 @@ public boolean cancelActiveStatement(int connectionId, int secret) {
...
+    if (StartupMessage.PROTOCOL_VERSION_3_0_IDENTIFIER == connectionToCancel.getProtocolVersion()) {
+      if (secret.length != PROTOCOL_3_0_KEY_LENGTH_BYTES) {
+        throw PGExceptionFactory.newPGException(
+            "protocol version 3.0 accepts only 4 bytes for the secret ",
+            SQLState.ProtocolViolation);
+      }
+    }
+
+    if (StartupMessage.PROTOCOL_VERSION_3_2_IDENTIFIER == connectionToCancel.getProtocolVersion()) {
+      if (secret.length != DEFAULT_KEY_LENGTH_BYTES) {
```
olavloite: See also the questions above: Is the 32 bytes a __default__? Or a __fixed__ value? If it is fixed; could we change the name of the constant to `PROTOCOL_3_2_KEY_LENGTH_BYTES`?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -894,6 +917,21 @@ public WellKnownClient getWellKnownClient() {
...
+  public int getProtocolVersion() {
+    return protocolVersion;
+  }
+
+  public void setProtocolVersionProperties(int protocolVersion) {
+    this.protocolVersion = protocolVersion;
+    int secretLen =
+        StartupMessage.PROTOCOL_VERSION_3_0_IDENTIFIER == protocolVersion
+            ? ConnectionHandler.PROTOCOL_3_0_KEY_LENGTH_BYTES
+            : ConnectionHandler.DEFAULT_KEY_LENGTH_BYTES;
```
olavloite, line 934: See also my other questions: If this is just a default, and the client can request any length between X and 256 bytes in the startup message, then we need to make the assignment variable here as well.

rayudu3745: client cannot request for the length of the secret , just that it can of any size between 4 and 256 bytes, 
we use 32 bytes for for 3.2 as postgres is also using the same length.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
rayudu3745: default in the sense , future versions will also have the same length except the 3.0 which is 4 bytes

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
rayudu3745: clients can't request for the length of the secret , they should be able to handle the key which can be of 4 to 256 bytes from 3.2
default naming is used as future versions can also use same length

src/main/java/com/google/cloud/spanner/pgadapter/wireprotocol/CancelMessage.java
rayudu3745: its fixed when we generate the key 
from 3.2 client can send secret  from 4 to 256 bytes , to extract the secret sent by client we calculate the length and verify it at later step

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
rayudu3745: Since the secret exists here, i thought this would be a good place

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
rayudu3745: will do the change

olavloite (approved):

PR #4098, opened 2026-01-14
chore: update experimental host config

sagnghos:
Previously the change was made in https://github.com/GoogleCloudPlatform/pgadapter/pull/2856 to distinguish between experimental host endpoint and cloud spanner endpoint using url pattern matching. Now we have a isExperimentalHost parameter which is much more deterministic in determining whether its a experimental host endpoint.

The PR also contains changes to make the Integration tests compatible with experimental host


olavloite (approved):

olavloite:
(The build error is unrelated and seems to be a bug in the emulator)

PR #4167, opened 2026-03-03
fix: use id, author and filename as pk for Liquibase databasechangelog table

molivasdat:
Users of Spanner using the PgAdapter plugin cannot reuse the same id anywhere within all of the changelogs.
That means that users of Liquibase cannot create a changeset within the same file or even different files if different authors use the same id.
This PR will fix that issue.
From the liquibase docs
https://docs.liquibase.com/secure/user-guide-5-1/what-is-the-databasechangelog-table

What is the **DATABASECHANGELOG** table?

Liquibase tracks which changesets have been run by using the DATABASECHANGELOG (DBCL) table. If the table does not exist in the database, Liquibase creates one automatically.

The default name of the DBCL is DATABASECHANGELOG. You can specify a different name with the --[database-changelog-table-name](https://docs.liquibase.com/secure/reference-guide-5-0/parameters/database-changelog-lock-table-name) parameter.
**Columns**

The table tracks each changeset as a row, identified by a combination of the id, author, and filename columns. **There is no primary key on the table. This is to avoid any database-specific restrictions on key lengths.** The composite of id, author, and filename is unique across all rows of the table.

Users of Liquibase can create changes and those changes are indexed by 
filename, id, and author. 
If a primary key on the table is required in Spanner when using it as a PostgreSQL database,then this change would fulfill the need for a composite key entry.


olavloite:
@molivasdat Thanks for this fix. We can only merge this fix if you sign the CLA (click on the failed cla/google check above for more details). I can also move this change to a new pull request, but I would prefer to merge your PR to make sure that you are listed as the contributor for this fix.

molivasdat:
@olavloite I did sign it.  My email address is private in the commit id.  I have signed the CLA.

There is this in the CLA docs.
You may have [Keep my email address private](https://github.com/settings/emails#:~:text=Keep%20my%20email%20addresses%20private) enabled. Without a visible email address, the CLA cannot be checked. Uncheck it and re-create the offending commit, or have your CLA point of contact add your @users.noreply.github.com address to the CLA group.

Not sure who my CLA point of contact would be to add 
The following contributors were found for this pull request:

❌ https://github.com/GoogleCloudPlatform/pgadapter/commit/9edbe336b0d4136e002fdb22d2dfaf98ddaac762 Author: @molivasdat <47544147+molivasdat​@users.noreply.github.com>

(Only the first commit for a unique contributor is listed.)

olavloite:
@molivasdat Ah, yes, re-running the check manually seems to succeed. Not sure why it did not turn green automatically. Thanks!

olavloite (approved):

PR #4257, opened 2026-03-23
chore: add a test file in pg-regress

hengfengli:
Also, fix a code issue that `bool` is a reserved keyword in the C23 standard.


olavloite (approved):
(I did not read all that, but I assume that this is good....)

PR #4275, opened 2026-03-31
fix: support COPY for FLOAT8/FLOAT4 NaN/Infinity

skuruppu:
Previously, an error was returned in the form "google.protobuf.Value cannot encode double values for infinity or nan, because they would be parsed as a string."

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/utils/CsvCopyParser.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/utils/CsvCopyParser.java
@@ -132,6 +132,22 @@ class CsvCopyParser implements CopyInParser {
       return getSpannerValue(sessionState, type, recordValue);
     }
 
+    private static Value getNonFiniteValue(String recordValue) {
+      if ("NaN".equalsIgnoreCase(recordValue)) {
+        return Value.string("NaN");
+      }
+      if ("Infinity".equalsIgnoreCase(recordValue)
+          || "+Infinity".equalsIgnoreCase(recordValue)
+          || "inf".equalsIgnoreCase(recordValue)
+          || "+inf".equalsIgnoreCase(recordValue)) {
+        return Value.string("Infinity");
+      }
+      if ("-Infinity".equalsIgnoreCase(recordValue) || "-inf".equalsIgnoreCase(recordValue)) {
+        return Value.string("-Infinity");
+      }
+      return null;
+    }
+
     static Value getSpannerValue(SessionState sessionState, Type type, String recordValue)
         throws SpannerException {
       try {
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/utils/CsvCopyParser.java
```diff
@@ -145,9 +145,25 @@ static Value getSpannerValue(SessionState sessionState, Type type, String record
           case INT64:
             return Value.int64(recordValue == null ? null : Long.parseLong(recordValue));
           case FLOAT32:
-            return Value.float32(recordValue == null ? null : Float.parseFloat(recordValue));
+            if (recordValue == null) {
+              return Value.float32(null);
+            }
+            if ("NaN".equals(recordValue)
```
rayudu3745: nit: should we use equalsIgnoreCase if postgres is case insensitive for these values ?

rayudu3745 (approved):

src/main/java/com/google/cloud/spanner/pgadapter/utils/CsvCopyParser.java
olavloite: Yes. We actually need a bit more flexible parsing of this (Inf, +Inf, inf, INF, Infinity, etc. are all valid). And I think that there is one more place where this goes wrong (parameter binding), so we should probably fix that in the same go. Let me try to extend this fix a bit.

skuruppu: Thanks both. Now updated to support all the possible non-finite values in PG.

In terms of the parameter binding, are we talking about DoubleParser.java, FloatParser.java logic? Sorry, I'm not very familiar with this repo.

olavloite: > In terms of the parameter binding, are we talking about DoubleParser.java, FloatParser.java logic?

Yes. And I think that is also the more appropriate place for this code to live, as it is also needed there and not only in CsvCopyParser. So I took the liberty of moving it there and applying it to the `bind(..)` methods in the respective parsers.

olavloite (approved):

src/main/java/com/google/cloud/spanner/pgadapter/utils/CsvCopyParser.java
olavloite: @rayudu3745 @skuruppu Would you mind taking a look at https://github.com/GoogleCloudPlatform/pgadapter/pull/4275/commits/ddc0a1147355a5886c4709cd18070e7b87d5718e and check if you are OK with that change as well?

rayudu3745 (approved):

Issue #4276, opened 2026-03-31
Add support for Python ADBC driver


olavloite:
Add support for Python ADBC driver. This requires re-writing a few startup queries that check which types are available in the database.

PR #4277, opened 2026-03-31
chore: emulate oid data type with hash function

olavloite:
Use the farm_fingerprint function to emulate the oid of tables, instead of using the table name directly. This ensures that the data type that is being returned is an int, which is much more like the original oid data type.

Towards #4276


src/main/java/com/google/cloud/spanner/pgadapter/statements/PgCatalog.java
```diff
@@ -788,17 +778,17 @@ public class PgConstraint implements PgCatalogTable {
     public static final String PG_CONSTRAINT_CTE =
         "pg_constraint as (\n"
             + "select\n"
-            + "    '''\"' || tc.constraint_schema || '\".\"' || tc.constraint_name || '\"''' as oid,\n"
+            + "    mod(abs(spanner.farm_fingerprint(tc.constraint_schema || '.' || tc.constraint_name)), 2147483648) as oid,\n"
```
rayudu3745, line 781: any reason we are not including table_name also along with constraint_schema and constraint_name for oid generation here ?

olavloite: In this change: Because it was not in the original expression. In this change, we are only changing the generation expressions from returning strings to returning a hash of the strings.

Originally: Because constraint names are unique within a schema in Spanner, even in the PostgreSQL dialect. So we do not need to include the table name in order to guarantee that we generate a unique name. See https://docs.cloud.google.com/spanner/docs/check-constraint/how-to#postgresql:

> Constraint names are scoped to the schema, along with the names for tables and indexes, and must be unique within the schema.

rayudu3745 (approved):

PR #4315, opened 2026-04-09
fix: emit variant-specific command tag for DEALLOCATE ALL and DISCARD

praboud-ant:
## Problem
When processing `DEALLOCATE ALL`, PGAdapter replies with `DEALLOCATE` - unlike a real postgres, which replies with `DEALLOCATE ALL`. A similar problem exists for `DISCARD`.

This matters for clients that validate the response against the expected string; for some clients (in particular: pgbouncer), it causes the client to think that PGAdapter has not actually processed the `DEALLOCATE`, which causes incorrect client behavior later on.

## Fix
Emit the full tag on each response, e.g: `DEALLOCATE ALL`, rather than `DEALLOCATE`.

## Tests
New unit tests to validate this behavior.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/statements/DeallocateStatement.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/statements/DeallocateStatement.java
@@ -64,7 +64,7 @@ public class DeallocateStatement extends IntermediatePortalStatement {
 
   @Override
   public String getCommandTag() {
-    return "DEALLOCATE";
+    return deallocateStatement.name == null ? "DEALLOCATE ALL" : "DEALLOCATE";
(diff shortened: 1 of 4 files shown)


olavloite:
@praboud-ant This looks generally good to me. Would you mind signing the CLA? Without it, it will not be possible to merge this pull request.

praboud-ant:
@olavloite yeah, sorry - I'm just waiting on our legal team to add me to the existing corporate CLA. I'll let you know when that goes through.

praboud-ant:
@olavloite I should be in the corporate CLA now - not sure if you need to boop the build step for it to see that.

olavloite (approved):

PR #4340, opened 2026-04-16
perf: move cache used for described statements from connection to server

olavloite:
Moves the cache that is used for described statements from the individual connections to the ProxyServer, and use the database name as a prefix for the cache entries. This improves cache usage for applications that do not use connection pooling, or for other reasons spread their statements across a large number of different connections.

This change also adds options for setting the cache size and expiration time.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
@@ -69,8 +69,6 @@ import com.google.cloud.spanner.pgadapter.wireprotocol.WireMessage;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableList;
 import com.google.spanner.admin.database.v1.InstanceName;
 import com.google.spanner.v1.DatabaseName;
@@ -82,7 +80,6 @@ import java.net.Socket;
 import java.net.SocketException;
 import java.security.SecureRandom;
 import java.text.MessageFormat;
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -818,12 +811,18 @@ public void registerStatement(String statementName, IntermediatePreparedStatemen
    * in the cache.
    */
   public Future<DescribeResult> getAutoDescribedStatement(String sql) {
-    return this.autoDescribedStatementsCache.getIfPresent(sql);
+    if (this.databaseName == null) {
```
rayudu3745, line 814: why would we need a null check here ? can it be possible for database be `null`?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
```diff
@@ -818,12 +811,18 @@ public void registerStatement(String statementName, IntermediatePreparedStatemen
...
+      return null;
+    }
+    return this.server.autoDescribedStatementsCache.getIfPresent(this.databaseName + sql);
   }
 
   /** Stores the parameter types of an auto-described statement in the cache. */
   public void registerAutoDescribedStatement(String sql, Future<DescribeResult> describeResult) {
-    this.autoDescribedStatementsCache.put(sql, describeResult);
+    if (this.databaseName == null) {
```
rayudu3745, line 822: why would we need a null check here ? can it be possible for database be `null`?

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
```diff
@@ -171,6 +178,13 @@ public ProxyServer(
       OptionsMetadata optionsMetadata, OpenTelemetry openTelemetry, Properties properties) {
     this.options = optionsMetadata;
     this.messageReader = new MessageReader(optionsMetadata);
+    this.autoDescribedStatementsCache =
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(Duration.ofMinutes(optionsMetadata.getDescribeCacheExpireMinutes()))
+            .maximumSize(optionsMetadata.getDescribeCacheMaxSize())
+            .concurrencyLevel(1)
```
rayudu3745, line 185: any reason for limiting concurrencyLevel to 1 ?

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: Pure theoretically, yes, but in reality no. It is null until the connection to Spanner has been established. And before that, it is not possible to execute any queries, so when this method is actually being called, then it will always be non-null, but code analyzers do not know that. So to prevent warnings, I added the null check here (and also below).

src/main/java/com/google/cloud/spanner/pgadapter/ProxyServer.java
olavloite: Yes, the vast majority of access to this cache are reads, as it only contains parameterized statements. Those are normally re-used in an application, meaning that they will normally also only be added once to the cache. Increasing the concurrency level would lead to higher storage cost, which I don't think in this case is justified, due to the very low expected gain from the added concurrency level.

src/main/java/com/google/cloud/spanner/pgadapter/ConnectionHandler.java
olavloite: See above

rayudu3745 (approved):

PR #4389, opened 2026-04-25
chore: add mock-server based benchmarks

olavloite:
Adds a simple benchmark class that uses a mock server. These benchmarks are easy to run and can be used to verify optimizations in PGAdapter code, for example to reduce GC pressure or reduce CPU time that is spent in PGAdapter.

It is not intended for benchmarking end-to-end latency to Spanner. For that, a benchmark that connects to Spanner should be used.

### DIFF
--- a/src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
+++ b/src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
@@ -1,0 +1,403 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.cloud.spanner.pgadapter.benchmark;
+
+import com.google.cloud.spanner.pgadapter.AbstractMockServerTest;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.concurrent.TimeUnit;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.profile.GCProfiler;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+/**
+ * Benchmarks for PGAdapter using the in-process mock Spanner server. This class is named without
+ * the 'Test' suffix to prevent it from running automatically during standard unit test execution.
+ */
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+public class MockServerBenchmark extends AbstractMockServerTest {
+
+  private Connection connection;
+
+  private static com.google.cloud.spanner.MockSpannerServiceImpl
+      createMockSpannerThatReturnsOneQueryPartition() {
+    return new com.google.cloud.spanner.MockSpannerServiceImpl() {
+      @Override
+      public void partitionQuery(
+          com.google.spanner.v1.PartitionQueryRequest request,
+          io.grpc.stub.StreamObserver<com.google.spanner.v1.PartitionResponse> responseObserver) {
+        responseObserver.onNext(
+            com.google.spanner.v1.PartitionResponse.newBuilder()
+                .addPartitions(
+                    com.google.spanner.v1.Partition.newBuilder()
+                        .setPartitionToken(com.google.protobuf.ByteString.EMPTY)
+                        .build())
+                .build());
+        responseObserver.onCompleted();
+      }
+    };
+  }
+
+  private static com.google.spanner.v1.ResultSetMetadata createNonArrayTypesResultSetMetadata() {
+    return com.google.spanner.v1.ResultSetMetadata.newBuilder()
+        .setRowType(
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/cloud/spanner/pgadapter/benchmark/SingleThreadedMockServerBenchmark.java
```diff
@@ -0,0 +1,403 @@
...
+/**
+ * Benchmarks for PGAdapter using the in-process mock Spanner server. This class is named without
+ * the 'Test' suffix to prevent it from running automatically during standard unit test execution.
+ */
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+public class MockServerBenchmark extends AbstractMockServerTest {
+
+  private Connection connection;
```
rayudu3745, line 56: this connection is shared across all threads if the benchmark is run with multiple threads in the future.
is this benchmark not intended for testing with multiple threads ?

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/SingleThreadedMockServerBenchmark.java
```diff
@@ -0,0 +1,403 @@
...
+
+  @TearDown(Level.Trial)
+  public void teardown() throws Exception {
+    if (connection != null) {
+      connection.close();
+    }
+    stopMockSpannerAndPgAdapterServers();
+  }
+
+  @State(Scope.Benchmark)
```
rayudu3745, line 296: should we use thread scope for running with multiple threads ?

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
```diff
@@ -0,0 +1,403 @@
...
+
+  @Benchmark
+  public void testSelectOneRowWithParam(ParameterState state) throws SQLException {
+    try (java.sql.PreparedStatement statement =
+        connection.prepareStatement("SELECT * FROM all_types WHERE col_bigint = ?")) {
+      statement.setLong(1, state.nextValue());
+      try (ResultSet resultSet = statement.executeQuery()) {
+        while (resultSet.next()) {
+          // Consume result set
+          resultSet.getLong(1);
```
rayudu3745: if the JIT compiler detects that the return value of a method `resultSet.getLong(1)` is never used, it can optimize away that line.
we should consider using JMH's `Blackhole` to consume the result here

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
```diff
@@ -0,0 +1,403 @@
...
+      }
+    }
+  }
+
+  @Benchmark
+  public void testSelect1() throws SQLException {
+    try (ResultSet resultSet = connection.createStatement().executeQuery("SELECT 1")) {
+      while (resultSet.next()) {
+        // Consume result set
+        resultSet.getLong(1);
```
rayudu3745: same as above

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+  }
+
+  @Benchmark
+  public void testSelectFiveRows() throws SQLException {
+    try (ResultSet resultSet =
+        connection.createStatement().executeQuery("SELECT * FROM TableWithFiveRows")) {
+      while (resultSet.next()) {
+        // Consume result set
+        resultSet.getString(1);
```
rayudu3745: same as above

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+  }
+
+  @Benchmark
+  public void testSelectLargeResultSet() throws SQLException {
+    try (ResultSet resultSet =
+        connection.createStatement().executeQuery("SELECT * FROM large_table")) {
+      while (resultSet.next()) {
+        // Consume result set
+        resultSet.getString(1);
```
rayudu3745: same as above

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
```diff
@@ -0,0 +1,403 @@
...
+  @Benchmark
+  public void testReadWriteTransaction() throws SQLException {
+    connection.setAutoCommit(false);
+    try {
+      try (java.sql.PreparedStatement statement =
+          connection.prepareStatement("SELECT * FROM table1 WHERE id = ?")) {
+        statement.setLong(1, 1L);
+        try (ResultSet resultSet = statement.executeQuery()) {
+          while (resultSet.next()) {
+            /* consume */
```
rayudu3745: we should use `Blackhole` to consume here

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/SingleThreadedMockServerBenchmark.java
olavloite: This benchmark is intentionally single-threaded, as its main purpose is to measure potential improvements in CPU or memory usage (especially memory allocations). For that, simple single-threaded runs are normally enough. If we need multi-threaded benchmarking in the future, it would probably be better to create a new subclass of this class (or in some other way share some of the base code).

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/SingleThreadedMockServerBenchmark.java
olavloite: See above. I also added some additional comments and renamed the class to make clarify the purpose of the benchmark class.

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
olavloite: Done

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
olavloite: Done

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
olavloite: Done

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
olavloite: Done

src/test/java/com/google/cloud/spanner/pgadapter/benchmark/MockServerBenchmark.java
olavloite: Done

rayudu3745 (approved):

PR #4575, opened 2026-07-01
chore: deprecate experimental host option/parameter

sagnghos:
Uptaking change in https://github.com/googleapis/google-cloud-java/pull/13236 to deprecate (experimental host / external host) parameters / options to uptake the new Type parameter

To run Integration test against Spanner Omni use below parameters:
```
-DUSE_PLAIN_TEXT=true
-DSPANNER_OMNI_HOST=localhost:15000
-DSPANNER_OMNI_CLIENT_CERT=/path/to/client/cert -DSPANNER_OMNI_CLIENT_KEY=/path/to/client/key
-Djavax.net.ssl.trustStore=$JAVA_HOME/lib/security/cacerts -Djavax.net.ssl.trustStoreType=JKS
```

Additional change in ITLiquibaseTest

Fix ProcessBuilder Deadlock in ITLiquibaseTest
Problem: ITLiquibaseTest was prone to freezing indefinitely when executing the mvn liquibase subprocess. The test code attempted to read the process's stderr stream to completion before reading stdout. Because the Liquibase Maven plugin generates a massive amount of stdout logs (fetching dependencies, evaluating changelogs), the OS stdout buffer would fill up, causing the subprocess to pause. This resulted in an I/O deadlock: Java was waiting for stderr to finish, while the subprocess was waiting for Java to clear stdout.

Solution: Merged the error and output streams by setting builder.redirectErrorStream(true) on the ProcessBuilder. This routes all stderr output directly into stdout, allowing the test thread to continuously drain a single unified stream. This prevents the buffer from filling up and completely eliminates the deadlock.

### DIFF
--- a/src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
+++ b/src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
@@ -33,14 +33,13 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.spanner.v1.DatabaseName;
-import io.grpc.ExperimentalApi;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.time.Duration;
-import java.util.HashMap;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
@@ -445,20 +444,6 @@ public class OptionsMetadata {
     Builder useClientCert(String clientCertificate, String clientKey) {
       this.clientCertificate = clientCertificate;
       this.clientKey = clientKey;
-      return this;
-    }
-
-    /**
-     * Configures connection to an experimental host endpoint
-     *
-     * @param experimentalHost url of the experimental host endpoint
-     * @deprecated Use {@link #setType(String)} and {@link #setEndpoint(String)} instead.
-     */
-    @ExperimentalApi("https://github.com/googleapis/java-spanner/pull/3574")
-    @Deprecated
-    Builder setExperimentalHost(String experimentalHost) {
-      setEndpoint(experimentalHost);
-      setType(SPANNER_OMNI_TYPE);
       return this;
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
```diff
@@ -431,27 +437,38 @@ Builder setStartupTimeout(Duration timeout) {
...
-    /*
+    /**
      * Configures connection to an experimental host endpoint
      *
-     * @params experimentalHost url of the experimental host endpoint
+     * @param experimentalHost url of the experimental host endpoint
+     * @deprecated Use {@link #setType(String)} and {@link #setEndpoint(String)} instead.
      */
     @ExperimentalApi("https://github.com/googleapis/java-spanner/pull/3574")
+    @Deprecated
```
olavloite: This is a package-private method. I don't think we need to mark it as deprecated. We can just remove it.

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
```diff
@@ -1199,23 +1216,23 @@ private static boolean usesEmulator(
...
+  private static boolean hasOmniType(CommandLine commandLine, Map<String, String> propertyMap) {
     if (propertyMap == null) {
       return false;
     }
     return commandLine.hasOption(OPTION_SPANNER_ENDPOINT)
-        && commandLine.hasOption(OPTION_JDBC_PROPERTIES)
-        && propertyMap.containsKey(IS_EXPERIMENTAL_HOST_PROPERTY_NAME);
+        && (commandLine.hasOption(OPTION_JDBC_PROPERTIES)
+            && (propertyMap.containsKey("isExperimentalHost")
+                || SPANNER_OMNI_TYPE.equalsIgnoreCase(propertyMap.get(TYPE_PROPERTY_NAME))));
```
olavloite, line 1220: This lookup in propertyMap will fail if someone added `Type=omni` to the connection string. This is actually a generic problem for multiple other properties here as well. Can we change the type of map that is used for this to this:

```java
Map<String, String> properties = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
sagnghos: Done

src/main/java/com/google/cloud/spanner/pgadapter/metadata/OptionsMetadata.java
sagnghos: Done

olavloite (approved):

PR #4840, opened 2026-09-04
perf: write remaining data types directly to the stream

olavloite:
Write all remaining scalar data types (boolean, string/json, bytes, numeric, timestamp, interval, and uuid) directly to the underlying DataOutputStream instead of first allocating intermediate byte arrays.

This eliminates per-column byte array allocations for all scalar column values during query result serialization, reducing heap churn and garbage collection pressure for large result sets.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -119,15 +119,8 @@ jobs:
       - name: Create PostgreSQL test database
         run: |
           psql -h /pg -U postgres -c "CREATE DATABASE pgadapter"
-      - name: Run unit tests
-        run: mvn test -B -Ptest-all -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
-        env:
-          POSTGRES_HOST: /pg
-          POSTGRES_PORT: 5432
-          POSTGRES_USER: postgres
-          POSTGRES_DATABASE: pgadapter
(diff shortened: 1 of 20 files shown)


.github/workflows/integration.yaml
```diff
@@ -119,15 +119,8 @@ jobs:
...
           psql -h /pg -U postgres -c "CREATE DATABASE pgadapter"
-      - name: Run unit tests
-        run: mvn test -B -Ptest-all -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
-        env:
-          POSTGRES_HOST: /pg
-          POSTGRES_PORT: 5432
-          POSTGRES_USER: postgres
-          POSTGRES_DATABASE: pgadapter
-      - name: Run unit tests with virtual threads
-        run: mvn test -Dpgadapter.use_virtual_threads=true -Dpgadapter.use_virtual_grpc_transport_threads=true -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -B -Ptest-all
```
olavloite, line 130: The unit tests were executed twice here, while they are also executed multiple times by the various units.yaml files that we have. So this is removed, and instead we only run the specific test class that requires a live PostgreSQL server. The latter is only set up for the integration tests.

rayudu3745 (approved):

PR #4896, opened 2026-09-17
fix: do not override signal handlers in embedded mode and improve --cmd shutdown

rayudu3745:
### Summary
Fixes JVM signal handling in [`Server.java`](https://github.com/GoogleCloudPlatform/pgadapter/blob/postgresql-dialect/src/main/java/com/google/cloud/spanner/pgadapter/Server.java) across embedded mode, standalone mode, and `--cmd` wrapper mode.

### Problem
1. **Embedded Mode Signal Override**: Previously, `Server` registered global `sun.misc.Signal` handlers (`TERM`, `INT`, `QUIT`) inside a `static {}` initializer block. Whenever a host Java application embedded `ProxyServer`, `Server.class` was loaded (via `Server.setupOpenTelemetry`), overwriting the host JVM's default `java.lang.Terminator` signal handlers with `Server::handleInt` and `Server::handleTerm`. Because `Server.shutdownHandler` is `null` in embedded mode, `SIGINT` (`Ctrl+C`) and `SIGTERM` were silently ignored (`no-op`), preventing host applications from shutting down or running `Runtime.addShutdownHook` threads.
2. **`--cmd` Mode Signal Handling**:
   - In `--cmd` mode (e.g. `psql`), `Server.shutdownHandler` was also `null`, causing `SIGTERM` to be ignored.
   - Pressing `Ctrl+C` (`SIGINT`) sends `SIGINT` to the entire terminal foreground process group (both PGAdapter and the child tool). PGAdapter needs to ignore `SIGINT` so the child tool (`psql`) can handle `Ctrl+C` to cancel active queries without closing the proxy connection.
   - Killing PGAdapter via `SIGTERM` did not terminate the spawned child process.

### Solution
- Remove `static { registerSignalHandlers(); }` so loading `Server.class` has zero side effects on host JVM signal handlers in embedded mode.
- Call `registerSignalHandlers()` explicitly in the standalone server branch of `Server.main()` after `Server.shutdownHandler` is initialized.
- In `runCommand()` (`--cmd` mode):
  - Call `ignoreInterruptSignal()` to ignore `SIGINT` in PGAdapter while allowing the child process to handle `Ctrl+C`.
  - Register a JVM shutdown hook (`Runtime.addShutdownHook`) that cleanly destroys the child process and stops `ProxyServer` when PGAdapter receives `SIGTERM`.


src/main/java/com/google/cloud/spanner/pgadapter/Server.java
```diff
@@ -121,66 +125,152 @@ public static void runCommand(
...
+      Runtime.getRuntime().addShutdownHook(shutdownHook);
+    } catch (IllegalStateException shutdownInProgress) {
+      stopCommand(process, proxyServer);
+      return;
+    }
+    try {
+      process.waitFor();
+    } finally {
+      removeShutdownHook(shutdownHook);
+    }
```
olavloite, line 143: nit: to prevent interrupts from disturbing stopping the child process, you could add a catch clause for InterruptedExceptions like this:


```java
try {
  process.waitFor();
} catch (InterruptedException interrupted) {
  stopCommand(process, proxyServer);
  throw interrupted;
} finally {
  removeShutdownHook(shutdownHook);
}
```

olavloite (approved):

src/main/java/com/google/cloud/spanner/pgadapter/Server.java
rayudu3745: Thanks, made the changes

PR #4901, opened 2026-09-18
perf: skip analyzeQuery when executing COPY TO WITH HEADER

olavloite:
Read column names directly from the query result set metadata instead of issuing a separate analyzeQuery(PLAN) RPC before running COPY TO.

Also emit the header row as its own CopyData message during result prefix generation. This avoids duplicate headers across threads when running partitioned queries and ensures the header is sent even if the query returns zero rows.


src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyToStatement.java
```diff
@@ -220,11 +201,23 @@ public IntermediatePortalStatement createPortal(
...
+    // is also the only option here: the rows of a partitioned COPY are written by several threads,
+    // so there is no row that is guaranteed to be written first.
     return new WireOutput[] {
-      new CopyOutResponse(
+      copyOutResponse,
+      new CopyDataResponse(
           this.outputStream,
-          resultSet.getColumnCount(),
-          this.parsedCopyStatement.format.getDataFormat().getCode())
+          csvFormat.format((Object[]) createHeader(resultSet)),
```
rayudu3745: nit: If `FORCE_QUOTE *` is used with `HEADER` (e.g., `COPY tbl TO STDOUT (FORMAT CSV, HEADER, FORCE_QUOTE *)`), `csvFormat` has `QuoteMode.ALL_NON_NULL`, so `csvFormat.format((Object[]) createHeader(resultSet))` will also wrap all header column names in quotes (`"col_bigint","col_bool"`).
In Postgres, `FORCE_QUOTE *` only applies to data values and not to the header row (which still uses `QuoteMode.MINIMAL`):

rayudu3745 (approved):

src/main/java/com/google/cloud/spanner/pgadapter/statements/CopyToStatement.java
olavloite: Fixed

PR #4905, opened 2026-09-18
fix: prevent failed Parse from leaving previous unnamed statement registered

olavloite:
Previously, when a Parse message failed during local statement creation (resulting in an InvalidStatement) or during a backend Describe('S') call, ParseMessage either skipped registering the statement in ConnectionHandler.statementsMap or only cleaned up named statements.

As a result, if a client previously executed an unnamed statement ("") and then sent a Parse for an invalid unnamed statement, the previous unnamed statement remained registered under "". Subsequent Bind/Execute messages (or the synthesized Bind/Describe/Execute messages in SimpleQueryStatement) would then bind to and re-execute the previous unnamed statement instead of failing.

Additionally, InvalidStatement was not queued on BackendConnection, so a failed Parse or client-side statement in a pipeline did not abort subsequent buffered statements before Sync, nor did it transition an active transaction to the aborted state (25P02).

This change fixes this by:
- Always registering the parsed statement (including InvalidStatement) during ParseMessage.buffer() so any previous unnamed statement is replaced, and closing the statement in ParseMessage.flush() and ParseMessage.abort() if it has an exception (for both named and unnamed statements, as well as failed Describe('S') messages).
- Queueing InvalidStatement on BackendConnection so that parse and client-side statement failures abort subsequent pipelined statements and mark active transactions as aborted.
- Aborting remaining buffered messages in ExtendedQueryProtocolHandler and SimpleQueryStatement as soon as an earlier message in the pipeline emits an ErrorResponse.

Fixes b/562755027


rayudu3745:
LGTM

rayudu3745 (approved):

PR #4922, opened 2026-09-21
fix: do not apply statement_timeout to transaction rollbacks

olavloite:
When a statement_timeout is set on a connection, the Spanner Java client applies that timeout to all RPCs executed on the transaction, including Rollback RPCs. If a statement in a read/write transaction fails (for example due to a short statement_timeout), the subsequent implicit or explicit rollback could also time out before reaching Spanner. Because exceptions during rollback are ignored by the client, the transaction would remain active on Spanner and continue holding locks until the server-side transaction timeout expired.

This change temporarily clears the statement timeout on the underlying Spanner connection while executing ROLLBACK and ROLLBACK TO SAVEPOINT, and restores the previous timeout afterward.


src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
```diff
@@ -1242,6 +1243,43 @@ private void flush(boolean isSync) {
...
+      spannerConnection.rollback();
+      return ROLLBACK_RESULT;
+    } finally {
+      if (timeoutUnit != null) {
+        spannerConnection.setStatementTimeout(previousTimeout, timeoutUnit);
+      }
+    }
+  }
+
+  private void rollbackToSavepointWithoutTimeout(String savepointName) {
```
rayudu3745: In `rollbackWithoutTimeout()` above, we abort any active DML batch (`if (spannerConnection.isDmlBatchActive()) spannerConnection.abortBatch()`) and clear the statement tag (`spannerConnection.setStatementTag(null)`), whereas `rollbackToSavepointWithoutTimeout()` does not (e.g., if `SAVEPOINT s1` was created before `START BATCH DML` or `SET spanner.statement_tag`, rolling back to `s1` would hit the active DML batch or leave the statement tag set).

Should `rollbackToSavepointWithoutTimeout()` also abort any active DML batch and clear `statement_tag` ?

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
```diff
@@ -954,4 +957,71 @@ public void testStartupConnectionPropertiesInUrl() throws SQLException {
...
+      statement.execute("set statement_timeout = '1ms'");
+      stopwatch = Stopwatch.createStarted();
+      statement.execute("rollback to savepoint s1");
+      assertTrue(stopwatch.elapsed(TimeUnit.MILLISECONDS) >= 20L);
+      statement.execute("rollback");
+      assertEquals(4, mockSpanner.countRequestsOfType(RollbackRequest.class));
+
+      try (ResultSet resultSet = statement.executeQuery("show statement_timeout")) {
+        assertTrue(resultSet.next());
+        assertEquals("1ms", resultSet.getString(1));
```
rayudu3745: Here `statement_timeout` is `0` before `begin` and `savepoint s1`, and `set statement_timeout = '1ms'` is executed *after* `savepoint s1`. After `rollback to savepoint s1` and `rollback`, we are asserting that `show statement_timeout` is still `'1ms'`. 
is the expected postgres behaviour ? or `set statement_timeout = '1ms'`  should also be rolled back ?

src/main/java/com/google/cloud/spanner/pgadapter/statements/BackendConnection.java
olavloite: That is a good point. Yes, BUT:
1. This pull request should not have added `rollbackToSavepointWithoutTimeout()` in the first place. This change is only intended for the __internal__ rollbacks that PGAdapter executes in order to emulate PostgreSQL's aborted transaction state. That is: Whenever an error happens in a transaction, that transaction should be marked as 'aborted' (failed), and all further statements should be rejected. In order to release locks on Spanner, we explicitly rollback those transactions when that happens, and it is these (and __only__ these) rollbacks that should be executed without using `statement_timeout`.
2. The points about aborting DML batches and clearing the statement_tag are correct, but they should be fixed in the Java client.

I removed the code for `rollbackToSavepointWithoutTimeout()` and also the changes that made 'normal' rollbacks be executed without `statement_timeout`. This pull request should __only__ change the behavior of the internal rollbacks.

src/test/java/com/google/cloud/spanner/pgadapter/JdbcSimpleModeMockServerTest.java
olavloite: Yes, but same as above: This is something that is handled by the Java client, and should be fixed there.

rayudu3745 (approved):

